PR #93, opened 2022-11-22
Different seo schema for differen pages

trean:
Added SEO schema (aka schema.org) for articles and benchmark page (`article` entity). Now schema generates from article parameters but also can be overridden with the `seo_schema` param. The `organization` entity now used only on the main page. 

Also, the `seo_schema` parameter can be optionally used on other pages.

### DIFF
--- a/qdrant-landing/themes/qdrant/layouts/partials/seo_schema.html
+++ b/qdrant-landing/themes/qdrant/layouts/partials/seo_schema.html
@@ -93,9 +93,6 @@
 </script>
 {{ end }}
 
-{{ if .IsHome }}
-{{ end }}
-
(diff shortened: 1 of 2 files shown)


qdrant-landing/themes/qdrant/layouts/partials/seo_schema.html
```diff
@@ -66,6 +67,34 @@
...
+            "@type": "Person",
+            "name": {{ .Params.author }},
+            "url": {{ .Params.author_link }}
+        }]
+    }
+</script>
+{{ end }}
+
+{{ if .IsHome }}
+{{ end }}
```
generall, line 97: is it required?

generall (approved):

PR #101, opened 2022-12-07
memory consumption article

generall:


qdrant-landing/content/articles/memory-consumption.md
```diff
@@ -0,0 +1,366 @@
...
+    3. Process may use disk cache.
+3. As a result, if you see `10Gb` memory consumption in `htop`, it doesn't mean that your process actually needs `10Gb` of RAM to work.
+-->
+
+
+When it comes to measuring the memory consumption of our processes, we often rely on tools such as `htop` to give us an indication of how much RAM is being used. However, this method can be misleading and doesn't always accurately reflect the true memory usage of a process.
+
+There are many different ways in which `htop` may not be a reliable indicator of memory usage. 
+For instance, a process may allocate memory in advance but not use it, or it may not free deallocated memory, leading to overstated memory consumption. 
+Process may be forked, which means that it will have a separate memory space, but it will share the same code and data with the parent process.
```
kacperlukawski, line 40: It should be either "A process" or "The process"

qdrant-landing/content/articles/memory-consumption.md
```diff
@@ -0,0 +1,366 @@
...
+
+We can use the `glove-100-angular` and scripts from the [vector-db-benchmark](https://github.com/qdrant/vector-db-benchmark) project to upload and query the vectors.
+With the first run we will use the default configuration of Qdrant with all data stored in RAM.
+
+```bash
+# Upload vectors
+python run.py --engines qdrant-all-in-ram --datasets glove-100-angular
+```
+
+After uploading vectors, we will repeat the same experiment with different allowed RAM usage to see how they affect the memory consumption and search speed.
```
kacperlukawski, line 103: Maybe just "different RAM limits" instead of "allowed RAM usage"?

qdrant-landing/content/articles/memory-consumption.md
```diff
@@ -0,0 +1,366 @@
...
+  ...
+  "optimizers_config": {
+    "mmap_threshold_kb": 20000
+  }
+}
+
+```
+This configuration tells Qdrant to use mmap for vectors if the segment size is greater than 20000Kb (which is approximately 40K 128d-vectors).
+
+Now the out-of-memory happens when we use allow only **600mb** RAM.
```
kacperlukawski, line 187: Shouldn't it be "Now the out-of-memory happens when we allow using **600mb** RAM only"?

qdrant-landing/content/articles/memory-consumption.md
```diff
@@ -0,0 +1,366 @@
...
+
+<details>
+  <summary>Experiments details</summary>
+
+| Memory | Requests/s    |
+|--------|---------------|
+| 1200mb | 759.94 |
+| 1100mb | 687.00 |
+| 1000mb | 10     |
```
kacperlukawski, line 196: Is it really just 10 rps with this setup? Why is it that different to 1100mb?

qdrant-landing/content/articles/memory-consumption.md
```diff
@@ -0,0 +1,366 @@
...
+read: IOPS=183k, BW=716MiB/s (751MB/s)(8192MiB/11438msec)
+```
+
+Let's see how these results translate into search speed:
+
+| Memory | RPS with IOPS=63.2k | RPS with IOPS=183k |
+|--------|---------------------|--------------------|
+| 600mb  | 5                   | 50                 |
+| 300mb  | 0.9                 | 13                 |
+| 200md  | 0.5                 | 8                  |
```
kacperlukawski, line 346: 200md -> 200mb

kacperlukawski (changes requested):
There are some minor changes only, but overall I'm surprised that it has been generated with GPT!

qdrant-landing/content/articles/memory-consumption.md
generall: yes, I was also a bit surprised, but yes. That's a difference between a very slow disk and RAM

kacperlukawski (approved):

PR #115, opened 2023-02-28
Add Bloop case study

kacperlukawski:
I am adding bloop case study, without a list view.


generall:
Let's use same template as for articles

PR #130, opened 2023-04-03
Add LangChain into integrations

kacperlukawski:
Surprisingly the LangChain integration was not mentioned. Thus I'm adding it.


kacperlukawski:
Wrong repo

PR #140, opened 2023-05-19
Article: Product Quantization

kacperlukawski:
This PR contains an article about Product Quantization.


qdrant-landing/content/articles/product-quantization.md
```diff
@@ -0,0 +1,211 @@
...
+         <td>474 s</td>
+      </tr>
+   </tbody>
+</table>
+
+It turns out that in some cases, Product Quantization may not only reduce the memory usage, 
+but also the search time.
+
+## Good practices
+
```
generall, line 208:
```suggestion

Compared to Scalar Quantization, Product Quantization offers a higher compression rate. However, this comes with considerable trade-offs in accuracy, and at times, in-RAM search speed.

Product Quantization tends to be favored in certain specific scenarios:

- Deployment in a low-RAM environment where the limiting factor is the number of disk reads rather than the vector comparison itself
- Situations where the dimensionality of the original vectors is sufficiently high
- Cases where indexing speed is not a critical factor

In circumstances that do not align with the above, Scalar Quantization should be the preferred choice.

```

agourlay (approved):
Great article, it improved my understanding of product quantization :+1:

PR #146, opened 2023-05-25
migrate blood case study for publication

davidmyriel:
Could you please take a look so we can publish. 
Not sure if all front matter is OK. Also, I didn;t touch the case studies static folder.


qdrant-landing/content/articles/bloop.md
```diff
@@ -4,6 +4,9 @@ short_description: bloop is a fast code-search engine that combines semantic sea
 description: bloop is a fast code-search engine that combines semantic search, regex search and precise code navigation
 social_preview_image: /case_studies_data/bloop/social_preview.png
 preview_dir: /case_studies_data/bloop/preview
```
kacperlukawski, line 6: It might be better to move the images to `articles_data` as well if we don't want to have case studies separated

kacperlukawski (changes requested):
Except for the images, it looks fine.

qdrant-landing/content/articles/bloop.md
davidmyriel: will do

PR #158, opened 2023-06-01
Mention the certificate TTL is currently only supported for REST

timvisee:
@davidmyriel I wasn't sure if this was still in your queue. Adding this PR just to be sure it gets included.

I'm not sure if this is the correct place to edit this since the recent changes anyway.


davidmyriel:
This is it. lgtm

PR #160, opened 2023-06-02
update doc as per old commit

davidmyriel:
transferred changes from old PR (https://github.com/qdrant/docs/pull/122) over to new docsite + a few more minor changes


davidmyriel:
Also, when referencing other docs, please use directory path rather than URL. Thanks!

timvisee (approved):

PR #161, opened 2023-06-02
fix aliases for SEO

davidmyriel:
In order to avoid indexing/ranking issues with the search engines, I fixed path aliases to keep the indexes connected to the same content.

The only pages that aren't the same are "Integrations" and "How-Tos". Those have been separated into multiple topics. I hope that doesn't cause any issues.

### DIFF
--- a/qdrant-landing/content/documentation/concepts/indexing.md
+++ b/qdrant-landing/content/documentation/concepts/indexing.md
@@ -1,6 +1,8 @@
 ---
 title: Indexing
 weight: 90
+xaliases:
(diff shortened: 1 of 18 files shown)


qdrant-landing/content/documentation/concepts/indexing.md
```diff
@@ -1,6 +1,8 @@
 ---
 title: Indexing
 weight: 90
+xaliases:
```
generall, line 4:
```suggestion
aliases:
```

PR #162, opened 2023-06-05
mighty article

llogiq:
This is missing the preview/title images. Will push those soon.


agourlay:
I am bit confused regarding the usage of Mighty.

It is not clear which sentence transformer model is used by Mighty and this article is not explaining how the Qdrant collection is initialized regarding its vector size.

I feel like an important step is missing.

llogiq:
It's using [HuggingFace's MiniLM-L6-v2 sentence transformers](https://huggingface.co/sentence-transformers/all-MiniLM-L6-v2).

llogiq:
@davidmyriel I removed the lambda stuff because it doesn't really fit here. As a result, the document becomes much more simple and readable.

PR #173, opened 2023-06-06
navigation: fix capitalization for github

squat:
The link to GitHub in the navigation bar should be stylized as "GitHub" rather than "Github".


davidmyriel:
thank you for spotting, fixed in a separate commit

PR #175, opened 2023-06-07
update examples page

davidmyriel:
-extracted notebooks from examples repo from Tutorials and put them in a separate section
-the logic is that Tutorials are bite sized, contained instruction sets, while examples are not as focused and have a more detailed set of explanations
-removed extra files that arent needed


kacperlukawski (approved):
It's definitely a better way

PR #185, opened 2023-06-14
Fix langchain docs

kacperlukawski:
Some users reported our examples were using deprecated parameters. This is fixed now, and I also included more examples of using Langchain with different Qdrant deployment modes.


davidmyriel:
lgtm

PR #188, opened 2023-06-15
Update points.md

suryaa62:
Under Retrieve points, It says that the method also has the with_vector parameter. For python qdrant_clients, its with_vectors

https://github.com/qdrant/qdrant-client/issues/189


dhruv-anand-aintech (approved):

davidmyriel:
lgtm, thanks for spotting

PR #197, opened 2023-06-21
Document customizable temporary directory

timvisee:
Draft for documentation on the customizable temporary directory feature.

I'm not sure if I like it being placed here, but I don't have a better suggestion. Any ideas? Putting it in _Configuration_ or in a _Tutorial_ section could be an option.


davidmyriel:
Still relevant to snapshots, but I updated the way top section looks. Configuring location is a section of its own now. 
I also included the code in config.yaml sample.

davidmyriel (approved):

generall (approved):

PR #201, opened 2023-06-25
fix: fix nested filter example

joein:
https://github.com/qdrant/qdrant-client/issues/196


davidmyriel:
lgtm, ty for spotting

PR #202, opened 2023-06-25
fix: fix with_vectors arg in python examples

joein:
qdrant-client does not have `with_vector` argument, instead it has `with_vectors`


davidmyriel (approved):
thank you for spotting, both work btw

PR #203, opened 2023-06-27
Update neural-search.md

aardoiz:
Change qdrant_client.search "top" parameter for "limit" as the former is no longer compatible with current version:

File "/venvs/qdn/lib/python3.10/site-packages/qdrant_client/qdrant_client.py", line 287, in search
    assert len(kwargs) == 0, f"Unknown arguments: {list(kwargs.keys())}"
AssertionError: Unknown arguments: ['top']


davidmyriel (approved):
thanks for spotting

PR #204, opened 2023-06-29
Serverless semantic search article

llogiq:
There are still a few TODOs left, but the gist of the article is there. I picked a cookbook style here.


qdrant-landing/content/articles/serverless.md
```diff
@@ -0,0 +1,290 @@
...
+    --statement-id url
+$ # Here for simplicity unauthenticated URL access. Beware!
+$ aws lambda create-function-url-config \
+    --function-name $LAMBDA_FUNCTION_NAME \
+    --region $LAMBDA_REGION \
+    --cors "AllowOrigins=*,AllowMethods=*,AllowHeaders=*" \
+    --auth-type NONE
+```
+
+Now you can go to your *Function Overview* and click on the Function URL. This should show something like the following:
```
davidmyriel, line 119: "You should see something like this:"

qdrant-landing/content/articles/serverless.md
```diff
@@ -0,0 +1,290 @@
...
+
+```text
+Hello, lambda!
+```
+
+Congratulations! You have set up a lambda function in Rust. On to the next ingredient:
```
davidmyriel, line 125: No congrats needed yet. Seems kinda fake

qdrant-landing/content/articles/serverless.md
```diff
@@ -0,0 +1,290 @@
...
+
+## Embedding
+
+Most providers supply a simple https GET or POST interface we can use. Most will have an API key, which you have to supply in an authentication header. Let's use OpenAI as an example, which is one of the more complex APIs. We need to extend our dependencies with `reqwest` and also add `anyhow` for easier error handling:
```
davidmyriel, line 129: Most twice......

qdrant-landing/content/articles/serverless.md
```diff
@@ -0,0 +1,290 @@
...
+        .send()
+        .await?
+        .json()?;
+    Ok(embedding)
+}
+```
+
+You should note that OpenAI's ada-002 model emits 1536-dimensioned vectors.
+
+Other providers have similar interfaces. Consult our [integration docs](https://qdrant.tech/documentation/integrations) for further information. See how little code it took to get the embedding? Also with [AI Grant](https://aigrant.org/), you can apply for free credits that allow for a lot of searches before you'll have to spend a dime.
```
davidmyriel, line 165: hmm it seems like advertising AI grants should be more prominent. after all the article is all about getting cheap/free resources

qdrant-landing/content/articles/serverless.md
```diff
@@ -0,0 +1,290 @@
...
+
+While we're at it, it's a good idea to write a small test:
```
davidmyriel, line 167: You need to say at the beginning what the test is for. Otherwise the reader has to go through the code to figure it out.

qdrant-landing/content/articles/serverless.md
```diff
@@ -0,0 +1,290 @@
...
+        collection_name,
+        limit: 5, // use what fits your use case here
+        with_payload: Some(true.into()),
+        vector: embed(client, text, api_key)?,
+        ..Default::default()
+    }).await?.result)
+}
+```
+
+And that's it. Obviously, you can also filter by adding a `filter: ...` field to the `SearchPoints`, and you will likely want to process the result further, but the example code already does that, so feel free to start from there in case you need this functionality.
```
davidmyriel, line 257: can't start the sentence with "and" its against literary rules......either remove or elaborate.

qdrant-landing/content/articles/serverless.md
```diff
@@ -0,0 +1,290 @@
...
+    Finished release [optimized] target(s) in 1m 27s
+$ # Update the function
+$ aws lambda update-function-code --function-name $LAMBDA_FUNCTION_NAME \
+     --zip-file fileb://./target/lambda/page-search/bootstrap.zip \
+     --region $LAMBDA_REGION
+```
+
+## Discussion
+
+Lambda works by spinning up your function once the URL is called, so they don't need to keep the compute on hand unless it is actually used. This means that the first call will be burdened by some 1-2 seconds of latency for loading the function, later calls will resolve faster. Of course, there is also the latency for calling the embeddings provider and Qdrant. On the other hand, the free tier doesn't cost a thing, so you certainly get what you pay for. And for many use cases, a result within a few seconds is acceptable.
```
davidmyriel, line 288: seems like some of this should be at the top.....

i really think we need to push the free aspect - people love free

qdrant-landing/content/articles/serverless.md
llogiq: Agreed. I'm moving it up to the "ingredients" part.

kacperlukawski (changes requested):
I couldn't get it running. In my case, all the requests end up with `Internal Server Error`. I even used the example from the repository directly, but it didn't help. I'll check the logs tomorrow.

qdrant-landing/content/articles/serverless.md
```diff
@@ -0,0 +1,291 @@
...
+
+## Ingredients
+
+* A [Rust](https://rust-lang.org) toolchain
+* [cargo lambda](cargo-lambda.info) (install via package manager, [download](https://github.com/cargo-lambda/cargo-lambda/releases) binary or `cargo install cargo-lambda`)
+* The [aws CLI](https://aws.amazon.com/cli)
+* Qdrant instance ([free tier](https://cloud.qdrant.tech) available)
+* An embedding provider service of your choice (see our [integration docs](https://qdrant.tech/documentation/integrations). I don't know any with a free tier, but you can get credits from [AI Grant](https://aigrant.org))
+* AWS Lambda account (12-month free tier available)
+
```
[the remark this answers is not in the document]
llogiq: That's right one line above.

qdrant-landing/content/articles/serverless.md
```diff
@@ -0,0 +1,291 @@
...
+$ export LAMBDA_REGION=us-east-1
+$ cargo lambda build --release --arm --output-format zip
+  Downloaded libc v0.2.137
+# [..] output omitted for brevity
+    Finished release [optimized] target(s) in 1m 27s
+$ # Upload the function
+$ aws lambda create-function --function-name $LAMBDA_FUNCTION_NAME \
+    --handler bootstrap \
+    --architectures arm64 \
+    --zip-file fileb://./target/lambda/page-search/bootstrap.zip \
```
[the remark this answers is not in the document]
llogiq: Good catch! :+1:

PR #206, opened 2023-06-30
Add warning to protect internal gRPC ports

timvisee:
This adds an (important) warning for users to protect their internal gRPC ports, used for cluster communication in distributed mode. These are not protected by an API key, which may be unexpected.

I'm not 100% happy with this text. @davidmyriel Maybe you can improve it a bit if you'd like.


davidmyriel:
I might need to rework some things on this page.

PR #210, opened 2023-07-04
Expose gRPC port in quick start command

timvisee:
This exposes port 6334 for gRPC in the quick start command.

The gRPC interface seems to be used more commonly. I've seen some issues caused by this port not being exposed. This hopefully prevents that.

Asking @generall for review to approve whether this is a good idea.


davidmyriel:
This is not for quickstart IMO. I am currently creating a more elaborate docker instruction set and will include this.

davidmyriel:
Addressed in https://github.com/qdrant/landing_page/pull/259

PR #225, opened 2023-07-17
remove all instances of plural from with_vector

davidmyriel:
Please LMK if I missed anything?

### DIFF
--- a/qdrant-landing/content/documentation/concepts/points.md
+++ b/qdrant-landing/content/documentation/concepts/points.md
@@ -631,7 +631,7 @@ client.retrieve(
 )
 ```
 
-This method has additional parameters `with_vectors` and `with_payload`. 
+This method has additional parameters `with_vector` and `with_payload`. 
 Using these parameters, you can select parts of the point you want as a result.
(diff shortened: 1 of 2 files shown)


agourlay:
@joein We have users being confused regularly by this on Discord.
This looks correct to me but I have just realized you performed the opposite transformation https://github.com/qdrant/landing_page/pull/202

What is the correct usages here?

davidmyriel:
There could be a mixup as per Luis' PR here: https://github.com/qdrant/qdrant/pull/1697

joein:
@agourlay we used to have a lot of complaints about it, but i haven't seen them for a while (since we introduced checks for incorrect keyword arguments in python client)
(correct me if i am wrong)

python client uses `with_vectors`

joein:
@davidmyriel sorry, i didn't see your comment in #202, `with_vector` does not work in python, it raises an exception

joein:
Actually, I think we can support both in python client, it is not that hard to implement

davidmyriel:
Closing as with_vectors seems appropriate

PR #228, opened 2023-07-24
fix: revised collections definition with respect to addition of named vectors in the API

BagchiMB:


qdrant-landing/content/documentation/overview/_index.md
```diff
@@ -114,7 +114,7 @@ are the terminologies you should get familiar with.
 
 - [Collections](../concepts/collections/): A collection is a named set of 
 points (vectors with a payload) among which you can search. Vectors within the same collection 
-can have different dimensionalities and be compared by a single metric.
```
timvisee, line 117: No, a collection can have multiple named vectors with different dimensionalities. So the original text is correct.

Here is an example with multiple named vectors: https://qdrant.tech/documentation/concepts/collections/#collection-with-multiple-vectors

The original text is a bit confusing though, so maybe we can improve on that.

You are correct that a named vector must always have vector data that has the same dimensionality, but that is not what the text here is referring to.

BagchiMB: I also looked into this, I tried to maintain the homogeneity with respect to the opening line [here](https://qdrant.tech/documentation/concepts/collections/#collections), I thought being that page as the introduction to all the subtopics that are mentioned in User Manual > Concepts. 

Maybe we can change at both of the places and add a reference to [https://qdrant.tech/documentation/concepts/collections/#collection-with-multiple-vectors](https://qdrant.tech/documentation/concepts/collections/#collection-with-multiple-vectors) at both the occurrences, any thoughts?

timvisee: Very good point. I wasn't aware that we had the same text in a different place.

Maybe we can change it to something like this:

> A collection is a named set of points (vectors with a payload) among which you can search. Vectors within the same collection must have the same dimensionality and be compared by a single metric. [Named vectors](https://qdrant.tech/documentation/concepts/collections/#collection-with-multiple-vectors) can be used to have multiple vectors in a single point, each of which can have their own dimensionality and metric requirements.

@BagchiMB If you think the above suggestion clears things up, could you change the occurrences to the above in this PR?

@davidmyriel What do you think about the above text change?

BagchiMB: I think it works! Making the change and updating the PR.

timvisee (approved):

BagchiMB:
HI @timvisee any update on when this will get merged?

timvisee:
> HI @timvisee any update on when this will get merged?

The person handling our docs updates is currently out of office.

Maybe @generall can take a look for a second opinion before merging.

BagchiMB:
Hi @timvisee any update on this?

timvisee:
Thanks for pinging. After a second careful read I'm fine with it to merge.

Thanks once again!

PR #230, opened 2023-07-25
Fix using `collection_name` in group requests

kacperlukawski:
This PR makes the examples consistent with the API. The issue was reported here: https://github.com/qdrant/qdrant/issues/2325

### DIFF
--- a/qdrant-landing/content/documentation/concepts/search.md
+++ b/qdrant-landing/content/documentation/concepts/search.md
@@ -811,7 +811,7 @@ POST /collections/chunks/points/search/groups
     // Lookup parameters
     "with_lookup": {
         // Name of the collection to look up points in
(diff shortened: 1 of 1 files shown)


timvisee (approved):

coszio (approved):
In hindsight it should've been `collection_name` in API, thanks for the doc fix 🙌

timvisee:
> In hindsight it should've been `collection_name` in API, thanks for the doc fix 🙌

Maybe we can introduce an alias. It currently has an explicit rename: https://github.com/qdrant/qdrant/blob/7164ac4a5987d28f1c93f5712aef8e09e7d93555/lib/collection/src/lookup/mod.rs#L21

generall (approved):

PR #231, opened 2023-07-27
New search-as-you-type article

llogiq:
~~I still need to find a good conclusion.~~ Concluding with a benchmark, as it should be.


qdrant-landing/content/articles/sayt.md
```diff
@@ -0,0 +1,112 @@
...
+
+The workflow looks like this:
+
+![Search Qdrant by Embedding](/articles_data/sayt/Qdrant_Search_by_Embedding.png)
+
+Unlike with some other libraries, the ort crate's `Session` type is `Send` and `Sync`, which makes it easy to preload the model and calculate the embeddings directly in the request handler. Wrapping the tokenizeri, session and Qdrant client in an actix `Data`, which works like an `Arc`, is sufficient to have everything in place for the search. Using actix' builtin JSON deserialization makes getting the search parameters a breeze.
```
timvisee, line 23: Likely a typo.

```suggestion
Unlike with some other libraries, the ort crate's `Session` type is `Send` and `Sync`, which makes it easy to preload the model and calculate the embeddings directly in the request handler. Wrapping the tokenizer, session and Qdrant client in an actix `Data`, which works like an `Arc`, is sufficient to have everything in place for the search. Using actix' builtin JSON deserialization makes getting the search parameters a breeze.
```

qdrant-landing/content/articles/sayt.md
```diff
@@ -0,0 +1,122 @@
...
+fn prefix_to_id(prefix: &str) -> PointId {
+    let len = prefix.len();
+    PointId::from(u64::from_le_bytes(if len < 8 {
+        let mut result = [0_u8; 8];
+        result[..len].copy_from_slice(prefix.as_bytes());
+        result
+    } else {
+        prefix.as_bytes()[..8].try_into().unwrap()
+    }))
+}
```
generall, line 100: This technique is, actually, controversial. Some languages, like Russian, have more than 1 byte per character in Unicode.  So 5 letter prefix won't fit 8 bytes of ID. In future I would prefer this to be replaced with hashing trick, as more universal.

qdrant-landing/content/articles/sayt.md
```diff
@@ -0,0 +1,122 @@
...
+```
+
+If the need arises, one could instead encode the names as UUID, which needs to time-consumingly hash the input, so I decided against this for now.
```
generall, line 103: > which needs to time-consumingly hash the input

Is it really time-consuming? It's nanoseconds

qdrant-landing/content/articles/sayt.md
```diff
@@ -0,0 +1,122 @@
...
+![Benchmark Results](/articles_data/sayt/benchmark.png)
+
+| query     | `io`       | `ring`    | `io_uring` |
+|-----------|------------|-----------|------------|
+| Python 🐍 | 4½ ± ½ ms* | 16 ± 4 ms | 16 ± 4 ms  |
+| Rust   🦀 | 1½ ± ½ ms  | 1½ ± ½ ms |  5 ± 1 ms  |
+
+\* full-text search only
+
+The Rust version consistently beats the Python version in terms of speed and offers a semantic search even on few-character queries. If the prefix cache is hit (as in the "ring" measurement), the semantic search can even get more than ten times faster than the Python version. The general speedup is due to both the relatively lower overhead of Rust + Actix Web compared to Python + FastAPI (even if that already performs admirably), as well as using ONNX Runtime instead of SentenceTransformers for the embedding. The prefix cache gives the Rust version a real boost by doing a semantic search without doing any embedding work.
```
generall, line 122: I would add here that all those milliseconds are important here for 2 reasons:

- search-as-you-type speed is bound to the user's typing speed. If it is doing to be significantly slower than the keystroke, user will notice.
- search-as-you-type generate search query on each letter. So it usually 3-5x more load than usual search queries.

qdrant-landing/content/articles/sayt.md
```diff
@@ -0,0 +1,122 @@
...
+Since the operations themselves are far faster than the network whose fickle nature would have swamped most measurable differences, I benchmarked both the Python and Rust services locally. Note that with search terms of up to 3 characters, the Python version merely does a text search, not a semantic search, while the Rust version always does a more complex semantic search. I'm measuring both versions on the same AMD Ryzen 9 5900HX with 16GB RAM running Linux. The table shows the average time and error bound in milliseconds. I only measured up to a thousand concurrent requests. None of the services showed any slowdown with more requests in that range. I do not expect our service to become DDOS'd, so I didn't benchmark with more load. Without further ado, here are the results:
+
+![Benchmark Results](/articles_data/sayt/benchmark.png)
+
+| query     | `io`       | `ring`    | `io_uring` |
+|-----------|------------|-----------|------------|
+| Python 🐍 | 4½ ± ½ ms* | 16 ± 4 ms | 16 ± 4 ms  |
+| Rust   🦀 | 1½ ± ½ ms  | 1½ ± ½ ms |  5 ± 1 ms  |
+
+\* full-text search only
```
generall, line 120: does this make sense to measure, if we compare text search with vector search?
Maybe it's better to just mention how long it takes to make an embedding? Especially, is the embedding is on remote server. I would say it's at least 40ms to just this encoding round-trip

generall (changes requested):

qdrant-landing/content/articles/sayt.md
llogiq: I agree that hashing is strictly more applicable in the general case, but currently our site is available in English only. So I still think we can take some freedom to use the specifics of our data to our advantage here.

qdrant-landing/content/articles/sayt.md
llogiq: Depends on the hash function. But I agree, even if it's a few 10s or even 100s of nanos, it's unlikely to show up in the profile.

qdrant-landing/content/articles/sayt.md
```diff
@@ -0,0 +1,122 @@
...
+
+To demo it, I used our [Qdrant documentation website](https://qdrant.tech/documentation)'s page search, replacing our previous Python implementation. So in order to not just spew empty words, here is a benchmark, showing different queries that exercise different code paths.
+
+Since the operations themselves are far faster than the network whose fickle nature would have swamped most measurable differences, I benchmarked both the Python and Rust services locally. Note that with search terms of up to 3 characters, the Python version merely does a text search, not a semantic search, while the Rust version always does a more complex semantic search. I'm measuring both versions on the same AMD Ryzen 9 5900HX with 16GB RAM running Linux. The table shows the average time and error bound in milliseconds. I only measured up to a thousand concurrent requests. None of the services showed any slowdown with more requests in that range. I do not expect our service to become DDOS'd, so I didn't benchmark with more load. Without further ado, here are the results:
+
+![Benchmark Results](/articles_data/sayt/benchmark.png)
+
+| query     | `io`       | `ring`    | `io_uring` |
```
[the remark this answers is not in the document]
llogiq: Those are the queries I benchmarked with, chosen to tease out different code paths:
"io" is less than 3 characters, so python does only text search. "ring" is more than 3, but less than 6 characters, so we're still in our prefix cache, but the python version uses a full semantic search. Finally `io_uring` is long enough for the Rust version to do a full search too (for now).

qdrant-landing/content/articles/sayt.md
llogiq: I think it's useful to compare all possible code paths. And I was sure to explain that the python version does a text search in that case. Yet the Rust version using recommend is still roughly 3 times faster.

qdrant-landing/content/articles/sayt.md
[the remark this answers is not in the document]
llogiq: I should perhaps explain my choice of search keywords (which resulted from recently having written that `io_uring` article) better.

qdrant-landing/content/articles/sayt.md
```diff
@@ -0,0 +1,122 @@
...
+            filter,
+            limit: 5,
+            with_payload: Some(true.into()),
+            ..Default::default,
+        }).await
+    {
+        Ok(SearchResponse { result, .. }) => result_json(result),
+        Err(e) => return HttpResponse::InternalServerError().body(e.to_string()),
+    }
+}
```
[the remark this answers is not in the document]
llogiq: I kind of disagree that Rust users are not the target audience; I suggest there will be some overlap. Rust currently gets a lot of hype, being the new language hotness on the block, and so many people, including those who would implement semantic search, are looking at it. Having both a Python and Rust client is a benefit here, because people can try Rust and if they decide on sticking with Python, they can still use Qdrant without having to change the DB.

generall (approved):

llogiq:
@generall I like the new style! Looks less noisy.

qdrant-landing/content/articles/search-as-you-type.md
```diff
@@ -0,0 +1,136 @@
...
+title: Semantic Search As You Type
+short_description: "A demo that does what the title says"
+description: To show off Qdrant's performance, we show how to do a quick search-as-you-type that will come back within a few milliseconds.
+social_preview_image: /articles_data/search-as-you-type/preview/social_preview.jpg
+small_preview_image: /articles_data/search-as-you-type/icon.svg
+preview_dir: /articles_data/search-as-you-type/preview
+weight: -2
+author: Andre Bogus
+author_link: https://llogiq.github.io
+date: 2023-08-07T10:00:00+01:00
```
generall, line 11:
```suggestion
date: 2023-08-14T00:00:00+01:00
```

PR #237, opened 2023-08-06
Introduce new preview images for documentation

generall:
- templates and pipeline to auto-generate images for documentation pages
- Update SEO schema generation to pick up preview files
- Update readme with instructions


trean (approved):
LGTM

PR #238, opened 2023-08-09
add batch upsert calculation doc

davidmyriel:
This PR adds a doc that teaches the user to calculate the size of upsert batches.


davidmyriel:
Addressed via https://github.com/qdrant/landing_page/pull/1226

PR #242, opened 2023-08-17
Clearly note that searches don't include payload/vector data by default

timvisee:
Clearly note that retrieval methods, such as search, don't return payload or vector data by default. Also point to information on how to enable it if desired.

I've seen various user reports on that this is unexpected and that it could be stated better. This attempts to improve that.


davidmyriel (approved):
LGTM

PR #243, opened 2023-08-17
Use QdrantVectorStore instead of GPTQdrantIndex that was removed from LI

kacperlukawski:
This PR changes the example of using LlamaIndex. We need a more comprehensive example, and I'll suggest one in a separate PR.


davidmyriel:
LGTM

PR #284, opened 2023-09-07
document indexed_only

generall:


qdrant-landing/content/documentation/concepts/search.md
```diff
@@ -117,6 +117,7 @@ Currently, it could be:
 
 * `hnsw_ef` - value that specifies `ef` parameter of the HNSW algorithm.
 * `exact` - option to not use the approximate search (ANN). If set to true, the search may run for a long as it performs a full scan to retrieve exact results.
+* `indexed_only` - this option allows to disable search in those segments where vector index is not built yet. This may be useful if you want to minimize the impact to the search performance during the parallel update of the collection. Using this option may lead to a partial result if the collection is not fully indexed yet, consider using it only if eventual consistency is acceptable for your use case.
```
timvisee, line 120: Nitpick, but this may be a _bit_ easier to understand:

```suggestion
* `indexed_only` - this option allows to disable search in those segments where vector index is not built yet. This may be useful if you want to minimize the impact to the search performance whilst the collection is also being updated. Using this option may lead to a partial result if the collection is not fully indexed yet, consider using it only if eventual consistency is acceptable for your use case.
```

timvisee (approved):
Nice!

PR #300, opened 2023-09-11
Articles and docs font

trean:
In articles and docs font changed to `font-size 16px;` and `line-height: 28px;`
Changed font to Roboto on mac.


timvisee:
@davidmyriel @trean 

It looks like this broke monospace fonts for code snippets. See this page as an example: https://qdrant.tech/documentation/concepts/points/

Now:

image

Before (manually modified CSS):

image

I think we want the latter for code snippets.

PR #316, opened 2023-09-21
Add query geo polygon docs

zzzz-vincent:
Change the order of `lon` and `lat` as it is defined in this order in the codebase. This is a very minor issue but it is good  to have everything consistant.
Add GeoPolygon examples to documentation

### DIFF
--- a/qdrant-landing/content/documentation/concepts/filtering.md
+++ b/qdrant-landing/content/documentation/concepts/filtering.md
@@ -846,12 +846,12 @@ Can be applied to [float](../payload/#float) and [integer](../payload/#integer) 
   "key": "location",
   "geo_bounding_box": {
     "bottom_right": {
-      "lat": 52.495862,
-      "lon": 13.455868
+      "lon": 13.455868,
+      "lat": 52.495862
     },
     "top_left": {
-      "lat": 52.520711,
-      "lon": 13.403683
+      "lon": 13.403683,
+      "lat": 52.520711
     }
   }
 }
@@ -862,12 +862,12 @@ models.FieldCondition(
     key="location",
     geo_bounding_box=models.GeoBoundingBox(
         bottom_right=models.GeoPoint(
-            lat=52.495862,
             lon=13.455868,
+            lat=52.495862,
         ),
         top_left=models.GeoPoint(
-            lat=52.520711,
             lon=13.403683,
+            lat=52.520711,
         ),
     ),
 )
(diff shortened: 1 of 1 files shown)


qdrant-landing/content/documentation/concepts/filtering.md
```diff
@@ -908,6 +908,100 @@ It matches with `location`s inside a circle with the `center` at the center and
 If several values are stored, at least one of them should match the condition.
 These conditions can only be applied to payloads that match the [geo-data format](../payload/#geo).
 
+#### Geo Polygon
+
```
agourlay, line 912: It would be good to have a little introduction explaining why it can be used.
I know the other sections do not have this but let's improve.

thesteve0:
```suggestion
Geopolygons search is useful for when you want to find points inside an irregularly shaped area, for example a country boundary or a forest boundary. A polygon always has an exterior ring and may optionally include an interior ring. A lake with an island would be an example of an interior ring. If you wanted to find points in the water but not on the island, you would make an interior ring for the island. 

When defining a ring, you must pick either a clockwise or counterclockwise ordering for your points. 

Currently, we only support unprojected global coordinates (decimal degrees longitude and latitude) and we are datum agnostic

Any points with locations that fall either in or on the boundary of a ring are considered a match .
```

qdrant-landing/content/documentation/concepts/filtering.md
```diff
@@ -908,6 +908,100 @@ It matches with `location`s inside a circle with the `center` at the center and
...
+                            lat=-65.0,
+                    )
+                ]
+            )
+        ]
+    )
+)
+```
+
+It matches with `location`s inside the given polygon's exterior but not inside any interiors.
```
thesteve0, line 1000:
```suggestion
A match is considered any point location inside the given polygon's exterior but not inside any interiors.
```

qdrant-landing/content/documentation/concepts/filtering.md
```diff
@@ -908,6 +908,100 @@ It matches with `location`s inside a circle with the `center` at the center and
...
+
+If several values are stored, at least one of them should match the condition.
```
thesteve0, line 1002:
```suggestion
If several location values are stored for a point, then any of them matching will include that point as a candidate in the resultset. 
```

thesteve0 (commented):
Thanks for fixing the coords and making them consistent!

agourlay (approved):

PR #319, opened 2023-09-22
google fonts upd: Roboto and Roboto Mono

trean:
- fixes _and changes_ code blocks font, now they use Roboto Mono,
- updated fonts, coming from Google Fonts


generall:
Let's undo everything and only change article font to the same as used in documentation.

Use same font size in docs and articles  - 16px

Also change font in article list, so it is not serif.

timvisee:
The monospace font is a bit too thin to my liking:

image

Maybe we can switch to 400. What do you think?

trean:
@timvisee As you can see in the above comment, we decided not to move to the Roboto at all and to change only the articles font and docs and articles font size. This is done in #322. This PR will be closed.

trean:
@generall requested changes are in #322.

PR #331, opened 2023-10-03
Create a collection: `recreate_collection` -> `create_collection`

coszio:
In https://qdrant.tech/documentation/concepts/collections/#create-a-collection, the python example suggests `client.recreate_collection()` which users will take as default. However `recreate` is a client utility for deleting and creating a collection with the same name. It has been a problem for some Discord users who are confused by their collection being deleted.

The utility is useful but I think the introduction to collection creation should not be the place to showcase it.

There are other occurrences of `recreate_collection` in other examples, but I'm not sure if we should change all of them now.

### DIFF
--- a/qdrant-landing/content/documentation/concepts/collections.md
+++ b/qdrant-landing/content/documentation/concepts/collections.md
@@ -48,7 +48,7 @@ from qdrant_client.http import models
 
 client = QdrantClient("localhost", port=6333)
 
-client.recreate_collection(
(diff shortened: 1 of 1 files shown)


timvisee:
Ouch! Definitely better to change this to create.

Should we mention the `recreate` function below it in our Python client?

timvisee (approved):

kacperlukawski (approved):

PR #336, opened 2023-10-06
Add recommendation strategies

coszio:
Refactors almost all of the **Recommend API** section, while adding the new `best_score` recommend strategy

### DIFF
--- a/qdrant-landing/content/documentation/concepts/search.md
+++ b/qdrant-landing/content/documentation/concepts/search.md
@@ -349,26 +349,15 @@ The result of this API contains one array per search requests.
 
 ## Recommendation API
 
-<aside role="alert">Negative vectors is an experimental functionality that is not guaranteed to work with all kind of embeddings.</aside>
-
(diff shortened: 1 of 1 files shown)


qdrant-landing/content/documentation/concepts/search.md
```diff
@@ -423,6 +412,49 @@ Example result of this API would be
...
+avg_positive + avg_positive - avg_negative
+```
+
+In the case of not having any negative examples, the search vector will simply be equal to `avg_positive`.
+
+This is the default strategy that's going to be set implicitly, but you can explicitly define it by setting `"strategy": "average_vector"` in the recommendation request.
+
+### Best score strategy
+
+A new strategy introduced in v1.6, is called `best_score`. It is based on the idea that the best way to find similar vectors is to find the ones that are closer to a positive example, while avoiding the ones that are closer to a negative one.
```
timvisee, line 433: I think I prefer an

```markdown
*Available as of v1.6.0*
```

note here instead.

generall (approved):

PR #338, opened 2023-10-09
JsonPath like in payload selector

agourlay:


qdrant-landing/content/documentation/concepts/search.md
```diff
@@ -237,6 +237,13 @@ client.search(
 )
 ```
 
+It is possible to target nested fields using a dot notation:
+- `payload.nested_field` - for a nested field
+- `payload.nested_array[].sub_field` - for projecting nested fields within an array
```
timvisee, line 242: Should this also mention `include`?

qdrant-landing/content/documentation/concepts/search.md
```diff
@@ -237,6 +237,13 @@ client.search(
...
+
+Accessing array elements by index is currently not supported.
+
+
```
timvisee, line 246:
```suggestion
```

qdrant-landing/content/documentation/concepts/search.md
agourlay: There is a doc about include vs exclude a few lines up.
Is it what you mean?

timvisee: You're right! I quickly skimmed the code, but missed it.

I was also under the assumption that you could use include and exclude together, but it appears I'm wrong on that.

I added a commit to your branch to format `include` and `exclude` as literals. I think that clarifies it a bit. Hope you don't mind :smiley:

timvisee (approved):

PR #342, opened 2023-10-12
Fixed variable name

pratyushmittal:
The variable name used in config is `snapshots_path`. Hence the related ENV variables is `QDRANT__STORAGE__SNAPSHOTS_PATH`.

Using `QDRANT__STORAGE__SNAPSHOT_PATH` doesn't work.


timvisee (approved):
Great catch! Thanks!

PR #343, opened 2023-10-13
Update `with_payload` examples, also describe using an array directly

timvisee:
The `with_payload` field also supports using an array directly, without explicit `include`/`exclude` usage.

For example:

```json
{
    "with_payload": ["city"]
}
```

### DIFF
--- a/qdrant-landing/content/documentation/concepts/search.md
+++ b/qdrant-landing/content/documentation/concepts/search.md
@@ -209,7 +209,33 @@ client.search(
 )
 ```
 
-Parameter `with_payload` might also be used to `include` or `exclude` specific fields only:
+The parameter `with_payload` may also be used to scope to or filter out a
+specific payload subset. You can specify an array of items to include. For
+example, to include `city`:
+
+```http
+POST /collections/{collection_name}/points/search
+
+{
+    "vector": [0.2, 0.1, 0.9, 0.7],
+    "with_payload": ["city"]
+}
+```
+
+```python
+from qdrant_client import QdrantClient
+from qdrant_client.http import models
+
+client = QdrantClient("localhost", port=6333)
+
+client.search(
+    collection_name="{collection_name}",
+    query_vector=[0.2, 0.1, 0.9, 0.7],
+    with_payload=["city"],
+)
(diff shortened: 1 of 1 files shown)


qdrant-landing/content/documentation/concepts/search.md
```diff
@@ -209,7 +209,33 @@ client.search(
...
+The parameter `with_payload` may also be used to scope to or filter out a
+specific payload subset. You can specify an array of items to include. For
+example, to include `city`:
+
+```http
+POST /collections/{collection_name}/points/search
+
+{
+    "vector": [0.2, 0.1, 0.9, 0.7],
+    "with_payload": ["city"]
```
mjang, line 221: Since you suggested that users can specify an array of items (line 213), I'd include an array `with_payload`, maybe something like:

```suggestion
    "with_payload": ["city", "village", "town"]
```

qdrant-landing/content/documentation/concepts/search.md
```diff
@@ -209,7 +209,33 @@ client.search(
 )
 ```
 
-Parameter `with_payload` might also be used to `include` or `exclude` specific fields only:
+The parameter `with_payload` may also be used to scope to or filter out a
+specific payload subset. You can specify an array of items to include. For
+example, to include `city`:
```
mjang, line 214: I'd rewrite the whole page in [active voice](https://developers.google.com/style/voice) to tell make a reader think "This is what I can do". Here's how you can set up this paragraph with active voice:

```suggestion
You can use `with_payload` to scope to or filter a specific payload subset. 
You can even specify an array of items to include. For example, to include `city`, 
`village`, and `town`:
```

mjang (commented):
Two suggestions, based in part on the [Google Developer Documentation Style Guide](https://developers.google.com/style).

timvisee:
> Two suggestions, based in part on the [Google Developer Documentation Style Guide](https://developers.google.com/style).

Good suggestions. Thanks @mjang!

Didn't know about that Google resource, thanks for that as well.

agourlay (approved):

PR #351, opened 2023-10-17
Enable code colouring

kacperlukawski:
I marked all the code snippets as `python`, so the code colouring works.


NirantK:
Ohh this is what I missed! Thanks for this @kacperlukawski !

PR #352, opened 2023-10-17
Add public cosign key

bashofmann:
We are starting to use cosign to sign our container images to increase supply chain security.

To make verification of this signature as easy as
possible, we need to make the public key accessible on a source that users trust, e.g. a company's
homepage.

In the future, a signature can be verified like this

```
cosign verify qdrant/qdrant:v1.6.1 --key https://qdrant.tech/cosign.pub
```


timvisee:
Do we possibly want to host this on a separate subdomain? Hosting it here could make it easy for malicious actors to bump the public key through this repository with one non-careful review.

I suggest to make it available on a manually managed host, to which no CI pipelines are attached.

bashofmann:
I'm fine with hosting it somewhere else. @fabriziobonavita let's discuss this later this week to see how and were we can easily add this.
@timvisee thanks for the quick feedback.

timvisee:
> I'm fine with hosting it somewhere else. @fabriziobonavita let's discuss this later this week to see how and were we can easily add this.

Note that we can enforce ownership of the key to specific users with [CODEOWNERS](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners), but I still think hosting it on a dedicated subdomain is better.

> @timvisee thanks for the quick feedback.

👍

PR #357, opened 2023-10-19
Simplify asynchronous API tutorial with the new AsyncQdrantClient

kacperlukawski:
This PR removes a tutorial using autogenerated clients and keeps the info about using Qdrant asynchronously mostly for SEO purposes.


joein:
if it is kept for seo purposes than i think it's good to go

maybe we could keep some trivial upsert + search scenario though

kacperlukawski:
@joein Thanks! I added two more calls. I'm merging the changes.

PR #360, opened 2023-10-20
Update Pinecone starter plan limits

kacperlukawski:
Pinecone has changed its policy for the starter plan. They reduced the capacity to around 100k OpenAI ADA embeddings.

See: 
1. https://www.pinecone.io/blog/updated-free-plan/
2. https://docs.pinecone.io/docs/indexes#starter-plan


joein:
Shouldn't we just write that they support 200k 768? (they've mentioned that it is proportional for other dimensionalities)

kacperlukawski:
> Shouldn't we just write that they support 200k 768? (they've mentioned that it is proportional for other dimensionalities)

Yeah, we probably could. Let me rephrase it

joein (approved):

PR #384, opened 2023-10-31
feat: Add case studies link to website footer

KShivendu:
This should help with making it more visible. I almost couldn't find it.


davidmyriel:
outdated due to new website

PR #387, opened 2023-11-01
Improve documentation for updating collection vector parameters

timvisee:
We've seen multiple users having issues with updating vector parameters in a collection that does not use named vectors. The endpoint for this always requires users to specify a vector name, unlike other endpoints. This was noted in the documentation, but it was quite hidden.

This:
- Restructures the `Update collection parameters` section
- Adds a `Update vector parameters` subsection
- Adds the 'use empty vector name' note at the top in a clear blue box
- Adds two examples for setting `on_disk: true` for a collection with and without named vectors.

All of this should make it a lot more clear on how to update vector parameters.

image


qdrant-landing/themes/qdrant/static/css/blocks/_documentation.scss
```diff
@@ -29,15 +29,15 @@
...
 
     h5 {
-      font-size: 1.25rem;
+      font-size: 1rem;
     }
 
     h6 {
-      font-size: 1.25rem;
+      font-size: 1rem;
     }
```
timvisee, line 41: I've updated the H4/H5/H6 sizes in our documentation. Otherwise the subsection I've added has the same font size as its parent which is quite confusing.

The alternative was to use just regular **bold** text as a header. But I like using an H4 with the above change better.

PR #388, opened 2023-11-02
Add jina embeddings to integrations

JohannesMessner:
Hey Qdrant team!

We just launched [Jina Embeddings](https://jina.ai/#enterprises) and would love to add them as an integration to your docs.

We also plan to make our models available through FastEmbed, so stay tuned for that PR as well.

Cheers!

### DIFF
--- a/qdrant-landing/content/documentation/integrations/jina-embeddings.md
+++ b/qdrant-landing/content/documentation/integrations/jina-embeddings.md
@@ -1,0 +1,59 @@
+---
+title: Jina Embeddings
+weight: 800
+---
+
+# Jina Embeddings
+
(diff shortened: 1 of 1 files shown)


timvisee:
Cool! Thank you for the effort of adding this.

timvisee (approved):

JohannesMessner:
Thanks for the review! I am marking it as draft again because our landing page that we want to link isn't up yet, should be done today. So I'd like to wait for that in order to avoid an immediate follow-up PR.

timvisee (approved):

kacperlukawski (approved):
LGTM, thanks @JohannesMessner!

PR #406, opened 2023-11-14
Rephrase sections of the distributed deployment page

mac-chaffee:
Tim and I wanted to improve this section of the docs since we had a little trouble understanding it initially (it is a very complex topic!) I ended up reviewing the whole page in the process and made the following changes:

* Changed the phrasing of some sentences to be a little clearer, including adding missing articles and changing tense.
* In the "sharding" section, added some extra details about Qdrant not automatically rebalancing shards
* Changed "Scaling replication factor" to "Creating new shard replicas" to point out that changing replication_factor won't create new replicas; you have to perform a replicate_shard operation.
* Capitalized Qdrant
* Trimmed trailing spaces

### DIFF
--- a/qdrant-landing/content/documentation/guides/distributed_deployment.md
+++ b/qdrant-landing/content/documentation/guides/distributed_deployment.md
@@ -31,11 +31,10 @@ cluster:
     tick_period_ms: 100
 ```
 
-With default configuration, Qdrant will use port `6335` for its internal communication.
+By default, Qdrant will use port `6335` for its internal communication.
 All peers should be accessible on this port from within the cluster, but make sure to isolate this port from outside access, as it might be used to perform write operations.
 
-Additionally, the first peer of the cluster should be provided with its URL, so it could tell other nodes how it should be reached.
-Use the `uri` CLI argument to provide the URL to the peer:
+Additionally, you must provide the `--uri` flag to the first peer so it can tell other nodes how it should be reached:
 
 ```bash
 ./qdrant --uri 'http://qdrant_node_1:6335'
@@ -50,7 +49,7 @@ To do this, they need to be provided with a bootstrap URL:
 ```
 
 The URL of the new peers themselves will be calculated automatically from the IP address of their request.
-But it is also possible to provide them individually using `--uri` argument.
+But it is also possible to provide them individually using the `--uri` argument.
 
 ```text
 USAGE:
@@ -63,9 +62,9 @@ OPTIONS:
 
         --uri <URI>
             Uri of this peer. Other peers should be able to reach it by this uri.
-            
+
             This value has to be supplied if this is the first peer in a new deployment.
-            
+
             In case this is not the first peer and it bootstraps the value is optional. If not
             supplied then qdrant will take internal grpc port from config and derive the IP address
             of this peer on bootstrap peer (receiving side)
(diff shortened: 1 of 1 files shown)


qdrant-landing/content/documentation/guides/distributed_deployment.md
```diff
@@ -198,27 +194,29 @@ POST /collections/{collection_name}/cluster
...
 
-Shard replication increases the reliability of the cluster by keeping several copies of a shard spread among the cluster.
-This ensure the availability of the shards in case of node failures, except if all replicas are lost.
+### Replication factor
 
-By default, all the shards in a cluster have a replication factor of one, meaning no additional copy is maintained.
+When you create a collection, you can control how many shard replicas you'd like to store by changing the `replication_factor`. By default, `replication_factor` is set to "1", meaning no additional copy is maintained automatically. You can change that by setting the `replication_factor` when you create a collection.
 
-The replication factor of a collection can be configured at creation time.
+The replication factor of a collection can only be configured at creation time.
```
elbart, line 219: I think, here we need more detail. In fact, you can change the replication_factor at runtime via the PATCH collection API, and it does reflect the change when fetching the collection config afterwards. However it has no effect on the actual amount of shard replicas in the running collection. Users need to adjust the actual replicas to the new replication factor themselves. Or is it better to deny such an update in the API to avoid confusion for now? WDYT @generall @agourlay @timvisee

qdrant-landing/content/documentation/guides/distributed_deployment.md
```diff
@@ -171,16 +169,14 @@ client.createCollection("{collection_name}", {
 });
 ```
 
-We recommend selecting the number of shards as a factor of the number of nodes you are currently running in your cluster.
+We recommend setting the number of shards to be a multiple of the number of nodes you are currently running in your cluster.
 For example, if you have 3 nodes, 6 shards could be a good option.
 
-### Cluster scaling
+Shards are evenly distributed across all existing nodes when a collection is first created, but Qdrant does not automatically rebalance shards if your cluster size changes (since this is an expensive operation on large clusters). See the next section for how to move shards after scaling operations.
```
timvisee, line 175:
```suggestion
Shards are evenly distributed across all existing nodes when a collection is first created, but Qdrant does not automatically rebalance shards if your cluster size or replication factor changes (since this is an expensive operation on large clusters). See the next section for how to move shards after scaling operations.
```

qdrant-landing/content/documentation/guides/distributed_deployment.md
```diff
@@ -171,16 +169,14 @@ client.createCollection("{collection_name}", {
...
 
-If you want to extend your cluster with new nodes or some nodes become slower than the others, it might be helpful to re-balance the shard distribution in the cluster.
+### Moving shards
 
-*As of v0.9.0:* Qdrant allows moving shards between nodes in the cluster and removing nodes from the cluster.
-
-This functionality unlocks the ability to dynamically scale the cluster size without downtime.
+*As of v0.9.0:* Qdrant allows moving shards between nodes in the cluster and removing nodes from the cluster. This functionality unlocks the ability to dynamically scale the cluster size without downtime. It also allows you to upgrade or migrate nodes without downtime.
```
timvisee, line 179: We format it this way in other places. I know you didn't change this, but I guess we can make it consistent as part of this PR as well now.

```suggestion
*Available as of v0.9.0

Qdrant allows moving shards between nodes in the cluster and removing nodes from the cluster. This functionality unlocks the ability to dynamically scale the cluster size without downtime. It also allows you to upgrade or migrate nodes without downtime.
```

qdrant-landing/content/documentation/guides/distributed_deployment.md
```diff
@@ -198,27 +194,29 @@ POST /collections/{collection_name}/cluster
...
 DELETE /cluster/peer/{peer_id}
 ```
 
-After that, Qdrant will exclude the node from the consensus, and the instance will be ready for the shutdown.
+After that, Qdrant will exclude the node from the consensus, and the instance will be ready for shutdown.
 
 ## Replication
 
-*As of v0.11.0:* Qdrant allows to replicate shards between nodes in the cluster.
+*As of v0.11.0:* Qdrant allows you to replicate shards between nodes in the cluster.
```
timvisee, line 210:
```suggestion
*Available as of v0.11.0*

Qdrant allows you to replicate shards between nodes in the cluster.
```

qdrant-landing/content/documentation/guides/distributed_deployment.md
timvisee: Yeah it is probably best to reject updates that change this parameter now.

Eventually we do want to support this though.

mac-chaffee: I reworded to "Currently, the replication factor of a collection can only be configured at creation time" to leave room for the future 👍

PR #408, opened 2023-11-15
Add Discovery API documentation

coszio:
Adds the `Discovery API` section to `search.md`

Pending: 
- [x] add images


qdrant-landing/content/documentation/concepts/search.md
```diff
@@ -767,6 +775,205 @@ The result of this API contains one array per recommendation requests.
 }
 ```
 
+## Discovery API
+
+*Available as of v1.7*
+
+REST API Schema definition available [here](https://qdrant.github.io/qdrant/redoc/index.html#tag/points/operation/discover_points)
+
+In this api, Qdrant introduces the concept of `context`, which is used for splitting the space, so that the space in which you search becomes constrained by the vectors you provide in the context. The interface for providing them is in vector examples (ids or raw vectors), just like in the recommendation API, but in this case, they need to be provided in the form of positive-negative pairs.
```
kacperlukawski, line 784:
```suggestion
In this API, Qdrant introduces the concept of `context`, which is used for splitting the space. Context is a set of positive-negative pairs. Each pair divides the space into positive and negative zones. In that mode, the search operation prefers points based on how many positive zones they belong to.

The interface for providing context is similar to the recommendation API (ids or raw vectors). Still, in this case, they need to be provided in the form of positive-negative pairs. 
```

qdrant-landing/content/documentation/concepts/search.md
```diff
@@ -767,6 +775,205 @@ The result of this API contains one array per recommendation requests.
...
+
+We can directly associate the score function to a loss function, where 0.0 is the maximum score a point can have, which means it is only in positive areas. As soon as a point exists closer to a negative example, its loss will simply be the difference of the positive and negative similarities.
+
+$$
+\text{context score} = \sum min(s(v^+_i) - s(v^-_i), 0.0)
+$$
+
+Where $v^+_i$ and $v^-_i$ are the positive and negative examples of each pair, and $s(v)$ is the similarity function.
+
+Using this kind of search, you can expect the output to not necessarily be around a single point, but rather, to be any point that isn’t closer to a negative example, which creates a constrained diverse result. So, even when the api is not called [`recommend`](#recommendation-api), recommendation systems can also use this approach and adapt it for their specific use-cases.
```
kacperlukawski, line 903:
```suggestion
Using this kind of search, you can expect the output to not necessarily be around a single point, but rather, to be any point that isn’t closer to a negative example, which creates a constrained diverse result. So, even when the API is not called [`recommend`](#recommendation-api), recommendation systems can also use this approach and adapt it for their specific use-cases.
```

kacperlukawski (changes requested):
Overall, it's a clear description. I suggested a small adjustment to the initial description (about the context).

qdrant-landing/content/documentation/concepts/search.md
coszio: Thanks! Applied suggestion

qdrant-landing/content/documentation/concepts/search.md
coszio: Thanks! Applied suggestion

coszio:
As per @generall request (which I agree), I moved recommend and discover to their own standalone page, I called it `Explore` so that we can later also include `diversity` there, but other suggestions are welcome.

generall:
ToDo:

- re-draw pictures using our style guidelines
- generate preview image for the new section

generall (approved):

timvisee (approved):

PR #417, opened 2023-11-21
Multitenancy: add link to authoritative definition

mjang:
Not everyone knows what's multitenancy -- and it's used in different contents (I'm used to the definition in Identity Management).

I think it'd help readers to have a link to an authoritative reference, so I propose adding one from IBM, https://www.ibm.com/topics/multi-tenant.

### DIFF
--- a/qdrant-landing/content/documentation/concepts/collections.md
+++ b/qdrant-landing/content/documentation/concepts/collections.md
@@ -25,7 +25,7 @@ These settings can be changed at any time by a corresponding request.
 
 ## Setting up multitenancy
 
-**How many collections should you create?** In most cases, you should only use a single collection with payload-based partitioning. This approach is called [multitenancy](https://www.ibm.com/topics/multi-tenant). It is efficient for most of users, but it requires additional configuration. [Learn how to set it up](../../tutorials/multiple-partitions/)
+**How many collections should you create?** In most cases, you should only use a single collection with payload-based partitioning. This approach is called [multitenancy](https://en.wikipedia.org/wiki/Multitenancy). It is efficient for most of users, but it requires additional configuration. [Learn how to set it up](../../tutorials/multiple-partitions/)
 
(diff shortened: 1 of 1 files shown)


qdrant-landing/content/documentation/concepts/collections.md
```diff
@@ -25,7 +25,7 @@ These settings can be changed at any time by a corresponding request.
 
 ## Setting up multitenancy
 
-**How many collections should you create?** In most cases, you should only use a single collection with payload-based partitioning. This approach is called multitenancy. It is efficient for most of users, but it requires additional configuration. [Learn how to set it up](../../tutorials/multiple-partitions/)
+**How many collections should you create?** In most cases, you should only use a single collection with payload-based partitioning. This approach is called [multitenancy](https://www.ibm.com/topics/multi-tenant). It is efficient for most of users, but it requires additional configuration. [Learn how to set it up](../../tutorials/multiple-partitions/)
```
mjang, line 28:
```suggestion
**How many collections should you create?** In most cases, you should only use a single collection with payload-based partitioning. This approach is called [multitenancy](https://en.wikipedia.org/wiki/Multitenancy). It is efficient for most of users, but it requires additional configuration. [Learn how to set it up](../../tutorials/multiple-partitions/)
```

mjang:
> I don't think it is something specific to IBM, so I'd prefer to link to Wikipedia:
> 
> https://en.wikipedia.org/wiki/Multitenancy

In the past, I've been told to avoid Wikipedia, since its content can change. But if that's the Qdrant preference, I'm good with it, and have modified the PR, per https://github.com/qdrant/landing_page/pull/417/files#r1401220948

timvisee:
> In the past, I've been told to avoid Wikipedia, since its content can change. But if that's the Qdrant preference, I'm good with it, and have modified the PR.

IBMs pages can change as well.

But I think that generally speaking, if content changes on Wikipedia pages like this, it's for the better.

I agree that Wikipedia is a good enough resource here.

timvisee (approved):

PR #418, opened 2023-11-22
ID admin tools, correct grammar

mjang:
I've modified the intro to:

- Help readers ID the admin tools at a glance. Readers who already know how to set up `Locking` and `Recovery mode` won't need to go further. Readers who don't know will have links to both.
- Correct grammar.
- Capitalize "Qdrant"

IMO, the whole page could use a good rewrite, but one step at a time.

### DIFF
--- a/qdrant-landing/content/documentation/guides/administration.md
+++ b/qdrant-landing/content/documentation/guides/administration.md
@@ -7,7 +7,12 @@ aliases:
 
 # Administration
 
-Qdrant exposes administration tools which enable to modify at runtime the behavior of a qdrant instance without changing its configuration manually.
+Qdrant exposes administration tools which allow you to modify the behavior of a Qdrant instance at runtime:
+
+- [Locking](#locking)
+- [Recovery mode](#recovery-mode)
+
+You do not have to change the instance configuration manually.
 
 ## Locking
 


qdrant-landing/content/documentation/guides/administration.md
```diff
@@ -7,7 +7,12 @@ aliases:
 
 # Administration
 
-Qdrant exposes administration tools which enable to modify at runtime the behavior of a qdrant instance without changing its configuration manually.
+Qdrant exposes administration tools which allow you to modify the behavior of a Qdrant instance at runtime:
+
+- [Locking](#locking)
+- [Recovery mode](#recovery-mode)
```
timvisee, line 13: Though I'm fine with listing these here, we also show them on the right in the table of contents.

qdrant-landing/content/documentation/guides/administration.md
```diff
@@ -7,7 +7,12 @@ aliases:
 
 # Administration
 
-Qdrant exposes administration tools which enable to modify at runtime the behavior of a qdrant instance without changing its configuration manually.
+Qdrant exposes administration tools which allow you to modify the behavior of a Qdrant instance at runtime:
```
timvisee, line 10: Recovery mode is not a runtime thing, so the original text was wrong.

What do you think about simplifying it to this?

```suggestion
Qdrant exposes administration tools:
```

mjang: > Recovery mode is not a runtime thing, so the original text was wrong.

That makes sense. Recovery mode (like runlevel 1) and runtime are two different concepts. While normally I dislike extra words:

```suggestion
Qdrant exposes the following administration tools:
```

mjang: This is not the only page with Qdrant admin tools. I'd either change the title of this page, or maybe:

- Move "locking" next to [Authentication](https://qdrant.tech/documentation/guides/security/#authentication). 
   - Even though "locking" is global, it's something that an admin would disable before others can create / add new data.
- Move "recovery mode" under [Troubleshooting](https://qdrant.tech/documentation/guides/common-errors/).
   - While the page title is "Solving common errors," the ToC entry is "Troubleshooting"

I'm tempted to replace this PR with a "grammar only" change, and set up a new issue to organize _all_ Qdrant admin tools.

I'll create a new issue, and edit this comment with a link.

Issue #424, opened 2023-11-26
doc: proposed update to Semantic Search 101 Tutorial


mjang:
While I like the [Semantic Search 101](https://qdrant.tech/documentation/tutorials/search-beginners/) page, it makes a few undocumented assumptions:

- The YouTube is run in an IDE
   - I think many users prefer the command line
   - If Qdrant prefers the IDE, the doc should state that explicitly (and then update this issue)
- The [linked tutorial](https://towardsdatascience.com/creating-and-using-virtual-environment-on-jupyter-notebook-with-python-db3f5afdd56a) is from Medium -- which is not available unless you have a subscription

My assumption: readers know how to work from the bash shell. I'd like to propose some changes to help readers run the steps in Semantic Search 101 from the (bash) command line.

- Update the YouTube to work from bash
- Specify that the commands are run from bash
- Change the link from a Medium article to Python docs, specifically [Creating Virtual Environments](https://docs.python.org/3/tutorial/venv.html#creating-virtual-environments) 
- Include prompts for when the reader needs to:
  - Activate the virtual environment
  - Activate the Python prompt (`>>>`)

After the tutorial is complete, add a section to help the reader "clean up" by showing how:
- To exit the Python prompt (Ctrl+D)
- Deactivate the virtual environment (`deactivate`)

timvisee:
Since <https://github.com/qdrant/landing_page/pull/425> is merged now, can we close this as well?

Or are there more outstanding changes you'd like to see applied.

PR #425, opened 2023-11-26
docs: Modify the Semantic Search 101 tutorial to work from bash

mjang:
Based on https://github.com/qdrant/landing_page/issues/424 

As published today, the Semantic Search 101 tutorial includes assumptions which I think may confuse readers who don't know Python. To set up a Python virtual environment, it also includes a link to a Medium article; I suggest replacing it with a link to the Python docs.

I'm also suggesting a few more instructions, to help readers know when to:

- Activate a virtual environment
- Set up a Python prompt

And to "clean up" afterwards, to:

- Exit the Python prompt
- Deactivate the virtual environment

I realize this doesn't quite work with the current video -- but the current doc doesn't explain how the video was set up (I think in GitHub desktop). 

However, if reviewers want a replacement video, I can set up an equivalent screenshare, which switches back and forth between the doc and the bash shell. (I've recorded one which I'm not completely happy with -- it's 7 minutes, and some of my descriptions don't feel sufficiently precise -- [link to video in G drive](https://drive.google.com/file/d/1BpvneX2n0O-ELL5z_-dqGzTaDLYMku2T/view?usp=sharing).)


timvisee (approved):
Good additions, thanks!

Also, good catch on those missing links.

PR #428, opened 2023-11-28
Document snapshot priority

timvisee:
Document the snapshot priority parameter, link to it from the recovery section and give an example of use.

This is an important parameter which can gave a significant effect on the end result of recovery. We didn't document it. This PR now describes it.

Rendered section: https://deploy-preview-428--condescending-goldwasser-91acf0.netlify.app/documentation/concepts/snapshots/#snapshot-priority

image

I've also added an info box linking to the new section here: https://deploy-preview-428--condescending-goldwasser-91acf0.netlify.app/documentation/concepts/snapshots/#recover-via-api

image


generall (approved):

agourlay (approved):
Thank you very much :+1:

PR #429, opened 2023-11-28
Remove empty lines in HTTP examples, reformat some of them

timvisee:
This removes the empty line in HTTP examples between the method/path and body.

I've done this because our Web UI doesn't support this formatting:

image


Anush008 (approved):
🫡

coszio:
Another rule candidate if we ever introduce `vale`

PR #432, opened 2023-11-29
Update aleph-alpha-search.md to take into account for deprecated func…

wolfgangihloff:
Update aleph-alpha-search.md to take into account for deprecated functions both in Aleph Alpha and Qdrant.
Thanks for the tutorial, should be easier for the next person coming around.


generall (approved):

generall:
Hi @wolfgangihloff! Thanks for the update!

PR #439, opened 2023-12-04
Fix fourth ToC link not working

timvisee:
The fourth table of contents link was incorrect. This PR fixes that.


kacperlukawski (approved):
Good catch!

PR #445, opened 2023-12-07
Sparse vectors documentation

agourlay:
This PR adds documentation for sparse vectors.

I am a bit unsure about adding a dedicated concept page because it is a crosscutting aspect through collections, search, update.


timvisee:
> I am a bit unsure about adding a dedicated concept page because it is a crosscutting aspect through collections, search, update.

Good point. I think it makes sense to have it spread over multiple pages.

Maybe we can add a dedicated page for it people keep asking questions.

timvisee (approved):

PR #446, opened 2023-12-07
Add user-defined sharding section

coszio:


qdrant-landing/content/documentation/guides/distributed_deployment.md
```diff
@@ -230,6 +235,153 @@ DELETE /cluster/peer/{peer_id}
...
+{
+    "shard_number": 1,
+    "sharding_method": "custom"
+    "replication_factor": 2
+}
+```
+
+Then you will have `1 * 10 * 2 = 20` total physical shards in the collection.
+
+To specify the shard for each point, you need to provide the `shard_key` field in the point payload:
```
timvisee, line 313: I think we better not name it '* payload' anywhere as the shard key isn't actually in payload data. It could be confusing for users.

```suggestion
To specify the shard for each point, you need to provide the `shard_key` field in the upsert request:
```

qdrant-landing/content/documentation/guides/distributed_deployment.md
```diff
@@ -230,6 +235,153 @@ DELETE /cluster/peer/{peer_id}
...
+        None,
+    )
+    .await?;
+```
+
+<aside role="alert">
+When using custom sharding, IDs are only enforced to be unique within a shard key. This means that you can have multiple points with the same ID, if they have different shard keys.
+
+This is a limitation of the current implementation, and is an anti-pattern that should be avoided because it can create scenarios of points with the same ID to have different contents. In the future, we plan to add a global ID uniqueness check.
+</aside>
```
timvisee, line 383: Not sure about this one.

Maybe we should just state:

> Using the same point ID across multiple shard keys is not supported and should be avoided.

qdrant-landing/content/documentation/guides/distributed_deployment.md
coszio: I wondered why copilot kept putting it inside the payload! I wrote the wrong thing here

qdrant-landing/content/documentation/guides/distributed_deployment.md
coszio: Added your concise description first 👍

qdrant-landing/content/documentation/guides/distributed_deployment.md
```diff
@@ -230,6 +235,161 @@ DELETE /cluster/peer/{peer_id}
...
+```
+
+```rust
+
+use qdrant_client::qdrant::{PointStruct, WriteOrdering, WriteOrderingType};
+
+client
+    .upsert_points_blocking(
+        "{collection_name}",
+        Some(vec![shard_key::Key::String("user_1".to_string())]),
```
generall, line 366: Into?

PR #453, opened 2023-12-08
fix: fix python examples on distributed deployment page

joein:


coszio (approved):
Thanks for the corrections!

PR #454, opened 2023-12-08
new: add upload_collection and upload_points

joein:


timvisee (approved):

joein:
It should be updated since `upload_records` has been deprecated in `qdrant-client==1.7.1`

joein:
I believe @coszio mentioned some time ago that we don't need to have 2 examples of record-oriented formats, I find this point fair, but I become a little bit sad when I delete `upload_points` examples..

I've reduced the number of examples in `upload_points` example, maybe it's useful to have them? 
If you don't agree, I will delete them... 🥲

coszio:
I have a bad memory for these things, but these changes seem fine to me

coszio (approved):

qdrant-landing/content/documentation/concepts/points.md
```diff
@@ -393,15 +393,65 @@ client
     .await?;
 ```
 
-<!--
```
mjang, line 396: Basic questions:

- by "everything in the memory", I'm guessing that you're referring to RAM
- I don't see any `upload_collection` or `upload_points` methods in our docs or API (if we have them, I'd want to link to them)
- Do we have a Qdrant `upsert API`, or is this a generic version?
- I see parallelization and retry support in your code samples. But I don't see an example of code that enables lazy batching support

qdrant-landing/content/documentation/concepts/points.md
```diff
@@ -393,15 +393,65 @@ client
...
 
-<!-- 
 
-The Python client has additional features for loading points.
-These include parallel loading, and also loading directly from a numpy file.
+The Python client has additional features for loading points. 
+These include parallelization, a retry mechanism and lazy batching support. 
+For example, you can read your data directly from disks and avoid having everything in the memory.
+These features are incorporated into the `upload_collection` and `upload_points` methods.
+Similar to the basic upsert API, these methods support both record-oriented and column-oriented formats.
```
mjang, line 401:
```suggestion
The Python client has additional features for loading points, which include:

- Parallelization
- A retry mechanism
- Lazy batching support

For example, you can read your data directly from hard drives, to avoid storing all data in RAM. You can use these
features with the `upload_collection` and `upload_points` methods.
Similar to the basic upsert API, these methods support both record-oriented and column-oriented formats.
```

mjang (commented):
Thanks for the look @joein ! 

I've listed some questions in my first comment. I think it'd help our readers if we included links. (Or do we need docs for details like `upload_collection` and `upload_points`?)

qdrant-landing/content/documentation/concepts/points.md
joein: > by "everything in the memory", I'm guessing that you're referring to RAM
yes

> I don't see any upload_collection or upload_points methods in our docs or API (if we have them, I'd want to link to them)

These are python-only methods provided for the convenience (they are wrappers around basic functionality)

> Do we have a Qdrant upsert API, or is this a generic version?

We sometimes refer to upload points as upsert
https://qdrant.tech/documentation/concepts/points/#upload-points

 > I see parallelization and retry support in your code samples. But I don't see an example of code that enables lazy batching support

There are lazy structures in python, e.g. generators ([example article](https://realpython.com/introduction-to-python-generators/), they do not load everything in memory from the very beginning, but load each value on demand (e.g. compute it or read from a disk). One can use them to read large files which don't fit into RAM. 

There are a set of operations which uploads the whole structure to RAM, we don't use them. We read everything iteratively, construct a batch and send it. So the number of points which are loaded into RAM simultaneously is limited and users can upload large files without an effort.

PR #462, opened 2023-12-13
Add Binary Q

NirantK:


NirantK:
Merging this since @generall also confirmed that the numbers look plausible + we posted this on social. Can revert or edit later.

PR #499, opened 2024-01-05
Update qdrant-alternatives.md

TarekSw:
You mean use Qdrant tools to migrate from pinecone to Qdrant


kacperlukawski (approved):
Great catch! Thanks!

PR #505, opened 2024-01-07
add hybrid saas form

generall:


bashofmann:
How about adding these optional fields:

* Approximate amount of Qdrant clusters you would like to run
* Approximate amount of vectors you want to store with which dimension
* Target infrastructure

qdrant-landing/content/surveys/hybrid-saas.md
```diff
@@ -0,0 +1,57 @@
...
+    name: experienced
+  - id: 5
+    label: Please describe the approximate size of the deployment you are planning to use. (optional)
+    placeholder: 3 machines 64GB RAM each, or a deployment capable of serving 100M OpenAI embeddings
+    type: text
+    name: clusterSize
+    required: False
+  - id: 6
+    label: What's your target infrastructure?
+    type: radio
```
bashofmann, line 50: Can we keep this a free-text field? I assume that a good chunk of users will fall on the other category and it would be great to already capture this information.

generall: we don't have ready component for that

Issue #374, opened 2023-10-26
Links in ToC sidebar do not set fragment


timvisee:
On documentation pages we have a table of contents sidebar on the right.

Example page: https://qdrant.tech/documentation/concepts/collections/

When you click one of the table of content items, the webpage does navigate to it. But the URL is not updated to include the header ID fragment. I believe this is something that should happen.

Currently, the URL remains: `.../documentation/concepts/collections/`
Instead, the URL should be updated to: `.../documentation/concepts/collections/#update-collection-parameters`

This did work before. I think this problem is introduced since merging <https://github.com/qdrant/landing_page/pull/363>.

timvisee:
@trean Late to the party. But thanks for fixing this!

PR #510, opened 2024-01-08
Added smooth scroll to anchors inside blog and articles

trean:
Added smooth scroll to anchors inside blog and articles and fixed problem with lacking offset on scroll


trean:
> Transition works, but when I click the link, the `#...` part is not preserved in the link

Fixed here and for #374

PR #514, opened 2024-01-09
Tutorial on measuring retrieval quality

kacperlukawski:
This tutorial covers evaluating the approximation quality using the exact search mode. It does not focus on the embedding models but only the HNSW precision.


kacperlukawski:
Preview: https://deploy-preview-514--condescending-goldwasser-91acf0.netlify.app/documentation/tutorials/retrieval-quality/

qdrant-landing/content/documentation/tutorials/retrieval-quality.md
```diff
@@ -0,0 +1,231 @@
...
+for the testing.
+
+```python
+dataset_iterator = iter(dataset)
+train_dataset = [next(dataset_iterator) for _ in range(10000)]
+test_dataset = [next(dataset_iterator) for _ in range(1000)]
+```
+
+Now, let's create a collection and index the training data. This collection won't be created with the default configuration, as we want to
+ensure the approximate search is performed. Therefore, we will set the `indexing_threshold` to `1`, which means that the indexing will be
```
atarora, line 80: 'indexing threshold' isn't too clear for a concept plus no references are found on the indexing page (which is provided as the reference) .

qdrant-landing/content/documentation/tutorials/retrieval-quality.md
```diff
@@ -0,0 +1,231 @@
...
+performed immediately after the first 1kB of vectors is added to the collection. The indexing process will build the HNSW graph required
```
atarora, line 81:
```suggestion
performed immediately after the first vector is added to the collection. The indexing process will build the HNSW graph required
```

qdrant-landing/content/documentation/tutorials/retrieval-quality.md
generall: I would agree, that setting `indexing_threshold: 1` is usually not the best practice. Optimizer will be triggered on each insertion.

I would propose to just use default value

qdrant-landing/content/documentation/tutorials/retrieval-quality.md
```diff
@@ -0,0 +1,231 @@
...
+for approximate search (Qdrant performs the exact search if there is no HNSW graph for that segment). We will also set the `default_segment_number`
```
generall, line 82: same for `default_segment_number`. It doesn't look to relevant for this tutorial.

qdrant-landing/content/documentation/tutorials/retrieval-quality.md
generall: Technically it is 1kB of vectors, but I would suggest to drop the mentions of that, as it looks like the tutorial should work just fine without the explicit configuration.

generall: And if it is not indexing, you can try to use more data

kacperlukawski: Okay, the updated version uses the defaults.

qdrant-landing/content/documentation/tutorials/retrieval-quality.md
kacperlukawski: Okay, the updated version uses the defaults.

qdrant-landing/content/documentation/tutorials/retrieval-quality.md
```diff
@@ -0,0 +1,221 @@
...
+
+```python
+from qdrant_client import QdrantClient, models
+
+client = QdrantClient("http://localhost:6333")
+client.create_collection(
+    collection_name="arxiv-titles-instructorxl-embeddings",
+    vectors_config=models.VectorParams(
+        size=768,  # Size of the embeddings generated by InstructorXL model
+        distance=models.Distance.COSINE,
```
atarora, line 91: It would be nice for someone to understand how different distance algorithms affect search results, I am sure, that falls under the measure of retrieval quality too.

kacperlukawski: Good idea, thanks! How about making a note about this?

atarora: Looks good ! Thanks !

qdrant-landing/content/documentation/tutorials/retrieval-quality.md
```diff
@@ -0,0 +1,227 @@
...
+```
+
+Response:
+
+```text
+avg(precision@5) = 0.9969999999999998
+```
+
+The precision has obviously increased, and we know how to control it. However, there is a trade-off between the precision and the search
+latency and memory requirements. In some specific cases, we might want to increase the precision as much as possible, so now we know how
```
atarora, line 216:
```suggestion
latency and memory requirements. In some specific cases, we may want to increase the precision as much as possible, so now we know how
```

qdrant-landing/content/documentation/tutorials/retrieval-quality.md
```diff
@@ -0,0 +1,227 @@
...
+to do it. 
+
+## Wrapping up
+
+Retrieval quality is an important factor of the semantic search quality. It might and should be measured if you care about the quality of
```
atarora, line 221:
```suggestion
Assessing the quality of retrieval is a critical aspect of evaluating semantic search performance. It is imperative to measure retrieval quality when aiming for optimal quality of.
```

qdrant-landing/content/documentation/tutorials/retrieval-quality.md
```diff
@@ -0,0 +1,227 @@
...
+your search results. Qdrant provides a built-in exact search mode, which can be used to measure the quality of the ANN algorithm itself, 
+even in an automated way, as part of your CI/CD pipeline.
+
+Again, **the quality of the embeddings is the most important factor**. HNSW does a pretty good job in terms of precision, but it is
```
atarora, line 225:
```suggestion
Again, **the quality of the embeddings is the most important factor**. HNSW does a pretty good job in terms of precision, and it is
```

qdrant-landing/content/documentation/tutorials/retrieval-quality.md
```diff
@@ -0,0 +1,227 @@
...
+parameterizable and tunable, when required. There are some other ANN algorithms available out there, such as IVF, but they usually [perform
```
atarora, line 226: Can you add link for IVF , its bit of polysemous! Thanks

atarora (changes requested):
Very informative article ! Great work @kacperlukawski !

atarora (approved):

PR #520, opened 2024-01-12
Documentation for parameterized integer index

timvisee:
Documentation for Qdrant 1.8.

This documents the parameterized integer index added in <https://github.com/qdrant/qdrant/pull/3380>.

It is a specialized integer index type, that some might want to use for further fine-tuning. Generally speaking it is recommended to use the standard integer index without parameters.

Rendered: https://deploy-preview-520--condescending-goldwasser-91acf0.netlify.app/documentation/concepts/indexing/#integer-index-parameters

### DIFF
--- a/qdrant-landing/content/documentation/concepts/indexing.md
+++ b/qdrant-landing/content/documentation/concepts/indexing.md
@@ -185,22 +185,40 @@ Available tokenizers are:
 
 See [Full Text match](../filtering/#full-text-match) for examples of querying with full-text index.
 
-### Integer index parameters
+### Parameterized index
 
 *Available as of v1.8.0*
 
-The `integer` index also has a parameterized variant which allows fine-tuning
-indexing and search performance.
+We've added a parameterized variant to the `integer` index, which allows
+you to fine-tune indexing and search performance.
 
-Generally speaking, we recommend using the standard `integer` index variant
-above without parameters or fine-tuning. Accidentally configuring this
-incorrectly will have a significant effect on performance.
+Both the regular and parameterized `integer` indexes use the following flags:
 
-The following flags can be set to build an optimized index:
-- `lookup` - `true` to enable support for direct lookup ([Match](../filtering/#match)) filters, `false` to disable
-- `range` - `true` to enable support for [Range](../filtering/#range) filters, `false` to disable
+- `lookup`: enables support for direct lookup using
+  [Match](/documentation/concepts/filtering/#match) filters.
+- `range`: enables support for
+  [Range](/documentation/concepts/filtering/#range) filters.
 
-For example, to set up an optimized integer index only supporting range filters:
+The regular `integer` index assumes both `lookup` and `range` are `true`. In
+contrast, to configure a parameterized index, you would set only one of these
+filters to `true`:
+
+| `lookup` | `range` | Result                      |
+|----------|---------|-----------------------------|
+| `true`   | `true`  | Regular integer index       |
+| `true`   | `false` | Parameterized integer index |
+| `false`  | `true`  | Parameterized integer index |
+| `false`  | `false` | No integer index            |
+
+The parameterized index can enhance performance in collections with millions
+of points. We encourage you to try it out. If it does not enhance performance
+in your use case, you can always restore the regular `integer` index.
+
+Note: If you set `"lookup": true` with a range filter, that may lead to
+significant performance issues.
+
+For example, the following code sets up a parameterized integer index which
+supports only range filters:
 
 ```http
 PUT /collections/{collection_name}/index
--- a/qdrant-landing/content/documentation/concepts/indexing.md
+++ b/qdrant-landing/content/documentation/concepts/indexing.md
@@ -185,6 +185,94 @@ Available tokenizers are:
 
 See [Full Text match](../filtering/#full-text-match) for examples of querying with full-text index.
 
+### Integer index parameters
+
+*Available as of v1.8.0*
+
+The `integer` index also has a parameterized variant which allows fine-tuning
+indexing and search performance.
+
+Generally speaking, we recommend using the standard `integer` index variant
+above without parameters or fine-tuning. Accidentally configuring this
+incorrectly will have a significant effect on performance.
+
+The following flags can be set to build an optimized index:
+- `lookup` - `true` to enable support for direct lookup ([Match](../filtering/#match)) filters, `false` to disable
+- `range` - `true` to enable support for [Range](../filtering/#range) filters, `false` to disable
+
+For example, to set up an optimized integer index only supporting range filters:
+
+```http
+PUT /collections/{collection_name}/index
+{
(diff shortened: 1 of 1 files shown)


qdrant-landing/content/documentation/concepts/indexing.md
```diff
@@ -185,6 +185,94 @@ Available tokenizers are:
...
+
+Generally speaking, we recommend using the standard `integer` index variant
+above without parameters or fine-tuning. Accidentally configuring this
+incorrectly will have a significant effect on performance.
+
+The following flags can be set to build an optimized index:
+- `lookup` - `true` to enable support for direct lookup ([Match](../filtering/#match)) filters, `false` to disable
+- `range` - `true` to enable support for [Range](../filtering/#range) filters, `false` to disable
+
+For example, to set up an optimized integer index only supporting range filters:
```
mjang, line 203: Bigger picture question, why would someone want to do "fine-tuning" of indexing. Does this also "fine-tune" "search performance"? In addition, what's the use case?

If there is a use case, I'd add that just after the "Available as of v1.8.0"

```suggestion
You can configure the `integer` index with parameters. This supports fine-tuning of indexing and search performance.

Note: If you make a mistake, this could negatively affect performance.

You can set the following flags to build an optimized index:
- `lookup` - `true` to enable support for direct lookup ([Match](../filtering/#match)) filters, `false` to disable
- `range` - `true` to enable support for [Range](../filtering/#range) filters, `false` to disable

With the following code, you can set up an optimized integer index to support range filters:
```

mjang (changes requested):
Thanks for the look @timvisee !

My suggested changes focus on readability and persona, so the user thinks "Oh, this is what I can do"

in the future, I want to implement Hugo tools which add a "WARNING" logo. Seems like it'd help with the warning that you've included.

qdrant-landing/content/documentation/concepts/indexing.md
generall: > Bigger picture question, why would someone want to do "fine-tuning" of indexing. 

It might allow less RAM usage.

timvisee: > Note: If you make a mistake, this could negatively affect performance.

Don't you think this is quite unactionable by itself? Wouldn't it be better to call to use regular `integer` instead, so the user knows what to do when reading this note/warning.

Maybe something like this?

```diff
-Note: If you make a mistake, this could negatively affect performance.
+Note: If you make a mistake, this could negatively affect performance. You can always use the regular `integer` index instead.
```

Also, I think this is an important distinction:

```diff
-With the following code, you can set up an optimized integer index to support range filters:
+With the following code, you can set up an optimized integer index to only support range filters:
```

A rough use case description at the top may be something like:

> A huge collection with millions of points, having an `integer` index just using Match or Range filters, is a good candidate for further fine tuning. Configuring the index to support either Match or Range will improve indexing time, improve search performance and reduce memory usage.

> in the future, I want to implement Hugo tools which add a "WARNING" logo. Seems like it'd help with the warning that you've included.

We do have something called a `status` and `alert` box, used [here](https://qdrant.tech/documentation/concepts/collections/#update-vector-parameters) (the blue box) for example. But I find it a bit too _in your face_ to use everywhere.

With all these changes, you'd get something like:

```suggestion
A huge collection with millions of points, having an `integer` index exclusively
using [Match](../filtering/#match) **or** [Range](../filtering/#range) filters,
is a good candidate for further fine tuning. Configuring the index to support
either match or range will improve indexing time, improve search performance and
reduce memory usage.

You can configure the `integer` index with parameters. This supports fine-tuning
of indexing and search performance.

Note: If you make a mistake, this could negatively affect performance. You can
always use the regular `integer` index instead.

You can set the following flags to build an optimized index:
- `lookup` - `true` to enable support for direct lookup
- ([Match](../filtering/#match)) filters, `false` to disable `range` - `true` to
- enable support for [Range](../filtering/#range) filters, `false` to disable

With the following code, you can set up an optimized integer index to only
support range filters:
```

@mjang What do you think? Feel free to change this further. I'm sure you have a good view on what end-users like to see.

mjang: Thanks. @timvisee , what you've done is good. If you want to "stick with it," I'm happy to approve.

As we sometimes say in the US, I could continue to edit "until the cows come home." With that in mind, here's my idea. My first sentence highlights the key feature from v1.8.0. I then use your explanation of "why," followed by the bulleted list of what you need to change to build an optimized index.

I now see that you've bolded the "**or**" in Match or Range. What happens if someone enables both?

My assumption: the default values for `lookup` and `range` are false. If true, there's no need to write "`false` to disable"

```suggestion
For larger collections, you can now fine-tune indexing and enhance search performance.

Larger collections have millions of points. They have an `integer` index which
exclusively uses [Match](/documentation/conceps/filtering/#match) **or** 
[Range](/documentation/concepts/filtering/#range) filters.
When you configure the index to support match or range, that can reduce indexing time,
improve search performance, and reduce memory usage.

You can optimize the `integer` index with parameters, which supports fine-tuning
of indexing and search performance. To build an optimized index, set the following flags:

- `lookup` - `true` to enable support for direct lookup ([Match](../filtering/#match)) filters
- `range` - `true` to enable support for [Range](../filtering/#range) filters

Note: If you make a mistake, this could negatively affect performance. If you see
problems, you can always restore the regular `integer` index.

With the following code, you can set up an optimized integer index to enable range
(and disable match) filters:
```

timvisee: This is intended to just enable one of lookup/range, not both. A user can enable both, but then it would be exactly the same as the regular `integer` index, so there's no use case for that.

These flags don't have a default to prevent confusion, so both must always be provided.

I've therefore modified it further to:

```suggestion
For larger collections, you can now fine-tune indexing and enhance search performance.

Large collections easily have millions of points. If you have an `integer` index on which you
either use [Match](/documentation/conceps/filtering/#match) or
[Range](/documentation/concepts/filtering/#range) filters (not both), you can make use of fine-tuning.
When you configure the index to exclusively support match or range filters, that can reduce indexing time,
improve search performance, and reduce memory usage.

You can optimize the `integer` index with parameters, which supports fine-tuning
of indexing and search performance. To build an optimized index, set the following flags:

- `lookup` - `true` to enable support for direct lookup ([Match](../filtering/#match)) filters, `false` to disable support
- `range` - `true` to enable support for [Range](../filtering/#range) filters, `false` to disable support

Note: If you make a mistake, this could negatively affect performance. If you see
problems, you can always restore the regular `integer` index.

With the following code, you can set up an optimized integer index to enable range
(and disable match) filters:
```

timvisee: We've both agreed on a new version, which I've committed in 6c95d676a3b1880ffae224703e84a419ab6e7930

mjang (approved):
Thanks @timvisee ! (For others, we discussed options in detail)

PR #546, opened 2024-01-23
Feedback buttons

trean:
Slightly fixed feedback buttons introduced by @mjang

### DIFF
--- a/qdrant-landing/config.toml
+++ b/qdrant-landing/config.toml
@@ -83,7 +83,7 @@ keywords = "search engine, vector database, neural network, matching, filter, Sa
 
 [params.feedback]
   enable = true
-  max_value = 1
+  max_value = 100
   title = "Feedback"
   question = "Was this page helpful?"
   positive = "Yes"


trean:
@mjang I can make this PR to your branch if you will rebase your branch to the master. I needed some fresh code from the master so that not fix outdated code.
Also:
 - I don't know the `max_value` you want to send to the analytics. I set it to `1` in `config.toml`.
- Google Analytics does not work in dev mode on purpose, only in prod. But you can temporarily comment conditions around GA scripts in the head to make it work.

qdrant-landing/config.toml
```diff
@@ -81,6 +81,16 @@ keywords = "search engine, vector database, neural network, matching, filter, Sa
   name = "Andrey Vasnetsov"
   email = "info@qdrant.tech"
 
+[params.feedback]
```
mjang, line 84: Based on my understanding of https://www.docsy.dev/docs/adding-content/feedback/#setup-1, I _think_ this should be:

```suggestion
[params.ui.feedback]
```

When I search Hugo discourse forums, https://discourse.gohugo.io/, I see references only to `params.ui.feedback`. If I understand the Docsy docs correctly, we can then see the results in Google Analytics.

mjang (commented):
OMG, thanks so much for this! I _assume_ that the colors that you selected are consistent with how we've set up our website. 

I have just one suggestion. Based on the docs I've read, it should transmit reports to Google Analytics.

mjang:
> @mjang I can make this PR to your branch if you will rebase your branch to the master. I needed some fresh code from the master so that not fix outdated code. Also:
> 
> * I don't know the `max_value` you want to send to the analytics. I set it to `1` in `config.toml`.
> * Google Analytics does not work in dev mode on purpose, only in prod. But you can temporarily comment conditions around GA scripts in the head to make it work.

Thanks! I've made one comment. Based on the [docs I've read](https://www.docsy.dev/docs/adding-content/feedback/#setup-1), the default `max_value` is 100. I can adjust it later if needed. 

I've also rebased my `mjang-setup-feedback-form` to master. You're welcome to do whatever is easier for you: PR to the branch or merge directly.

qdrant-landing/config.toml
trean: This is config for Hugo and Google Analytics can not know anything about it because GA works with a static site build. On the site, this code will look just like a number:

image


I didn't find any native parameters connected to feedback in Hugo docs, so I don't see a reason to introduce an additional level of nesting in this place, so I propose to leave it like this.

generall (approved):

qdrant-landing/config.toml
mjang: Thanks so much! I'll keep you posted on how it goes. My understanding is that it can take up to 24 hours for the Yes/No "clicks" to appear in Google Analytics numbers. If I don't see anything tomorrow morning (my time -- US Pacific), I might `s/params.feedback/params.ui.feedback/` then.

PR #550, opened 2024-01-24
Removed sendFeedback function

trean:
@generall @mjang it's appeared that we should not trigger GA events programmatically, and will do it from GTM interface. You can ask @Alyonkka for additional information.

In this PR: removed `sendFeedback` function and it's call because we need to do it from GTM UI, not programmatically.


mjang (approved):
@trean , thank you for explaining. @Alyonkka has already DM'd me, and we'll figure out how to set it up from GA.

I understand what you've done to remove programmatic feedback from the Hugo code, and therefore approve this PR.

PR #558, opened 2024-01-26
Correction to features we don't support

mjang:


bashofmann (approved):

mjang:
Merging, based on @bashofmann 's approval. If there are other mistakes, we'll correct them as we identify them.

PR #559, opened 2024-01-26
Set up an index page for concepts

mjang:
I'd like us to present an overview of all concepts shown in our docs. With this PR, I've set this up as a table, with links to each concept.

Build of new page: https://deploy-preview-559--condescending-goldwasser-91acf0.netlify.app/documentation/concepts/


joein:
It might be worth comparing <h2> with <h3>, however already looks good enough both on a desktop and a phone

joein (approved):
It might be worth comparing `<h2>` with `<h3>`, however already looks good enough both on a desktop and a phone

mjang:
Thanks for your help, @joein (We brainstormed different heading levels, for comparative results.)

PR #560, opened 2024-01-26
Moving what-is-vdbms to articles

sabrinaaquino:
Moving the new post to articles - I've created an alias for the old link.


mjang:
@sabrinaaquino I know I said that I was going to review this further. But since this is now published, I'm going to comment on the conflict. Before merging, you'll need to delete the `qdrant-landing/content/blog/what-is-a-vector-database.md` file. I think that should solve the conflict shown in the build:
Screenshot from 2024-01-29 07-24-36

mjang (commented):
I see that the build seems to work as before. I've got some "big picture" open issues that I need to prioritize (IMO, we need a better process for `weight`, and should start including "alt text" with images.) But they're not blockers for this PR.

But as I noted in my other comment, I think you need to remove the original blog Markdown file before merging.

generall:
One annoying thing we have in articles (which we will get rid off once we have a new design), is that we have icons for articles. They should match the design: 
image

I suggest something like this https://qdrant.tech/articles_data/chatgpt-plugin/chatgpt-plugin-icon.svg (white)

PR #561, opened 2024-01-26
Update collections page HTTP code samples

mjang:
WIP. I've set up the HTTP code with `curl`, so readers can 'copy/paste' the REST calls in their CLIs. 

This process has also helped me ID a couple of cases with invalid JSON (a couple of extra commas, so far)


qdrant-landing/content/documentation/concepts/collections.md
```diff
@@ -101,6 +107,10 @@ client.createCollectionAsync("test_collection",
         VectorParams.newBuilder().setDistance(Distance.Dot).setSize(4).build()).get();
 ```
 
+If successful, these commands write data to the `qdrant_storage` subdirectory. You
+can verify changes in the `config.json` file in the directory associated with
```
generall, line 111: Do you think we need to expose internal structure in docs?

mjang: We don't have to. IMO the responses to these code samples should be "good enough". (My work heritage is heavy on open source, so my instinct is to 'expose everything'.) I'll delete these lines.

mjang:
> Maybe we can add another code block with `curl`, which does show all required curl parameters.

This is exactly what @generall suggested that I do. I'll update accordingly. (I'll set this up in different PRs, a page at a time.)

mjang:
I had some rebase issues. Moved to https://github.com/qdrant/landing_page/pull/565

PR #567, opened 2024-01-31
Fix social preview options

mjang:
Correct error in `social_preview.jpg` file.


mjang:
Superseded by #568 . Thanks @coszio

PR #576, opened 2024-02-02
Update fastembed.md

sabrinaaquino:
image2 was used twice instead of image3


mjang:
> image2 was used twice instead of image3

Suggestion -- change the file names to something more descriptive, maybe 
- generate-embeddings-query.png
- generate-embeddings-from-docs.png

@NirantK , these look like mermaid.js diagrams. Future reference -- if you set up diagrams that need minor changes, our doc tooling should be able to handle them, with minor revisions, ref https://gohugo.io/content-management/diagrams/#mermaid-diagrams

NirantK (approved):

PR #587, opened 2024-02-06
Article on vector embeddings

sabrinaaquino:


mjang (approved):
I previously reviewed the content in a Google Doc, and I see that @sabrinaaquino has addressed my requests here.

PR #597, opened 2024-02-09
WIP: Add documentation page for Hybrid Cloud

bashofmann:


mjang (commented):
Thanks for the preview, @bashofmann . 

Some first comments:
- I'd like to see a ref to k8s docs. Example, when referring to a standard cluster, I wonder if this doc might work? https://kubernetes.io/docs/concepts/cluster-administration/. The current link is to a k8s training page. 
- Open question: with the current cloud config, we tell the user to go to AWS, GCP,or Azure and install our plugin. It might help to do a comparison, so users can choose: "Do I import my own cluster, or do I set it up "the old way" with Qdrant?"

I'll do a detailed review once we see the process for activating and using "Private Regions"

bashofmann:
The URL https://www.cncf.io/training/certification/software-conformance/ is pointing to the list of CNCF verified Kubernetes distributions.

mjang:
I've pushed a couple of commits with my suggested rewrites. I also have a list of questions / unknowns, which I think we should resolve before publishing:

Unknowns:
- Is this available as of some Qdrant version number (next would be v1.8.2)
  - Alternative: specify a date
- Are "Qdrant Cloud" and "Hybrid Cloud" both "products?" (do I capitalize both words?)
- What to enter in the "Operator Configuration" section 
  - UI suggests that docs should supply a "Full list of configuration options"
  - I have a blank "Operator configuration" section 
- "Signing up for Qdrant Cloud"
  - Do we still need this section?

Empty sections:
- Authentication
  - There's a reference to an API key. Is that enough?
  - Are there any other authentication requirements
- Exposing Qdrant clusters to your client applications
  - Best guess, requires open ports in the clients (80, 443, 6333, 6334)

davidmyriel (approved):
would like to merge with the main doc.....will add more info later over the weekend

bashofmann:
@davidmyriel merged

PR #618, opened 2024-02-16
Rework the snapshots page for clarity

mac-chaffee:
After working with a Qdrant Cloud customer to transfer data to a new cluster, they found it difficult to follow our current docs due to the mixture of instructions that only apply to Qdrant Cloud and instructions that only apply to self-hosted Qdrant. They also had some trouble with `curl` syntax and the difference between the "restore" vs. "upload and restore" endpoints.

So I've reworked the page with the following changes:

* Mention backups as an alternative to snapshots for Qdrant Cloud users
* Split the restoration section into 3 sections based on the 3 methods, calling out specifically the use-cases for each method
* Explain how the snapshot upload API can accept a remote URL or a file URI
* Mention that missing collections are automatically created during restore
* Link to the API docs
* Add a curl command for downloading snapshots that shows how to use the `--output` flag.
* Add api-key headers to all curl commands (the extra header is harmless for self-hosted users, but its absence confuses qdrant cloud users)
* Ensure `{collection_name}` is always shown in brackets to indicate a user needs to fill in that part
* Clarify snapshot priority
* Mention that whole storage snapshots can't be restored in Qdrant Cloud (maybe we should create an API for that at some point?)

Also
* Small improvements to the create-snapshot tutorial

### DIFF
--- a/qdrant-landing/content/documentation/concepts/snapshots.md
+++ b/qdrant-landing/content/documentation/concepts/snapshots.md
@@ -294,7 +294,7 @@ If you wish instead to overwrite an existing collection, use the `--force_snapsh
 ### Snapshot priority
 
 When recovering a snapshot to a non-empty node, there may be conflicts between the snapshot data and the existing data. The "priority" setting controls how Qdrant handles these conflicts. The priority setting is important because different priorities can give very
-different end results. The default priority may not be best for all sitations.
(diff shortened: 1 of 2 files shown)


timvisee (approved):
Thank you for the careful rewriting!

generall (approved):

PR #622, opened 2024-02-20
Fix upload_records deprecation for upload_points

amartinh:
While going through the [Semantic Search for Beginners tutorial](https://qdrant.tech/documentation/tutorials/search-beginners/), I got a deprecation notice at step 5 while trying to use  `upload_records` method. Following the recommendation given by the notice, this PR proposes to use `upload_points` method and `points` related property instead.


timvisee (approved):
I've confirmed this based on this resource: https://python-client.qdrant.tech/qdrant_client.async_qdrant_client.html?highlight=upsert#qdrant_client.async_qdrant_client.AsyncQdrantClient.upload_points

Thanks!

joein (changes requested):
Hey @amartinh 

Thanks for pointing it out

`upload_points` accepts not `models.Record`, but `models.PointStruct`

Would you like to fix it?

amartinh:
Sure! will do later today.

amartinh:
@joein Requested changes are done. Kindly asking for your review again. Thanks.

joein:
`models.Record` actually contains all the same fields as `models.PointStruct`, but also contains a field for a custom shard key. We use `models.Record` as an output model, not input.

By the way, `upload_points` is available as of `qdrant-client v1.7.1`, thus we need to update `pip install qdrant-client` command.
It should either be `pip install "qdrant-client>=1.7.1"` or `pip install -U qdrant-client` with a warning that this tutorial requires qdrant-client at least 1.7.1

amartinh:
> models.Record actually contains all the same fields as models.PointStruct, but also contains a field for a custom shard key. We use models.Record as an output model, not input.
Thanks for the info!
> By the way, upload_points is available as of qdrant-client v1.7.1, thus we need to update pip install qdrant-client command.
It should either be pip install "qdrant-client>=1.7.1" or pip install -U qdrant-client with a warning that this tutorial requires qdrant-client at least 1.7.1
Happy to add that change. Do you want a separate PR or this one is fine?

joein:
This one is fine, since the changes are related

joein:
Yes, thank you for the contribution!

joein (approved):

PR #626, opened 2024-02-21
Update llama-index-multitenancy.md to work with llama-index v0.10

uncleDecart:
Starting v0.10 [llama-index](https://github.com/run-llama/llama_index/releases/tag/v0.10.1) is split between llama_index and llama_index.core libraries. This commit updates documentation to work with llama-index v0.10 and higher. Also this commit adds note on how to run it with local llamaIndex model (like llama.cpp)

### DIFF
--- a/qdrant-landing/content/documentation/tutorials/llama-index-multitenancy.md
+++ b/qdrant-landing/content/documentation/tutorials/llama-index-multitenancy.md
@@ -43,7 +43,8 @@ of `QdrantClient`.
 
 ```python
 from qdrant_client import QdrantClient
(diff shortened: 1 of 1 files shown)


kacperlukawski:
@uncleDecart I guess the dependencies should also be updated to include `llama-index-vector-stores-qdrant`, shouldn't they?

uncleDecart:
> @uncleDecart I guess the dependencies should also be updated to include llama-index-vector-stores-qdrant, shouldn't they?
You're correct @kacperlukawski, updated PR. Thanks!

PR #631, opened 2024-02-22
Add documentation for `order_by`

coszio:
Adds the documentation for the new `order_by` feature for the `scroll` endpoint

Related issue: qdrant/qdrant#3521

Drafting only because it should be merged after v1.8 release

### DIFF
--- a/qdrant-landing/content/documentation/concepts/points.md
+++ b/qdrant-landing/content/documentation/concepts/points.md
@@ -1718,6 +1718,8 @@ order_by: Some(OrderBy {
 
 An ID offset does not work when the order is based on a non-unique value. Since pagination is not explicitly enabled when using `order_by`, you will not see the `next_page_offset` field in the response. However, you can use a combination of `"order_by": { "start_from": ... }` and a `{ "must_not": [{ "has_id": [...] }] }` filter to achieve the same effect.
 
+When sorting is based on a non-unique value, it is not possible to rely on an ID offset. Thus, next_page_offset is not returned within the response. However, you can still do pagination by combining `"order_by": { "start_from": ... }` with a `{ "must_not": [{ "has_id": [...] }] }` filter.
+
 ## Counting points
 
 _Available as of v0.8.4_
--- a/qdrant-landing/content/documentation/concepts/points.md
+++ b/qdrant-landing/content/documentation/concepts/points.md
@@ -369,7 +369,6 @@ client.upsert("{collection_name}", {
 });
 ```
 
-
 or record-oriented equivalent:
 
 ```http
@@ -897,7 +896,7 @@ client.upsert(
             vector={
                 "text": models.SparseVector(
                     indices=[1, 2, 3, 4, 5],
-                    values= [0.1, 0.2, 0.3, 0.4, 0.5],
+                    values=[0.1, 0.2, 0.3, 0.4, 0.5],
                 )
             },
         ),
@@ -913,7 +912,7 @@ client.upsert("{collection_name}", {
       vector: {
         text: {
           indices: [6, 7],
-          values: [1.0, 2.0]
+          values: [1.0, 2.0],
         },
       },
     },
@@ -921,8 +920,8 @@ client.upsert("{collection_name}", {
       id: 2,
       vector: {
         text: {
-          indices=[1, 2, 3, 4, 5],
-          values= [0.1, 0.2, 0.3, 0.4, 0.5],
+          indices: [1, 2, 3, 4, 5],
+          values: [0.1, 0.2, 0.3, 0.4, 0.5],
         },
       },
     },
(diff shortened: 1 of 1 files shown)


mjang:
I've arbitrarily added a `do not merge` label, to avoid accidents (until we release v1.8). You're welcome to substitute some other safeguard.

qdrant-landing/content/documentation/concepts/points.md
```diff
@@ -1632,12 +1631,97 @@ All resulting points are sorted by ID. To query the next page it is necessary to
...
 
-<!-- 
-Python client:
+### Order points by a payload key
+
+_Available as of v1.8_
+
+When using the [`scroll`](#scroll-points) API, it is possible to sort the results by the contents of a payload field. This is useful when you want to retrieve points in a specific order, for example, by a timestamp, for chronological order.
+
+<aside role="status">Without an appropriate index, custom ordering would be a very expensive operation. For this reason, in Qdrant, it is a requirement to have an index capable of <a href=/documentation/concepts/indexing/#payload-index target="_blank">Range filtering conditions</a> on the field you want to `order_by`. Once the index is set up, these ordering operations can be really cheap.</aside>
```
mjang, line 1640: By "expensive," I'm guessing you mean as in load on the system (not actual cost). I therefore suggest:

```suggestion
<aside role="status">Without an appropriate index, custom ordering can create costly loads. Qdrant therefore requires an index which supports <a href=/documentation/concepts/indexing/#payload-index target="_blank">Range filtering conditions</a> on one ore more specific `order_by` fields.</aside>
```

qdrant-landing/content/documentation/concepts/points.md
```diff
@@ -1632,12 +1631,97 @@ All resulting points are sorted by ID. To query the next page it is necessary to
...
+        order_by: Some(OrderBy {
+            key: "timestamp".to_string(),  // <-- this!
+            ..Default::default(),
+        }),
+        ..Default::default()
+    })
+    .await?;
+```
+
+The `order_by` `key` parameter specifies the payload key to order the results by, but there are also other fields that can be set to control the ordering, like `direction` and `start_from`:
```
mjang, line 1681:
```suggestion
The `order_by` `key` parameter is one way to order the results. You can substitute other parameters for `key` such as `direction` and `start_from`:
```

qdrant-landing/content/documentation/concepts/points.md
```diff
@@ -1632,12 +1631,97 @@ All resulting points are sorted by ID. To query the next page it is necessary to
...
+
+```rust
+order_by: Some(OrderBy {
+    key: "timestamp".to_string(),
+    direction: Some(Direction::Desc) // default is Direction::Asc
+    start_from: Some(123),
+})
+```
+
+Be mindful that, for array-like payloads, each point can appear as many times as the number of elements in the array. For example, if you have a point payload with a `timestamp` key, and the value for the key is an array of 3 elements, the same point will appear 3 times in the results, one for each timestamp.
```
mjang, line 1715:
```suggestion
Note: for payloads organized in arrays, you may get what seems like duplicate results. For example, if you have a point payload with a `timestamp` key, and the value is an array of 3 elements, the same point appears 3 times in the results.
```

qdrant-landing/content/documentation/concepts/points.md
```diff
@@ -1632,12 +1631,97 @@ All resulting points are sorted by ID. To query the next page it is necessary to
...
+
+<aside role="alert">Pagination is disabled when using the `order_by` parameter. </aside>
```
mjang, line 1717:
```suggestion
<aside role="alert">When you use the `order_by` parameter, pagination is disabled. </aside>
```

qdrant-landing/content/documentation/concepts/points.md
```diff
@@ -1632,12 +1631,97 @@ All resulting points are sorted by ID. To query the next page it is necessary to
...
+
+As you may expect, pagination does not work exactly the same way than when the order is an ID, which is unique. Right now, pagination is not explicitly enabled when ordering by a payload key, so the `next_page_offset` field will not be present in the response. However, there is a trick you can do to achieve pagination:
```
mjang, line 1719:
```suggestion
Pagination works differently when  `order_by` is an ID. Since pagination is not explicitly enabled, you do not see the `next_page_offset` field in the response. However, you can use the following steps to set up pagination:
```

qdrant-landing/content/documentation/concepts/points.md
```diff
@@ -1632,12 +1631,97 @@ All resulting points are sorted by ID. To query the next page it is necessary to
...
+
+1. For each page, store the last value of the `order_by` field.
+2. Accumulate all ids which have the same "last value" in their `order_by` `key`.
+3. Make a `has_id` filter with all the ids you have accumulated.
+4. Use this filter and the "last value" as the `start_from` in the next request.
```
mjang, line 1724: Tip: As long as you've set up spaces correctly, you don't have to specify consecutive numbers. This makes life easier if  you have to insert / delete a step.

```suggestion
1. For each page, store the last value of the `order_by` field.
1. Accumulate all IDs with the same "last value" in their `order_by` `key`.
1. Create a `has_id` filter with accumulated IDs.
1. Use this filter and the "last value" as the `start_from` in the next request.
```

mjang (changes requested):
Thanks for the look!

qdrant-landing/content/documentation/concepts/points.md
coszio: Today I learned about this, thanks!

qdrant-landing/content/documentation/concepts/points.md
coszio: Rephrased it a little different

qdrant-landing/content/documentation/concepts/points.md
```diff
@@ -1632,12 +1631,97 @@ All resulting points are sorted by ID. To query the next page it is necessary to
 For convenience, this ID is also returned in the field `next_page_offset`.
 If the value of the `next_page_offset` field is `null` - the last page is reached.
 
-<!-- 
-Python client:
+### Order points by a payload key
+
+_Available as of v1.8_
+
+When using the [`scroll`](#scroll-points) API, it is possible to sort the results by the contents of a payload field. This is useful when you want to retrieve points in a specific order, for example, by a timestamp, for chronological order.
```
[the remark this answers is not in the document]
coszio: Rephrased it a little different

qdrant-landing/content/documentation/concepts/points.md
coszio: "costly" is not the same as "expensive"?

mjang: > "costly" is not the same as "expensive"?

I asked ChatGPT the question, and I like the answer:

Screenshot from 2024-02-22 12-25-53

qdrant-landing/content/documentation/concepts/points.md
mjang: I don't feel strongly about this, will approve whether you commit this suggestion or not.

mjang (approved):
Thanks for the updates. Ready to go when 1.8 is ready.

qdrant-landing/content/documentation/concepts/points.md
coszio: Yeah, I think the suggestion _suggests_ that the `key` param is optional, but it is not. What I wanted to convey is that you can specify these other extra parameters along with the `key`

mjang: Ah thank you for explaining. Then maybe:

```suggestion
Use the `order_by` `key` parameter to specify the payload key. You can then add other fields to control the ordering, like `direction` and `start_from`:
```

joein:
@coszio I think this one is hard to understand:

1) 
```
However, you can use the following steps to set up pagination:

For each page, store the last value of the order_by field.
Accumulate all IDs with the same “last value” in their order_by key.
Create a has_id filter with accumulated IDs.
Use this filter and the “last value” as the start_from in the next request.
```

It is not straightforward what a user needs to do here.

`For each page, store the last value of the order_by field.`

There is no explicit starting point, it begins with `for each page`. 
We need to tell users to make at least the first scroll request.

`Create a has_id filter with accumulated IDs.` - we should say that this filter should be used for the exclusion of already seen points

We might also need to say "set this filter as a `scroll_filter` param", but maybe it is not necessary.

2)

`Without an appropriate index, custom ordering would create costly queries. Qdrant therefore requires a payload index which supports Range filtering conditions on the field used for order_by`

It's probably just a matter of taste, but I would rather swap and rewrite these sentences a little bit. 
The main point here is that Qdrant requires payload of certain types - the types which are compatible with `Range` payload index. (I assume that `order_by` does not work with keywords at all)
Without such indices these queries will be costly.

I didn't do my best to write the final version of the sentences, cause I am not that good at writing.
I am just suggesting you to consider these ideas, feel free to reject them if you don't like them.

coszio:
@joein In the end I decided that it is too hacky to mention the exact logic users would have to implement for pagination to work if pagination is not enabled in the server. 
So my solution is to just mention the tools needed to achieve it if they want it, wdyt?

coszio:
@Anush008 Could you please add the Java and C# examples once you're done with the client updates? 🙏

Anush008:
@coszio, will do 👍

qdrant-landing/content/documentation/concepts/points.md
```diff
@@ -1632,24 +1631,144 @@ All resulting points are sorted by ID. To query the next page it is necessary to
 For convenience, this ID is also returned in the field `next_page_offset`.
 If the value of the `next_page_offset` field is `null` - the last page is reached.
 
-<!-- 
-Python client:
+### Order points by a payload key
+
+_Available as of v1.8_
```
xzfc, line 1636: We did specify patch version for older releases, e.g.
https://github.com/qdrant/landing_page/blob/069857d0c307abfaba8e4a3ae4a300c1c2004e11/qdrant-landing/content/documentation/concepts/points.md#L818


```suggestion
_Available as of v1.8.0_
```

PR #655, opened 2024-02-29
Set up separate Web UI doc

mjang:
I think this took me < 1h to set up, and addresses IMO one of the biggest misses in our docs.


davidmyriel (approved):
lgtm

PR #667, opened 2024-03-04
Add social preview images for various documentation pages

NirantK:
These are generated from [the script](https://github.com/qdrant/landing_page/blob/master/automation/generate-all-docs-preview.sh)

The script advocates generating social previews for all documentation pages. Since this'd make our social previews also consistent, I hope that this is useful?


generall:
something is wrong with this change, too many files changed

generall:
this is how it is supposed to look like https://github.com/qdrant/landing_page/pull/672/files

NirantK:
I've used the very same script. From the PR which you shared, the placement is wrong? We'd prefer to have these nested in directories for each post?

generall:
> We'd prefer to have these nested in directories for each post?

this is not just preferred. It is how it works and majority of previews are already there.

> I've used the very same script.

I am not sure what could go wrong, but apparently we need to move it into CI

mjang:
Question: why do we need preview images for our documentation?

mjang:
As suggested by @NirantK , I'm closing this PR, and will defer to @generall 's guidance.

PR #675, opened 2024-03-04
Broken links fixes

trean:


mjang:
Thanks, @trean . (Begs the question, why did this fail in lychee?)

PR #676, opened 2024-03-04
Fix broken social image links

mjang:


generall (changes requested):
Why it uses images from the old article?

mjang:
Closing, as lychee no longer shows broken links from this file. (I see based on https://github.com/qdrant/landing_page/pull/677/files)

PR #679, opened 2024-03-05
Removed pages with taxonomy

trean:
Hugo automatically creates pages for tags and categories, but we currently don't need them.
In this PR, I have disabled the rendering of those pages.
However, we can still utilize taxonomy for internal purposes such as displaying related posts on our blog.


generall (changes requested):
Let's specify aliases in blog for removed pages

trean:
@generall added aliases

**Note for the content team**
There is no need to add redirects for future new tags and categories!

PR #686, opened 2024-03-06
Set up release post template

mjang:
Starting point for future release posts


davidmyriel (approved):

davidmyriel:
we should do one for blogs and case studies then too

PR #701, opened 2024-03-08
Add clarifying process to code search demo

mjang:


kacperlukawski (approved):
@mjang Thanks for putting it together. I moved the process description to the section about our code search demo. It is worth noting that the demo and tutorial are separate, and the latter is a lot simpler and does not follow the same process. I wanted to make sure everything was clear and moved it to a separate section.

Also, I checked the process implemented in the demo once again and added more details.

Please let me know if that's okay for you. I'm already approving the PR.

PR #705, opened 2024-03-11
Add accordion dropdown to Hugo

mjang:
Purpose: we have a lot of docs with info that "not everyone needs"

I've explained this in the build, in https://deploy-preview-705--condescending-goldwasser-91acf0.netlify.app/documentation/test-partials/

@trean , someone suggested that you might have already implemented a similar feature.

I've tested on my local system -- it works in blogs (and therefore should also work in articles)

- This will also help marketing, as their recommendation is to use this kind of feature in blogs/articles


trean (commented):
Thanks for adding a new feature @mjang 

To me, looks good, the only thing I would change is the style of the chevron to make it more vertically centered and maybe a little bit smaller.

But I have a question, why do you prefer this shortcode over just using <detail> tag which works natively in browsers without js? Like this:
```
<details>
  <summary>"More info on HNSW"</summary>
  Once we have a canonical Qdrant definition for HNSW, we can add it here.
</details>
```

mjang:
@trean, I've updated this PR per your suggestion.

At a later time, if we decide we need a different approach, I could revive the Hugo option. I'd like to keep the `test-partials.md` file to demo other Hugo features. Since it's hidden from the ToC, I think it's OK.

trean:
> @trean, I've updated this PR per your suggestion.
> 
> At a later time, if we decide we need a different approach, I could revive the Hugo option. I'd like to keep the `test-partials.md` file to demo other Hugo features. Since it's hidden from the ToC, I think it's OK.

Ok, but let's make `test-partials.md` a draft, so it would not be rendered in prod.

mjang:
> > @trean, I've updated this PR per your suggestion.
> > At a later time, if we decide we need a different approach, I could revive the Hugo option. I'd like to keep the `test-partials.md` file to demo other Hugo features. Since it's hidden from the ToC, I think it's OK.
> 
> Ok, but let's make `test-partials.md` a draft, so it would not be rendered in prod.

Makes sense. Since we're not going to use expand.html, I've removed that as well.

trean (approved):

trean:
@mjang I missed that the build was broken. It happened because we can not use a non-existent shortcut inside markdown (because the file was removed). I reverted the merge of this PR for now.

PR #711, opened 2024-03-13
docs(faq): add downgrade question

SamirPS:
This commit permit to tell our users that we don't support downgrading of clusters

### DIFF
--- a/qdrant-landing/content/documentation/faq/qdrant-fundamentals.md
+++ b/qdrant-landing/content/documentation/faq/qdrant-fundamentals.md
@@ -66,9 +66,10 @@ But in some cases, we might be able to help you with that through manual interve
 
 ## Versioning
 
-### Do you support downgrade?
+### Do you support downgrades?
 
-We does not support cluster downgrade. This applies to both cloud and core products. Deploying a newer version of Qdrant will automatically migrate your data to the newer storage format, which is not reversible.
+We does not support downgrading a cluster on any of our products. If you deploy a newer version of Qdrant, your
+data is automatically migrated to the newer storage format. This migration is not reversible.
 
 ### How do I avoid issues when updating to the latest version?
 
--- a/qdrant-landing/content/documentation/faq/qdrant-fundamentals.md
+++ b/qdrant-landing/content/documentation/faq/qdrant-fundamentals.md
@@ -68,7 +68,7 @@ But in some cases, we might be able to help you with that through manual interve
 
 ### Do you support downgrade?
 
-We does not support cluster downgrade. This applies to both cloud and core product.
+We does not support cluster downgrade. This applies to both cloud and core products. Deploying a newer version of Qdrant will automatically migrate your data to the newer storage format, which is not reversible.
 
 ### How do I avoid issues when updating to the latest version?
 
--- a/qdrant-landing/content/documentation/faq/qdrant-fundamentals.md
+++ b/qdrant-landing/content/documentation/faq/qdrant-fundamentals.md
@@ -66,6 +66,10 @@ But in some cases, we might be able to help you with that through manual interve
 
 ## Versioning
 
+### Do you support downgrade?
+
+We does not support cluster downgrade. This applies to both cloud and core product.
+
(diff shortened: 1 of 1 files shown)


qdrant-landing/content/documentation/faq/qdrant-fundamentals.md
```diff
@@ -66,6 +66,10 @@ But in some cases, we might be able to help you with that through manual interve
 
 ## Versioning
 
+### Do you support downgrade?
+
+We does not support cluster downgrade. This applies to both cloud and core product.
```
mac-chaffee, line 71: Maybe it would be nice to include the reasoning?

```suggestion
We does not support cluster downgrade. This applies to both cloud and core products. Deploying a newer version of Qdrant will automatically migrate your data to the newer storage format, which is not reversible.
```

mjang: Question: what happens if someone has their own backup and installs a new instance of an old version of Qdrant? Example, with a config file such as:
```
services:
  qdrant:
    image: qdrant/qdrant: 1.7
```

SamirPS: The backup was made on 1.7 or 1.8?

qdrant-landing/content/documentation/faq/qdrant-fundamentals.md
```diff
@@ -66,6 +66,10 @@ But in some cases, we might be able to help you with that through manual interve
 
 ## Versioning
 
+### Do you support downgrade?
+
+We does not support cluster downgrade. This applies to both cloud and core products. Deploying a newer version of Qdrant will automatically migrate your data to the newer storage format, which is not reversible.
```
mjang, line 71:
```suggestion
### Do you support downgrades?

We does not support downgrading a cluster on any of our products. If you deploy a newer version of Qdrant, your
data is automatically migrated to the newer storage format. This migration is not reversible.
``````

mac-chaffee (approved):

qdrant-landing/content/documentation/faq/qdrant-fundamentals.md
mjang: In this case, presumably 1.7

SamirPS: I would suppose that it's would work. But i don't know if we should promote this , @mac-chaffee any thoughts about this?

mac-chaffee: Yeah if you take a physical backup from v1.7 and then restore it to a qdrant v1.7 cluster, then yes it would work...

If you take a physical backup from v1.7 and then restore it to a qdrant v1.7 cluster, this would also work, since Qdrant would upgrade the data format at start time.

If you take a physical backup from v1.8 and then restore it to a qdrant v1.7 cluster, this would not work, since Qdrant can't downgrade its own storage format.

mjang: > But i don't know if we should promote this

Are we still supporting v1.7?

mjang: >  But i don't know if we should promote this 

Thinking this through, maybe not. Users who understand downgrades, especially of databases, should already understand the scenarios that @mac-chaffee described. Approving.

mjang (approved):

generall:
maybe worth mentioning that we don't support downgrade between minior versions? Between patch versions it should be fine (on the other hand we don't test it explicitly, so I am fine with keeping as-is)

SamirPS:
> maybe worth mentioning that we don't support downgrade between minior versions? Between patch versions it should be fine (on the other hand we don't test it explicitly, so I am fine with keeping as-is)

I can add it if you want but we would need to test it before :)

mac-chaffee:
I vote merge as-is. The main reason for this PR was to tell cloud users that they can't downgrade, and in the cloud you can't even downgrade to a minor version (the UI blocks it).

PR #723, opened 2024-03-14
Update tutorial to describe migration

mjang:
Hi @joekrejci . I agree, this tutorial should serve your purpose. The [Restore from snapshot](https://qdrant.tech/documentation/tutorials/create-snapshot/#restore-from-snapshot) section already suggests that you can use the steps to restore info to a new cluster.

I've added a paragraph which should be consumed by our search interface, so we won't have trouble finding info for your use case in the future.


joekrejci:
Looks good to me!

joekrejci (approved):
Looks good to me, approved!

PR #726, opened 2024-03-15
Fix broken links from issue #703

mjang:
From https://github.com/qdrant/landing_page/issues/703

I didn't fix all the errors. Specifically, IMO, lychee should _not_ have pointed to code examples are errors in https://qdrant.tech/articles/chatgpt-plugin/. Not sure how to address that, but at least with a few minutes of work, I can reduce the number of errors that we have.


mjang:
I think I figured out the other broken links. Will try again in a different PR.

PR #727, opened 2024-03-15
Fix seo schema

trean:


trean:
@generall could you review please?

generall (approved):

PR #750, opened 2024-03-22
Updated styles for docker commands to make copying more user-friendly

trean:
In this PR:
- line numbers excluded from text select
- removed position absolute from code tag to make selecting of text more comfortable


trean:
Force push after rebase

generall (approved):

PR #752, opened 2024-03-23
Add grid layout

nshershnev:


generall:
image

gaps between cards are way to big. Previous version was more accurate

generall:
image

Cards from this section were supposed to be hidden behind the border of screen, not the grid layout. If it is too complicated to implement with screen, I am fine with keeping previous version.

qdrant-landing/themes/qdrant-2024/layouts/partials/core-features.html
```diff
@@ -1,27 +1,33 @@
...
+          <h4 class="core-features__card-title">{{ .Params.cloudFeature.title }}</h4>
+          <p class="core-features__card-content">{{ .Params.cloudFeature.content }}</p>
+          <a class="link link_dark link_md" href="{{ .Params.cloudFeature.link.url }}">
+            {{ .Params.cloudFeature.link.text }}
+          </a>
+        </div>
+        <div class="core-features__cloud-native-image"></div>
+      </div>
+    </div>
+    <div class="core-features__cards row-4 row-xl-12">
```
trean, line 20: It's better to put cards inside `.col`.
`.row-xl-{n}` `.col-xl-{n}` are for bigger devices.
Something like this:
``` html
<div class="core-features__cards row-12 row-xl-4">
  <div class="col-12 col-xl-4">
    <div class="card core-features__card">
      <!-- more code -->
    </div>
  </div>
</div>
<!-- And so on -->
 ```    

Also, if we switch to the bootstrap grid layout we need to adjust some of the existing styles. For example, no need to fixate the width of `.core-features__card` etc.

File: qdrant-landing/themes/qdrant-2024/layouts/partials/core-features.html [hunk 2]
```diff
@@ -1,27 +1,33 @@
...
-  <p class="core-features__subtitle">{{ .Params.description }}</p>
-  <a class="link link_xl" href="{{ .Params.link.url }}">{{ .Params.link.text }}</a>
-  <div class="card core-features__cloud-native">
-    <div>
-      <h4 class="core-features__card-title">{{ .Params.cloudFeature.title }}</h4>
-      <p class="core-features__card-content">{{ .Params.cloudFeature.content }}</p>
-      <a class="link link_dark link_md" href="{{ .Params.cloudFeature.link.url }}">
-        {{ .Params.cloudFeature.link.text }}
-      </a>
+  <div class="container">
```
trean, line 2: Is the default bootstrap `.container` width match our design? If it's not it can be [adjusted](https://getbootstrap.com/docs/5.1/layout/containers/#sass) in `qdrant-landing/themes/qdrant-2024/assets/css/_theme-variables.scss` like this:
``` scss
$container-max-widths: (
  sm: 540px,
  md: 720px,
  lg: 960px,
  xl: 1140px,
  xxl: 1320px
);
```

trean (changes requested):
I left couple of examples of how we would prefer to change the grid usage.

nastyapash (approved):

generall:
On the big screen cards doesn't match layout of all others: 
image


I propose to rollback those cahnges

generall:
Distances between those parts changed. Previous version looked nicer

Screenshot from 2024-03-25 18-20-33
Screenshot from 2024-03-25 18-20-02

trean (approved):
Looks good to me 👍

PR #762, opened 2024-03-27
add enterprise solutions page

nastyapash:


trean (changes requested):
Seems that this file got here by accedent: `qdrant-landing/static/documentation/frameworks/Langchain4J-social-preview.png`

PR #765, opened 2024-03-28
HC AWS / Cohere / Airbyte tutorial

kacperlukawski:
This tutorial showcases implementing a customer support automation system with Airbyte, Cohere (Embed, Command-R, RAG), running on AWS.


davidmyriel (approved):
Might need to do another pass before publication - but looks good so far.  Thanks for following the "shorter" format.

PR #769, opened 2024-03-30
add hybrid cloud page

nastyapash:


qdrant-landing/themes/qdrant-2024/assets/css/partials/_hybrid-cloud-cases.scss
```diff
@@ -0,0 +1,64 @@
...
+      color: $neutral-30;
+    }
+  }
+
+  @include media-breakpoint-up(xl) {
+    padding-top: $spacer * 2.5;
+    padding-bottom: $spacer * 6.5;
+
+    &__wrapper {
+      max-width: rem(1134);
```
trean, line 40: Is this needed? `.container` already gets work done, isn't it?

trean (changes requested):
Looks good. 
I have the only question: why does each section contain a wrapper with a custom class restricting its width? Seems that the `.container` class on the same element already does it.
Could you please clarify this place because I don't see the difference if I remove wrapper classes? Or fix it if it was unintentional?

qdrant-landing/themes/qdrant-2024/assets/css/partials/_hybrid-cloud-cases.scss
nastyapash: Yes, this is not necessary, I will delete it in all my PRs

trean: Thank you so much!

trean (approved):

PR #773, opened 2024-04-01
add private cloud page

nastyapash:


trean (changes requested):
1. The same question as in #769:
> Why does each section contain a wrapper with a custom class restricting its width? Seems that the .container class on the same element already does it.
Could you please clarify this place because I don't see the difference if I remove wrapper classes? Or fix it if it was unintentional?

2. Customers' logos are missing. Just use the same as on the main page.

nastyapash:
@trean I’ve worked with Customers' logos in my other PR (#767).  I’m waiting for it to be approved and merged so there’s no conflict

trean (approved):

PR #783, opened 2024-04-04
Vultr / DSPy / Ollama tutorial on information extraction

kacperlukawski:
This PR adds the tutorial on using DSPy and Ollama with Vultr to create an information extraction tool from the text. The process does not cover optimization yet, but I'll try to add it if there is enough time.


kacperlukawski:
@davidmyriel Visual are ready here as well.

davidmyriel (approved):

PR #801, opened 2024-04-10
About us page

nshershnev:


trean (changes requested):
Thank you for PR!
There seems to be an issue with this section:

Screenshot 2024-04-11 at 14 49 57

nshershnev:
> Thank you for PR! There seems to be an issue with this section:
> 
> Screenshot 2024-04-11 at 14 49 57

@trean actually it follows the design in Figma (scrollable component from left to right for the whole screen). Do we need to have a special mode for carousel component for extra large screens in this case?

nshershnev:
@trean add the required changes to resize carousel component on the x-large screens

trean (approved):
Thank you!

PR #808, opened 2024-04-11
Add JWT RBAC documentation

coszio:
## TODO

- [x] Add access table (global read, global manage, collection read, collection read-write, payload constraint)


qdrant-landing/content/documentation/guides/security.md
```diff
@@ -132,6 +130,209 @@ export QDRANT__SERVICE__READ_ONLY_API_KEY=your_secret_read_only_api_key_here
...
+        "access": "r",
+        "payload": {
+          "user_id": "user_123456"
+        }
+      }
+    ]
+  }
+  ```
+
+  This `payload` claim will be used to implicitly filter the points in the collection, so that they only see their user, for example.
```
xzfc, line 262: Maybe add this explicitly:

    This is equivalent of appending this filter to each request:
    ```json
    {"filter": {"must": [{"key": "user_id", "match": {"value": "user_123456"}}]}}
    ```

qdrant-landing/content/documentation/guides/security.md
```diff
@@ -132,6 +130,209 @@ export QDRANT__SERVICE__READ_ONLY_API_KEY=your_secret_read_only_api_key_here
...
+
+Check out this table to see which actions are allowed or denied based on the access level.
+
+This is also applicable to using api keys instead of tokens. In that case, `api_key` maps to **manage**, while `read_only_api_key` maps to **read-only**.
+
+<div style="text-align: right"> <strong>Symbols:</strong> ✅ Allowed | ❌ Denied | 🎟️ Allowed, but filtered </div>
+
+
+
+| Action | manage | read-only | collection read-write | collection read-only | collection with payload claim (r / rw) |
```
xzfc, line 276: Is this table auto-generated from the `auth_tests`? If so, we need to document how to regenerate it.

And preferably to get it into the automation, so it wouldn't got outdated when we change rules.

qdrant-landing/content/documentation/guides/security.md
```diff
@@ -132,6 +130,209 @@ export QDRANT__SERVICE__READ_ONLY_API_KEY=your_secret_read_only_api_key_here
...
+          "user_id": "user_123456"
+        }
+      }
+    ]
+  }
+  ```
+
+  This `payload` claim will be used to implicitly filter the points in the collection, so that they only see their user, for example.
+
+<!-- TODO: add access table -->
```
xzfc, line 264: Isn't it's already added? 😅

xzfc (changes requested):

qdrant-landing/content/documentation/guides/security.md
```diff
@@ -132,6 +130,209 @@ export QDRANT__SERVICE__READ_ONLY_API_KEY=your_secret_read_only_api_key_here
...
+To use JWT-based authentication, you need to provide it as a bearer token in the `Authorization` header, or as an key in the `Api-Key` header of your requests.
+```json
+{ "Authorization": "Bearer <JWT>" }
+// or 
+{ "Api-Key": "<JWT>" }
+```
+
+#### Generating JSON Web Tokens
+
+Anyone who knows the `api_key` can generate tokens.
```
xzfc, line 165: Also, I think it is worth mentioning explicitly that:

- An user don't need to access to the Qdrant instance to generate the key.
  It could be generated offline using any of the existing libraries and tools.

- Qdrant Web UI provides a convenient UI to generate keys.

qdrant-landing/content/documentation/guides/security.md
coszio: No, I wrote it by hand. While automating it is a good idea, everything else in the docs is also manually maintained. Do you have a strong opinion on automating it?

qdrant-landing/content/documentation/guides/security.md
coszio: oops, not enough proof-reading

qdrant-landing/content/documentation/guides/security.md
xzfc: It is already present in a machine-readable form, so I think it would be a low-hanging fruit.

Maybe we could put the generation scripts into our repo in the same vein as autogenerated gRPC docs.

qdrant-landing/content/documentation/guides/security.md
```diff
@@ -132,6 +130,209 @@ export QDRANT__SERVICE__READ_ONLY_API_KEY=your_secret_read_only_api_key_here
...
+
+- **JWT Signature** - The signature is generated by encoding the header and payload with the `api_key` using the `HS256` algorithm.
+
+  ```
+  HMACSHA256(
+    base64UrlEncode(header) + "." +
+    base64UrlEncode(payload),
+    api_key
+  )
+  ```
```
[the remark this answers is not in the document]
coszio: There is no fault in being explicit, how about this?

> The signature is generated by base-64-url-encoding the header and payload, concatenating them with a `.`, and signing them with the `api_key` using the `HS256` algorithm.

qdrant-landing/content/documentation/guides/security.md
generall: Any reason people would prefer this table over just trying for themselves? I propose to minimize it to the groups of the operations, which should allow us to e.g. improve on filtered updates without breaking the documented interface.
Sometimes less explicit is better

qdrant-landing/content/documentation/guides/security.md
```diff
@@ -132,6 +130,216 @@ export QDRANT__SERVICE__READ_ONLY_API_KEY=your_secret_read_only_api_key_here
 
 Both API keys can be used simultaneously.
 
+### Granular access control with JWT
+
+*Available as of v1.9.0*
+
+For more complex cases, Qdrant supports granular access control with [JSON Web Tokens (JWT)](https://jwt.io/). This allows you to have [Role-based access control (RBAC)](https://en.wikipedia.org/wiki/Role-based_access_control) integrated into the system. In this way, you can define roles and permissions for users and restrict access to sensitive endpoints.
```
generall, line 137: this wiki link describes something completely different

timvisee (approved):

qdrant-landing/content/documentation/guides/security.md
coszio: @generall Are you talking about the autogenerated one?

generall (approved):

PR #850, opened 2024-04-26
Article and blog pages update

trean:
- added link to posts
- other missing links
- subscribe Hubspot form
- page-search plugin (require re-indexing, because the HTML structure changed and plugin gets wrong selector from the backend)


generall:
ToDo (@generall): search script on paginated pages

PR #855, opened 2024-04-30
Docs: add FAQ for not being able to open collection WAL

timvisee:
This describes what to do when the collection WAL cannot be opened.

It seems to be quite a common problem across the board.

### DIFF
--- a/qdrant-landing/content/documentation/guides/common-errors.md
+++ b/qdrant-landing/content/documentation/guides/common-errors.md
@@ -31,3 +31,24 @@ ulimit -n 10000
 ```
 
 Please note, the command should be executed before you run Qdrant server.
+
+## Can't open Collections meta Wal
+
+When starting a Qdrant instance as part of a distributed deployment, you may
+come across an error message similar to this:
+```bash
+Can't open Collections meta Wal: Os { code: 11, kind: WouldBlock, message: "Resource temporarily unavailable" }
+```
+
+It means that Qdrant cannot start because a collection cannot be loaded. Its
+associated [WAL](../../concepts/storage/#versioning) files are currently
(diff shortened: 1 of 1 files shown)


qdrant-landing/content/documentation/guides/common-errors.md
```diff
@@ -31,3 +31,24 @@ ulimit -n 10000
 ```
 
 Please note, the command should be executed before you run Qdrant server.
+
+## Can't open Collections meta Wal
+
+When starting a Qdrant instance as part of a distributed deployment, you may
+come across an error message similar to this:
+```bash
```
SamirPS, line 39: can you add a line break here?

qdrant-landing/content/documentation/guides/common-errors.md
```diff
@@ -31,3 +31,24 @@ ulimit -n 10000
...
+Can't open Collections meta Wal: Os { code: 11, kind: WouldBlock, message: "Resource temporarily unavailable" }
+```
+
+It means that Qdrant cannot start because a collection cannot be loaded. Its
+associated [WAL](../../concepts/storage/#versioning) files are currently
+unavailable, likely because the very same files are already being used by
```
SamirPS, line 45: not sure if we need the very word

SamirPS (changes requested):

qdrant-landing/content/documentation/guides/common-errors.md
timvisee: Used it to emphasize 'same files'. But I removed it now.

SamirPS (approved):

agourlay (approved):

PR #900, opened 2024-05-15
feat: Persist snippets language preference

### LINKED ISSUE #899: Choosing language in one snippet should change language in others
When I pick `python` as my language in my snippet, I should be updated in all the other snippets. 

This preference should be stored in the browser and be applied even when I refresh. Would reduce a lot of pain.

Anush008:
Resolves #899 

## [PREVIEW](https://deploy-preview-900--condescending-goldwasser-91acf0.netlify.app/documentation/concepts/collections/)

### DIFF
--- a/qdrant-landing/themes/qdrant/static/js/lang-switcher.js
+++ b/qdrant-landing/themes/qdrant/static/js/lang-switcher.js
@@ -1,4 +1,9 @@
 (function () {
+
+  const SWITCH_EVENT = 'switch-lang';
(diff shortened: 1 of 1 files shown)


Anush008:
> Let's try to implement it in the new theme. WDYT?

Yes. Thanks.

> Now if language changes on a snippet in the middle or the end of the page it makes the whole page jump.

Right. It's very evident and surprising.

> Maybe you know examples of sites where you like how similar functionality works?

Not yet. They're usually different pages per code block.

PR #910, opened 2024-05-17
Docs sticky sidebars

trean:
- Sticky side menu and TOC on the docs page


generall (changes requested):
- scroll bar is always visible
- left pannel can be bigger than screen and it is impossible to reach last items

trean:
Outdated and implemented in another place

PR #913, opened 2024-05-17
Expand advice on shard count and horizontal scaling

mac-chaffee:
Based on this conversation: https://qdrant.slack.com/archives/C04KX0FSW8G/p1715977792502279

This PR expands our recommendations about shard count, including a focus on horizontal scaling. When you horizontally scale, you have to have enough shards for each node, or you have to recreate the collection. So telling everyone to use at least 2x shards per node allows for expansion, with 12 shards being a good round number for most use-cases.

### DIFF
--- a/qdrant-landing/content/documentation/guides/distributed_deployment.md
+++ b/qdrant-landing/content/documentation/guides/distributed_deployment.md
@@ -278,7 +278,9 @@ await client.CreateCollectionAsync(
 );
 ```
 
-The number of shards must be a multiple of the number of nodes you are currently running in your cluster.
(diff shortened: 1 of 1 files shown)


Brian-Williams (approved):
LGTM

qdrant-landing/content/documentation/guides/distributed_deployment.md
```diff
@@ -275,9 +278,11 @@ await client.CreateCollectionAsync(
 );
 ```
 
-We recommend setting the number of shards to be a multiple of the number of nodes you are currently running in your cluster.
+The number of shards must be a multiple of the number of nodes you are currently running in your cluster.
```
generall, line 281: this is not a hard requirement. Some use-cases might require less shards than nodes

PR #915, opened 2024-05-18
replace mmap_threshold with on_disk=true param

generall:


qdrant-landing/content/documentation/concepts/storage.md
```diff
@@ -145,7 +145,7 @@ This will create a collection with all vectors immediately stored in memmap stor
 This is the recommended way, in case your Qdrant instance operates with fast disks and you are working with large collections.
 
 
-- Set up `memmap_threshold_kb` option. This option will set the threshold after which the segment will be converted to memmap storage.
+- Set up `memmap_threshold_kb` option (deprecated). This option will set the threshold after which the segment will be converted to memmap storage.
```
timvisee, line 148: Do we want to do soft deprecation? In the sense that we mark it as deprecated but we still keep it around?

timvisee (approved):

qdrant-landing/content/documentation/concepts/storage.md
generall: yeah, I propose we just hide this option eventually

PR #934, opened 2024-05-30
Add Qdrant for startups page

nastyapash:


trean (changes requested):
Thank you @nastyapash,
It looks good, only here one border is missing under the "Who is eligible" question (I suppose it's in a closed state when we can not see an answer text).
image

Also, could you please add right in this PR some js for:
1. Scrolling to the form when clicking on the `Apply Now` button
2. Accordion in the `Programm FAQ` section?

trean (approved):

PR #948, opened 2024-06-06
docs: updated langchain usage

JonZeolla:
langchain has split their packages up a bit, so the prior code examples no longer work.

There is also a `langchain-qdrant` package, which can also be used to access `Qdrant`, but we would have still needed to instruct the user to install `langchain-community` in order to get the huggingface embeddings, so this is the simplest walkthrough I could imagine.

### DIFF
--- a/qdrant-landing/content/documentation/frameworks/langchain.md
+++ b/qdrant-landing/content/documentation/frameworks/langchain.md
@@ -8,21 +8,20 @@ aliases:
 
 # LangChain
 
-LangChain is a library that makes developing Large Language Models based applications much easier. It unifies the interfaces
+LangChain is a library that makes developing Large Language Model-based applications much easier. It unifies the interfaces
 to different libraries, including major embedding providers and Qdrant. Using LangChain, you can focus on the business value
 instead of writing the boilerplate.
 
 Langchain distributes their Qdrant integration in their community package. It might be installed with pip:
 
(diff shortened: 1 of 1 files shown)


qdrant-landing/content/documentation/frameworks/langchain.md
```diff
@@ -8,22 +8,22 @@ aliases:
...
 ```
 
-Qdrant acts as a vector index that may store the embeddings with the documents used to generate them. There are various ways 
+Qdrant acts as a vector index that may store the embeddings with the documents used to generate them. There are various ways
 how to use it, but calling `Qdrant.from_texts` is probably the most straightforward way how to get started:
 
 ```python
-from langchain.vectorstores import Qdrant
-from langchain.embeddings import HuggingFaceEmbeddings
+from langchain_community.vectorstores import Qdrant
```
Anush008, line 25: How about replacing this import with `langchain-qdrant` instead? The community one is deprecated now.

Anush008 (changes requested):

Anush008:
Hi @JonZeolla. I've added the changes myself to not take up more of your time.
Thanks for contributing.

Anush008 (approved):
Hi @JonZeolla. I've added the changes to not take up more of your time.
Thanks for contributing.

PR #964, opened 2024-06-11
Adding Stauts Page to Footer

joekrejci:
Added https://status.qdrant.io/ to Footer under Developer Section.


fabribet (approved):
~Screenshots about how it looks like would be helpful ~

I just realized there is a preview :)

PR #977, opened 2024-06-18
Apply security headers X-* on html files only

Rendez:
relates to: https://github.com/qdrant/cloud-pm/issues/381#issuecomment-2173635087

Add headers only to HTML files and not every static asset, etc. I hope targeting .html file paths has the best outcome, alternatively we can later try to target `/` and `/admin`, but I believe `/` covers that and `/index.html` - we'll see after deployment.

https://docs.netlify.com/routing/headers/#syntax-for-the-netlify-configuration-file


Rendez:
Can't be done better because header paths are cummulative, and we can't be targetting URLs we know one by one.

PR #997, opened 2024-06-28
BLOG: Weaviate to Qdrant migration guide

kartik-gupta-ij:


davidmyriel:
@infoslack please check filtering; we might need to compare filtrable hnsw vs pre/post filtering

infoslack:
> @infoslack please check filtering; we might need to compare filtrable hnsw vs pre/post filtering
I did not understand this part.

davidmyriel:
> * It would be nice to add a link to the example code.
> * As an operations engineer (DevOps), I missed a tool to automate the process. Maybe you can include something at the end [like I did with Pinecone](https://github.com/infoslack/qdrant-example/blob/main/vector-migration/tutorial.md).

I don't think vdd-io has integrated it. A a PR was started months ago. Can you suggest some other tool?

infoslack:
> > * It would be nice to add a link to the example code.
> > * As an operations engineer (DevOps), I missed a tool to automate the process. Maybe you can include something at the end [like I did with Pinecone](https://github.com/infoslack/qdrant-example/blob/main/vector-migration/tutorial.md).
> 
> I don't think vdf-io has integrated it. A a PR was started months ago. Can you suggest some other tool?

I thought about Airbyte, but it would be too far from simple. I'd rather ignore that idea and leave it as it is.

Regarding filtering, I think what is mentioned about what Qdrant offers is enough. Since the idea is to migrate data only, I don't know if we need to make more comparisons. (I would ignore this difference in filtering in both as it will not impact the migration.)

infoslack (approved):
I think it's good; need the link to the demonstrated code and the final image.

PR #998, opened 2024-06-28
initial commit for segment introduction

jasonbryant84:
## Context
We are introducing Segment to the Marketing site for the sake of better Marketing/Growth Analytics [notion file](https://www.notion.so/qdrant/Data-ELT-Pipeline-Warehousing-with-Segment-Postgres-5f6807c20c0e4e5eb340ef49ed0b74a3). Users' events/behavior will be stored in Mixpanel and a Postgres Warehouse solution to tie it together with our other (business, CRM) analytics.

## Solution
We are loading Segment on the frontend when a user consents to tracking via the cookit.js banner. In the future we will likely implement Segment on the backend which will only require removing the Segment injection logic and swapping Segment's front end calls for backend endpoint API calls.

## Outstanding Issue(s)
- [ ] How can we implement an environment variable for the writeKey for Segment? We need production to write to a different Segment environment than staging or local development. I've attempted with `segmentWriteKey` in the `config.toml` file

### DIFF
--- a/qdrant-landing/config.toml
+++ b/qdrant-landing/config.toml
@@ -80,6 +80,7 @@ disableKinds = ["taxonomy", "term"]
   cloudDocVersion = "v0.1.x"
 
   googleTagManager = "GTM-KRLCXD5"
+  segmentWriteKey = ""
 
 [params.author]
   name = "Andrey Vasnetsov"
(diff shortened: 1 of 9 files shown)


generall (approved):
whatever

PR #1000, opened 2024-06-28
Revert "Add S3 config into example"

mogery:
Reverts qdrant/landing_page#543

The counterpart of that PR was not merged. This config is therefore incorrect.


generall:
agree, the s3 config is different

Thanks @mogery

timvisee:
I had already patched this in <https://github.com/qdrant/landing_page/pull/993>, but it is pending merge due to the release.

Thanks for the PR anyway!

PR #1031, opened 2024-07-17
Add documentation for config value storage.async_scorer

bashofmann:
See also https://qdrant.tech/articles/io_uring/#and-what-about-qdrant

### DIFF
--- a/qdrant-landing/content/documentation/guides/configuration.md
+++ b/qdrant-landing/content/documentation/guides/configuration.md
@@ -168,7 +168,10 @@ storage:
     # Default is to allow 1 transfer.
     # If null - allow unlimited transfers.
     #outgoing_shard_transfers_limit: 1
-
+    
+    # enable the async scorer which uses io_uring instead of a mmap based backend
+	#async_scorer: true
+      
   optimizers:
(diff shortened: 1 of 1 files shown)


timvisee (commented):
We did not document this on purpose because we thought it may be more confusing than helpful. It's only useful in specific tuning cases. However, I'm fine with documenting it.

Also, with each update we replace this section with that's in [`config.yaml`](https://github.com/qdrant/qdrant/blob/master/config/config.yaml).

So if we update it here, we should update it in the core repository too.

bashofmann:
Personal opinion: It feels weird to have a config option that is available and even mentioned in a blog post which is not documented in our docs. 
I'd rather then add a big warning text above this is only useful in very specific tuning cases.

@generall Do you want to have this documented?

If yes, happy to also add to to the config.yaml example in the main repo and open a PR there.

davidmyriel:
We need to document all options, no matter the edge case. I will merge this, but add a note for the user.

davidmyriel (approved):

davidmyriel:
added it here as well: https://github.com/qdrant/qdrant/pull/4702

PR #1039, opened 2024-07-21
Stream Data with Managed Kafka (Confluent) to Qdrant

davidmyriel:
This is a rough draft. 

- [x] Need to add front matter
- [ ] Preview image
- [x] Edit and proofread text


pavanjava (changes requested):
the qdrant docker image run command has error. the port number that is mapped is `6333` instead of `6334`. I have corrected that in my actual article this version need the change as well.

Issue #1048, opened 2024-07-23
Wrong units in multiple articles


morsik:
Some articles contains technical information with wrong units. In [some](https://qdrant.tech/articles/dedicated-service/) [places](https://qdrant.tech/articles/scalar-quantization/) there are `Gb` units (Gigabits) instead of clearly something that should be as `GB` (Gigabytes).

Images in first link also clearly shows wrong units.

kacperlukawski:
Thank you, @morsik. The PR is ready and will hopefully be merged soon.

PR #1056, opened 2024-07-26
Use proper units (gigabytes and megabytes, not gigabits and megabits)

kacperlukawski:
As reported in #1048, some articles used the wrong units (*bits, not *bytes). This PR fixes that.


morsik:
This PR is missing images ;)

That's why I didn't create this PR in the first place - because I'm no graphics artist so I didn't want to create half-fix ;(

davidmyriel (approved):

PR #1072, opened 2024-08-07
Article: Every model might be a late interaction model if you give it a chance

kacperlukawski:
This PR adds a draft of the article summarizing how to use standard sentence transformers as if they were late interaction models. Results on some of the BeIR datasets are pretty promising.


qdrant-landing/content/articles/late-interaction-models.md
```diff
@@ -0,0 +1,292 @@
...
+author: Kacper Łukawski
+author_link: https://kacperlukawski.com
+date: 2024-08-05T00:00:00.000Z
+---
+
+\* At least every Open Source model, because you need to access its internals.
+
+Qdrant 1.10 introduced support for multi-vector representations, and late interaction, are the most prominent example of 
+such models. In a nutshell, both documents and queries are represented by multiple vectors, and finding the most 
+relevant documents requires calculating a score based on similarity between the pairs of query and document embeddings.
```
Anush008, line 17:
```suggestion
relevant documents require calculating a score based on the similarity between the query and document embeddings pairs.
```

qdrant-landing/content/articles/late-interaction-models.md
```diff
@@ -0,0 +1,293 @@
...
+    # and return just the top-10. Please note we call the same model, but
+    # we ask for the token embeddings by setting the output_value parameter.
+    query=model.encode(query, output_value="token_embeddings").tolist(),
+    using="output-token-embeddings",
+    limit=10,
+)
+```
+
+In a real world scenario, you would probably go even step further and calculate the token embedding first, and then 
+perform pooling to get the single vector representation. This way you can do everything in a single pass.
```
Anush008, line 286:
```suggestion
In a real-world scenario, you would probably go even step further and calculate the token embedding first, and then 
perform pooling to get the single vector representation. This way you can do everything in a single pass.
```

Anush008 (commented):
Very nice read. 2 little nits from me.

qdrant-landing/content/articles/late-interaction-models.md
kacperlukawski: Thanks! Actually, "requires" refers to "finding" in that case, so I believe that was the correct form. However, I'm taking all the other changes :smiley:

PR #1074, opened 2024-08-08
FastEmbed Docs Pt. 2 - Supported Models

davidmyriel:
I added four sections for each of the model groups.
We are showing code how to retrieve the model lists.

The problem is that models change often, so we should just pick 1 or two of the best ones per embedding type.

@mrscoopers can you pick 2 from text embedding, 2 from sparse, 1 from late interaction (Colbert) and 1 from image embedding? Perhaps you can do a table instead of bullet points like I did. 

We should make a list that we don't have to change often.

### DIFF
--- a/qdrant-landing/content/documentation/fastembed/fastembed-models.md
+++ b/qdrant-landing/content/documentation/fastembed/fastembed-models.md
@@ -1,0 +1,99 @@
+---
+title: "Supported Models"
+weight: 4
+---
+
+# Supported Models
+
+FastEmbed supports a variety of models for different tasks, including dense text embeddings, sparse text embeddings, late interaction text embeddings, and image embeddings. Here are some examples of the models supported:
+
+## How to list all available models
+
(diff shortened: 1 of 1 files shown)


joein:
Having code to look at the supported models and installing pandas for it looks like a bit of an overkill :(
I think a manually updated table is better, however, it is not perfect as well :(

mrscoopers:
@davidmyriel We can do a model per provider, it's just 11 textual models now if not to consider different sizes/versions/languages (CLIP, e5, MPNet, BAII/bge, paraphrase-multilingual-MiniLM-L12-v2, gte-large, mxbai-embed-large-v1, snowflake-arctic-embed, nomic-embed-text-v1.5, all-MiniLM-L6-v2 & Jina)
Afaik we don't upload models too often, that's why text descriptions of models & their capabilities seem feasible-ish, and then people can do some selection based on the provided info: we can add retrieval benchmark results, prompts required to make retrieval work, etc

PR #1080, opened 2024-08-13
Version 1.11 Blog Post

davidmyriel:
Everything should be there except two things:

- Snippets (@Anush008)
- YouTube video voiceover (@davidmyriel)

There is too much content, we should talk about immutability in a separate article.


Anush008 (approved):

timvisee (approved):
Nice! I left my two-cents as review comments.

Note that I've added a commit to correct the size of UUIDs in string representation.

PR #1101, opened 2024-08-16
[SEO] Vector Search Revamped

davidmyriel:
Suggestions as per SEO agency. Nothing special, just the title and headings....few links here and there.


kacperlukawski (approved):
I don't necessarily like the new title, but I guess it's important for SEO.

davidmyriel:
> I don't necessarily like the new title, but I guess it's important for SEO.

I know. I just made one more change. I hope you like this one better @kacperlukawski

PR #1116, opened 2024-08-20
[Docs] Multimodal Search with FastEmbed

davidmyriel:
This PR adds a simple multimodal search tutorial.


mrscoopers:
I think the link to Collab in the markdown header should be pointing somewhere where it's not a private account but at least an organization account(?)

mrscoopers (changes requested):
Added everything, I think(?)

mrscoopers (approved):

PR #1123, opened 2024-08-23
Clean up embedding model providers on the website

thierrypdamiba:
Made some formatting changes to the list of embedding model providers


davidmyriel:
Taken care of

PR #1133, opened 2024-08-28
Prevent empty pages rendering

trean:
We have some markdown files that are not intended to be displayed as site pages. Although they are excluded from the site map, they can still be accessed through a link. This pull request adds Hugo parameters to these files to prevent them from being built.


davidmyriel (approved):
lgtm

PR #1158, opened 2024-09-10
Quantization article - first draft

sabrinaaquino:
[Rendered](https://deploy-preview-1158--condescending-goldwasser-91acf0.netlify.app/articles/what-is-vector-quantization/)

Version to review. 

Content Todos:
- [x] Basic Quantization example with use case (needs to be a story for the user) - The article already covers the basics. Adding more basic examples might risk redundancy
- [ ] Advanced Quantization example if possible - I feel like this should be a separate notebook that we can link to since the article already covers all methods + best practices. 
- [x] Oversampling can be maybe better described.
- [x] Best Practices section (based on [tips here](https://qdrant.tech/documentation/guides/quantization/#quantization-tips) + more)
- [x] Rescoring section in more detail.

Misc Todos:
- [ ] Headings and subheadings.
- [x] Images and alt text for SEO.
- [ ] Quantization tutorials in Web UI.

Conclusion: All the theory is in one place. We need to show the reader what "they can do" with quantization.


davidmyriel:
Left todos/comments above.

qdrant-landing/content/articles/what-is-quantization.md
```diff
@@ -0,0 +1,242 @@
...
+
+There are several methods to achieve this, and here we will focus on three main ones:
+
+
+![Types of Quantization: 1. Scalar Quantization, 2. Product Quantization, 3. Binary Quantization](/articles_data/what-is-vector-quantization/types-of-quantization.png)
+
+
+## 1. What is Scalar Quantization?
+
+In Qdrant, each vector is represented by a float32 value, which uses **4 bytes** of memory. When using [Scalar Quantization](https://qdrant.tech/documentation/guides/quantization/#scalar-quantization), we are mapping our vectors to a range that the smaller int8 type can represent. An int8 can store 256 values (from -128 to 127) which uses only **1 byte.** This typically results in a **75% reduction** in memory size.
```
kacperlukawski, line 61:
```suggestion
In Qdrant, each dimension is represented by a float32 value, which uses **4 bytes** of memory. When using [Scalar Quantization](https://qdrant.tech/documentation/guides/quantization/#scalar-quantization), we are mapping our vectors to a range that the smaller int8 type can represent. An int8 can store 256 values (from -128 to 127, or 0 to 255) which uses only **1 byte.** This results in a **75% reduction** in memory size.
```

Memory size is always reduced by 75%.

qdrant-landing/content/articles/what-is-quantization.md
```diff
@@ -0,0 +1,242 @@
...
+
+
+![Codebook and Centroids example](/articles_data/what-is-vector-quantization/codebook.png)
+
+
+Qdrant always selects **K = 256** for the number of centroids in its implementation based on the fact that 256 is the maximum number of unique values that can be represented by a single byte.
+
+This makes the compression process efficient because each centroid index can be stored in a single byte.
+
+After the codebook is created, the original high-dimensional vectors are split into smaller **sub-vectors.**
```
kacperlukawski, line 130: The process description seems to be inaccurate. As far as I know, the codebook is created separately for each subvector, so we have different centroids for each segment. That has to be confirmed with the core team, but I think it makes sense to do it separately, as different segments of the original vectors may have different characteristics.

qdrant-landing/content/articles/what-is-quantization.md
```diff
@@ -0,0 +1,242 @@
...
+    "params": {
+        "quantization": {
+            "rescore": True
+        }
+    },
+    "limit": 10
+}
+```
+
+Product Quantization is generally more memory efficient than **Scalar Quantization**, being able to reduce the memory footprint by up to **64x** while still retaining much of the essential information. However, the computational overhead of mapping sub-vectors to centroids results in the **slowest performance** compared to the other methods.
```
kacperlukawski, line 157: Should we describe this method to retain much of the essential information, especially if we mention the 64x config in the same sentence? Overall, the quality drops quite significantly in our benchmarks:

image

Also, PQ always results in slower indexing, but the search might be even faster than with the original vectors:

image

qdrant-landing/content/articles/what-is-quantization.md
```diff
@@ -0,0 +1,242 @@
...
+
+If your application requires high precision or real-time performance, this slower computation might be a significant factor to consider.
```
kacperlukawski, line 159: Same as above. In some scenarios, PQ might be faster than original vectors.

qdrant-landing/content/articles/what-is-quantization.md
```diff
@@ -0,0 +1,242 @@
...
+    },
+    "quantization_config": {
+        "binary": {
+            "always_ram": True
+        }
+    }
+}
+```
+
+Binary Quantization is by far the quantization method that will give you the most processing **speed gains **when compared to Scalar and Product Quantizations. This is because the binary representation allows the system to use highly optimized CPU instructions, such as [XOR](https://en.wikipedia.org/wiki/XOR_gate#:~:text=XOR%20represents%20the%20inequality%20function,the%20other%20but%20not%20both%22.) and [Popcount](https://en.wikipedia.org/wiki/Hamming_weight), for fast distance computations.
```
kacperlukawski, line 201:
```suggestion
Binary Quantization is by far the quantization method that will give you the most processing **speed gains** when compared to Scalar and Product Quantizations. This is because the binary representation allows the system to use highly optimized CPU instructions, such as [XOR](https://en.wikipedia.org/wiki/XOR_gate#:~:text=XOR%20represents%20the%20inequality%20function,the%20other%20but%20not%20both%22.) and [Popcount](https://en.wikipedia.org/wiki/Hamming_weight), for fast distance computations.
```

The rendering is kind of strange. Hopefully, this change will fix it.

image

qdrant-landing/content/articles/what-is-quantization.md
```diff
@@ -0,0 +1,242 @@
...
+
+It can speed up search operations by **up to 40x**, depending on the dataset and hardware. Here's the speed comparison of all three methods:
+
+<img src="/articles_data/what-is-vector-quantization/speed-methods.png" alt="Speed by Quantization Method" width="700">
```
kacperlukawski, line 205: I think this image needs some refreshing. We don't know what settings were used for PQ. Should I run a benchmark and measure it?

qdrant-landing/content/articles/what-is-quantization.md
```diff
@@ -0,0 +1,242 @@
...
+    "params": {
+        "quantization": {
+            "rescore": True
+        }
+    },
+    "limit": 10
+}
+```
+
+When you enable rescoring in Qdrant, it refines the top search results by recalculating their similarity with the original floating-point vectors.
```
kacperlukawski, line 237: It's an important paragraph that deserves some more attention. Qdrant keeps the original vectors, so there is a way back even if someone enables quantization on a specific collection.

kacperlukawski (commented):
I left some comments. The ones about PQ are probably the most important.

qdrant-landing/content/articles/what-is-quantization.md
sabrinaaquino: Yes, I'm still working on the rescoring and reranking explanations!

qdrant-landing/content/articles/what-is-quantization.md
sabrinaaquino: If you could that'd be great! I took this from our BQ article so I can't tell @kacperlukawski

qdrant-landing/content/articles/what-is-quantization.md
sabrinaaquino: In our benchmarks, what's the compression we used? I think 0.7 precision would be for 64x, correct? I'd assume for smaller compression the loss would be less significant. If you have a notebook I can use for some benchmarks that'd be great

sabrinaaquino: Also, can you clarify on why is the speed 0.5? Is that including the indexing?

qdrant-landing/content/articles/what-is-quantization.md
```diff
@@ -0,0 +1,249 @@
...
+
+The **quantile** is used to calculate the quantization bounds. For example, if you specify 0.99 as the quantile, 1% of extreme values will be excluded from the quantization bounds. 
+
+This parameter only affects the resulting precision, not the memory footprint. You can tune it if you experience a significant decrease in search quality.
+
+The primary benefit of Scalar Quantization is **memory reduction.** It is especially useful for large-scale datasets, where memory and processing power become limiting factors. It also slighly improves performance, as distance calculations (such as dot product or cosine similarity) using int8 values are computationally simpler than using float32 values. 
+
+However, these performance gains are significantly lower compared to Binary Quantization, which we'll discuss later.
+
+# 2. What is Product Quantization?
```
IvanPleshkov, line 96: I strongly recommend telling first about BQ and then about PQ. PQ isn't good technique in practice and it's important to tell about BQ while the reader is not tired and doesn't stop reading

IvanPleshkov: In my personal opinion, we can only mention that PQ exists, attach the link to documentation, and don't spend users' time and energy for this. But I don't think my colleagues will agree

qdrant-landing/content/articles/what-is-quantization.md
```diff
@@ -0,0 +1,249 @@
...
+![Types of Quantization: 1. Scalar Quantization, 2. Product Quantization, 3. Binary Quantization](/articles_data/what-is-vector-quantization/types-of-quantization.png)
+
+
+## 1. What is Scalar Quantization?
+
+![](/articles_data/what-is-vector-quantization/astronaut-mars.jpg)
+
+In Qdrant, each dimension is represented by a float32 value, which uses **4 bytes** of memory. When using [Scalar Quantization](https://qdrant.tech/documentation/guides/quantization/#scalar-quantization), we are mapping our vectors to a range that the smaller int8 type can represent. An int8 can store 256 values (from -128 to 127, or 0 to 255) which uses only **1 byte.** This results in a **75% reduction** in memory size.
+
+For example, if our data lies in the identified range of -1.0 to 1.0, Scalar Quantization will transform these values to a range that int8 can represent, that is, within -128 to 127. So, the system **maps** the float32 values into this range.
```
IvanPleshkov, line 64: I checked SQ article https://qdrant.tech/articles/scalar-quantization/
And I see a common mistake. In SQ we use range `[0..127]`. Yeah, we skip one bit from a byte in SQ. We did that to achieve better SIMD acceleration. I don't know what to do: fix in both places or just say "nevermind" =)

qdrant-landing/content/articles/what-is-quantization.md
```diff
@@ -0,0 +1,249 @@
...
+        }
+    }
+}
+```
+
+The **quantile** is used to calculate the quantization bounds. For example, if you specify 0.99 as the quantile, 1% of extreme values will be excluded from the quantization bounds. 
+
+This parameter only affects the resulting precision, not the memory footprint. You can tune it if you experience a significant decrease in search quality.
+
+The primary benefit of Scalar Quantization is **memory reduction.** It is especially useful for large-scale datasets, where memory and processing power become limiting factors. It also slighly improves performance, as distance calculations (such as dot product or cosine similarity) using int8 values are computationally simpler than using float32 values.
```
IvanPleshkov, line 92: Did you write about the primary benefit just because you compared SQ with "40x times faster BQ", right? =)

When we published an article about BQ and 40x times measurement, we compared on-disk float vectors with RAM-compressed vectors and became such a measurement. We did that comparison because we chose a case where vectors don't fit RAM, but BQ was fine for RAM.
If we compare on-disk float vectors with RAM SQ, we also will get 10-20 times faster.

We need to think more about the performance diagram above because it was a completely different measurement.

It means, that `slighly improved performance` is an underestimation and the primary benefit is not the memory only. We can say that for PQ but I don't recommend making anti-advertisement for SQ

IvanPleshkov: It's not trivial to compare SQ and BQ directly and we cannot just say like "BQ is better". It depends. If we need to achieve high accuracy, we need to set high oversampling and may become even worse search performance as SQ because of rescoring with on-disk vectors

qdrant-landing/content/articles/what-is-quantization.md
IvanPleshkov: I find this diagram not correct because it was different measurements explained in an another comment what's wrong with this comparison. in short: BQ here is a comparison between disk and RAM vectors, SQ here is a comparison between RAM and RAM

qdrant-landing/content/articles/what-is-quantization.md
```diff
@@ -0,0 +1,249 @@
...
+```
+
+Binary Quantization is by far the quantization method that will give you the most processing **speed gains** when compared to Scalar and Product Quantizations. This is because the binary representation allows the system to use highly optimized CPU instructions, such as [XOR](https://en.wikipedia.org/wiki/XOR_gate#:~:text=XOR%20represents%20the%20inequality%20function,the%20other%20but%20not%20both%22.) and [Popcount](https://en.wikipedia.org/wiki/Hamming_weight), for fast distance computations.
+
+It can speed up search operations by **up to 40x**, depending on the dataset and hardware. Here's the speed comparison of all three methods:
+
+<img src="/articles_data/what-is-vector-quantization/speed-methods.png" alt="Speed by Quantization Method" width="700">
+
+
+But it’s important to note that not all models are equally compatible with Binary Quantization. Some models may suffer a larger loss in accuracy when quantized, so it's recommended to use it on models of **at least 1024 dimensions.**
```
IvanPleshkov, line 213: As I remember, cohere 768-embeddings work fine with BQ. I propose to ask someone from ML team if we can decrease recommended dimensions to 768 or even 512

IvanPleshkov (changes requested):
I'm requesting changes because of SQ description and non-completed part about oversampling & restoring.

Also, I find very(!!) important to tell about disk&RAM. If a user has RAM vectors and he turns on quantization - we won't have reduced RAM, he has to set on-disk flag manually. Moreover, quantization by default is stored as original vectors. And `always_ram` changes this behaviour and forces to store quantization in RAM - we need to tell about it too. It's untrivial and important logic

qdrant-landing/content/articles/what-is-quantization.md
sabrinaaquino: You're right, that’s an important detail I hadn’t fully noticed before. I’ve applied the changes in the new commit

qdrant-landing/content/articles/what-is-quantization.md
```diff
@@ -0,0 +1,249 @@
...
+The quantization strips away some detail from the original vector. However, Qdrant mitigates this with **oversampling**, which retrieves extra vectors during the search and then uses the original values to rescore them, improving the accuracy of the final results.
+
+Here’s how you can enable rescoring to maintain higher accuracy:
+
+```python 
+search_payload = {
+    "query": [0.22, -0.01, -0.98, 0.37, ...],
+    "params": {
+        "quantization": {
+            "rescore": True
```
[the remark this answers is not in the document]
sabrinaaquino: I go more in-depth into oversampling and rescoring in another section.

qdrant-landing/content/articles/what-is-quantization.md
sabrinaaquino: I think the key is informing the user how it works and why it might not always be the best option, just like we know about it. The article consistently highlights the drawbacks.

sabrinaaquino: I've adjusted the sequence

qdrant-landing/content/articles/what-is-quantization.md
sabrinaaquino: Is there a specific reason why that is? Wouldn't the SIMD still work with the full range of an int8 (from -128 to 127)?

qdrant-landing/content/articles/what-is-quantization.md
sabrinaaquino: based on our website benchmarks, scalar shows up to 2x speed improvement. I’ve adjusted the wording to avoid suggesting that it’s slow, but I’m not sure if we should present different numbers than what’s in the documentation

qdrant-landing/content/articles/what-is-quantization.md
sabrinaaquino: I've added a Disk & RAM Storage section!

timvisee (commented):
I really liked the article! I feel like the way it describes the quantization methods makes sense, and it's easily understandable.

I left some review comments with thoughts/improvements. Most are nitpicks though, so take them with a grain of salt and apply what you feel works best.

qdrant-landing/content/articles/what-is-quantization.md
```diff
@@ -0,0 +1,552 @@
...
+
+---
+
+> Vector quantization is a technique for reducing the size of high-dimensional data. Instead of storing the original vectors, the data is stored in a compressed format, which reduces memory usage while maintaining most of the essential information. This compression allows for more efficient storage and faster search operations, especially in large datasets.
+
+When working with high-dimensional vectors, such as embeddings from AI models like OpenAI, one single 1536-dimensional vector requires **6KB of memory**.
+
+<img src="/articles_data/what-is-vector-quantization/vector-size.png" alt="1536 dimentional vector size is 6KB" width="700">
+
+If your dataset has **millions of vectors**, the memory and processing demands are very significant.
```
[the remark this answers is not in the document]
sabrinaaquino: For sure! but since we're talking about millions of vectors, not just one million, I've adjusted some of the wording:

With 1 million vectors needing around 6GB of memory, as your dataset grows to multiple millions of vectors, the memory and processing demands increase significantly.

qdrant-landing/content/articles/what-is-quantization.md
```diff
@@ -0,0 +1,552 @@
...
+
+To understand why this process is so computational demanding, let's take a look at the nature of the [HNSW index.](https://qdrant.tech/documentation/concepts/indexing/#vector-index)
+
+
+The **HNSW (Hierarchical Navigable Small World) index** organizes vectors in a layered graph, connecting each vector to its nearest neighbors. At each layer, the algorithm narrows down the search area until it reaches the lower layers, where it efficiently finds the closest matches to the query.
+
+
+<img src="/articles_data/what-is-vector-quantization/hnsw.png" alt="HNSW Search visualization" width="500">
+
+So each time a new vector is added, the system must determine its position in the existing graph, a process similar to searching. This makes both inserting and searching for vectors complex operations.
```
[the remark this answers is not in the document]
sabrinaaquino: I would love to know more about it. Since I don't know how to explain it better, I'll leave it, and we can possibly revisit this later.

qdrant-landing/content/articles/what-is-quantization.md
```diff
@@ -0,0 +1,552 @@
...
+
+You can adjust the `oversampling` factor to find the right balance between search speed and result accuracy.
+
+If quantization is affecting performance in an application that needs high accuracy, combining oversampling with rescoring is a great choice. But if you need faster searches and can tolerate some loss in accuracy, you might choose to use oversampling without rescoring, or adjust the oversampling factor to a lower value.
+
+
+### Disk & RAM Storage Tuning
+
+
+Qdrant stores both the quantized and original vectors. When you enable quantization, both the original and quantized vectors are stored in RAM by default. If you want to reduce RAM usage, just enabling quantization won't be enough—you need to explicitly move the original vectors to disk by setting `on_disk=True`.
```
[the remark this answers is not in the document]
sabrinaaquino: Great add!

qdrant-landing/content/articles/what-is-quantization.md
```diff
@@ -0,0 +1,552 @@
...
+    quantization_config=models.BinaryQuantization(
+        binary=models.BinaryQuantizationConfig(
+            always_ram=True  # Store only quantized vectors in RAM
+        )
+    )
+)
+```
+
+Without explicitly setting `on_disk=True`, you won't see any RAM savings, even with quantization enabled. So, ensure that you configure both storage and quantization options based on your memory and performance needs. If your storage has high disk latency, you can try disabling rescoring to maintain speed.
+
```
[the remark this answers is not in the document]
sabrinaaquino: I think this is a great add to the rescoring section! Just added

qdrant-landing/content/articles/what-is-quantization.md
```diff
@@ -0,0 +1,552 @@
...
+#### Speeding Up Rescoring with io_uring
+
+When dealing with large collections of quantized vectors, frequent disk reads are required to retrieve both original and compressed data for rescoring operations. Traditional methods like `mmap` involve context switching between user and kernel space, adding latency to each disk access, which can slow down rescoring for large datasets.
```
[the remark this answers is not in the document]
sabrinaaquino: I've adjusted it in the last commit

davidmyriel (approved):

agourlay:
I really enjoyed reading this article, fantastic job :clap: 

My only nitpick is that I am not sure I would recommend product quantization to anyone in practice.
It is rather expensive to compute k-means on large dataset.

Maybe in the section `when to use` we could reiterate the fact that it is not for system which needs fast indexing.

timvisee (approved):

IvanPleshkov (approved):

PR #1159, opened 2024-09-10
Qdrant relari

thierrypdamiba:
Content Todos:
- [ ] In depth explanation of what the metrics mean. We defined two main eval metrics and how to implement them. Then we stopped.
- [ ] What the user can do with these metrics. Can they somehow improve their use of Qdrant?
- [ ] Tell the user what to do to improve their RAG App

Misc Todos:
- [ ] Needs images for preview and sections
- [ ] Possibly better headings
- [ ] Take out marketing from footer and offer more help to the reader


davidmyriel:
another PR merged already

PR #1163, opened 2024-09-12
chore: update Jina Embedding doc for v3 release

DresAaron:


DresAaron:
This one is not necessary because of #1164. I will just close it.

PR #1169, opened 2024-09-14
Fix typo in multiple-partitions documentation

Jensssen:
Typo is present in python example for creating payload index.
`keywprd` vs `keyword`


generall:
Thanks!

PR #1184, opened 2024-09-18
[Docs] Add Self Query Tutorial

davidmyriel:


qdrant-landing/content/documentation/tutorials/self-query.md
```diff
@@ -0,0 +1,226 @@
...
+)
+```
+
+Each wine review is uploaded as a point in the collection, with both the vector representation and metadata attached to it.
+
+---
+
+### Step 6: Self-Query Retrieval in Action
+
+Let’s use **self-query retrieval** to allow the system to infer relevant filters and conditions from a natural language query. Suppose we want to find US wines priced between $15 and $30 with a rating (points) above 90. Instead of manually constructing the query, the system will automatically parse these constraints.
```
Anush008, line 182: I found this to be a problem.
The tutorial doesn't show how to infer these filters at all.

PR #1201, opened 2024-09-25
[Docs] Optimize Search 

davidmyriel:


qdrant-landing/content/documentation/guides/optimize.md
```diff
@@ -737,19 +750,20 @@ client.Query(context.Background(), &qdrant.QueryPoints{
 })
 ```
 
-- `hnsw_ef` - controls the number of neighbors to visit during search. The higher the value, the more accurate and slower the search will be. Recommended range is 32-512.
-- `exact` - if set to `true`, will perform exact search, which will be slower, but more accurate. You can use it to compare results of the search with different `hnsw_ef` values versus the ground truth.
+## 4. Balancing Latency and Throughput
+
+When optimizing search performance, ;atency and throughput are two main metrics to consider:
```
Anush008, line 755:
```suggestion
When optimizing search performance, latency and throughput are two main metrics to consider:
```

qdrant-landing/content/documentation/guides/optimize.md
```diff
@@ -737,19 +750,20 @@ client.Query(context.Background(), &qdrant.QueryPoints{
 })
 ```
 
-- `hnsw_ef` - controls the number of neighbors to visit during search. The higher the value, the more accurate and slower the search will be. Recommended range is 32-512.
-- `exact` - if set to `true`, will perform exact search, which will be slower, but more accurate. You can use it to compare results of the search with different `hnsw_ef` values versus the ground truth.
+## 4. Balancing Latency and Throughput
```
Anush008, line 753:
```suggestion
## Balancing Latency and Throughput
```

Should be 3 points as per the intro.

qdrant-landing/content/documentation/guides/optimize.md
```diff
@@ -737,19 +750,20 @@ client.Query(context.Background(), &qdrant.QueryPoints{
...
+The following optimization approaches are not mutually exclusive, but in some cases it might be preferable to optimize for one or another.
 
-- There are two main approaches to measure the speed of search:
-  - latency of the request - the time from the moment request is submitted to the moment a response is received
-  - throughput - the number of requests per second the system can handle
+### Minimizing Latency
 
-Those approaches are not mutually exclusive, but in some cases it might be preferable to optimize for one or another.
+To minimize latency, you can set up Qdrant to use as many cores as possible for a single request.
+You can do this by setting the number of segments in the collection to be equal to the number of cores in the system.
```
Anush008, line 764: Different than what L898 suggests.

davidmyriel: @timvisee Can you take a quick look at this?

I am rewriting the optimization doc for clarity, but this part is a little tough for the reader.

timvisee: That's correct. A shard has segments. If searching on a shard, all of it's segments are searched concurrently (on different CPU cores).

In fact, the Qdrant configuration has a property to define the number of shards. That's the number of CPUs by default.

There are some nuances, as always. But I don't think that's important to cover here.

timvisee (approved):

PR #1203, opened 2024-09-25
Fix quickstart for Python due to fetching payload by default

agourlay:
A python user was confused on Discord when following the quickstart to get a different result.

This is because the Python client fetches payload by default.

Regarding other clients, I just checked the Rust one and it does not request payload by default.


joein (approved):

Anush008 (approved):

agourlay:
Thanks for checking!

PR #1220, opened 2024-10-04
fix: Internal link checker, Use abs links

Anush008:
https://github.com/qdrant/landing_page/blob/master/.github/workflows/internal-dead-links.yml isn't able to detect relative bad links because these are written into the HTML as is. Without the `"http://localhost:6333"` prefix.
Screenshot 2024-10-04 at 5 06 00 PM

So these are excluded as per our workflow command.
```
--exclude '.*' --include 'http://localhost:1313/.*' qdrant-landing/public/
 ```
 
 Lychee supports checking such links. See https://lychee.cli.rs/recipes/relative-links/. We can do
 
 ```
--exclude '.*' --include 'http://localhost:1313/.*' --base http://localhost:1313/ qdrant-landing/public/
 ```
 
 But this only works for absolute links like `/documentation/tutorials/bulk-upload/`.
 Not relative ones like `./bulk-upload/`. As these are resolved to `http://localhost:1313/bulk-upload`. Which is wrong.
 
 I can replace all the relative links in the docs with absolute ones(Which is a good thing generally as we can move files) and we can start using this check in the CI. What do you think?


generall:
> I can replace all the relative links in the docs with absolute ones(Which is a good thing generally) and we can start using this check in the CI. What do you think?

I think this is a good idea. We might have other broken relative links out there, so if we can enforce uniform format for links with CI - that would be nice

generall (approved):

PR #1285, opened 2024-11-14
Dev portal fixes

trean:
In PR:
- added requested fixes @generall,
- updated the cards partial so it become more universal,
- moved some files.


trean:
Force-push after resolving conflicts through rebasing to the `create-dev-portal-pages`

generall (approved):

PR #1293, opened 2024-11-15
Hero button text and link upd

trean:
requested by marketing


generall (approved):
approve, but I don't like it @maddie-qdrant

PR #1306, opened 2024-11-19
remove scrollbar in docs menu

generall:


davidmyriel (approved):

timvisee:
I actually do like the scrollbar, partially because I use it and partially for accessibility.

How about only showing the scrollbar if the sidebar is hovered. Would that be a good middle ground?

https://stackoverflow.com/a/8631811/1000145

timvisee:
> > How about only showing the scrollbar if the sidebar is hovered. Would that be a good middle ground?
> 
> this won't work as scroll itself does have width. The whole content will jump around on hover. It is even reproducible in the stackoverflow answer itself. Ideal solution would be to use something like https://www.radix-ui.com/primitives/docs/components/scroll-area but it only comes as a part of framework

It appears to be broken in Chrome based browsers. It works totally fine with an overlay scrollbar in Firefox :cry: 


https://github.com/user-attachments/assets/f1e80074-e992-47b2-be37-1d569b9bd0f7

PR #1348, opened 2024-12-11
fix: Use .Site.Config.Services.GoogleAnalytics.ID instead.

Anush008:
- Fixes:

```console
ERROR deprecated: .Site.GoogleAnalytics was deprecated in Hugo v0.120.0 and will be removed in Hugo 0.135.0. Use .Site.Config.Services.GoogleAnalytics.ID instead.
```

- I've also run: `npm audit fix`

### DIFF
--- a/qdrant-landing/config.toml
+++ b/qdrant-landing/config.toml
@@ -3,7 +3,6 @@ languageCode = "en-us"
 title = "Qdrant - Vector Database"
 theme = "qdrant-2024"
 
-googleAnalytics = "G-NZYW2651NE"
 enableRobotsTXT = true
 canonifyURLs = true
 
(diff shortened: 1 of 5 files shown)


trean (changes requested):
Hey @Anush008, thanks for updating this! Would you mind to update `config.toml` too [like here](https://gohugo.io/templates/embedded/#configure-google-analytics)? 
Your update works, but Hugo uses the `googleAnalytics` param from the top level of the config because Hugo has [this fallback](https://github.com/gohugoio/hugo/blob/master/config/services/servicesConfig.go#L83-L87) for backward compatibility. It's ok, just let's prepare for the future and make config canonical 🙂

trean (approved):

PR #1365, opened 2025-01-02
Hybrid search: disable HNSW for late interaction

kacperlukawski:
As requested, the hybrid search article is extended with comments regarding using late interaction models for reranking. Building the HNSW graph is heavy and should be disabled if nobody runs a search with multi-vectors. 

@davidmyriel, I didn't add a search diagram, as you wanted to fix images independently.


davidmyriel (approved):
lgtm, feel free to merge and ill add images for all articles over the next few days

PR #1377, opened 2025-01-10
ColPali/ColQwen2 tutorial

mrscoopers:


davidmyriel (approved):

generall (approved):
I would prefer either remove or re-draw excalidraw image, otherwise it should be fine

PR #1395, opened 2025-01-16
[stars_list] updated my job position

nickprock:
Hi guys, 
Since this month I work in a new company.
I changed my job position and description in stars list.
Thanks in advance


davidmyriel:
Thanks for the update Nicola

PR #1412, opened 2025-01-23
Removing GTM and GA script, they are to be used trough the Segment now

trean:
Marketers now connect GTM and GA via Segment, so we no longer need a GTM script on the site. 

Don't merge without notifying @Alyonkka, she will check if everything works as expected.


jasonbryant84:
we have a meeting on Friday with the agency that I think was central to setting this up. We can confirm then with them if things should be unaffected and continue functioning as needed/expected. Hope that's ok. Thanks.

jasonbryant84 (commented):
LGTM. We got the greenlight from the consulting marketing agency that we're good to remove this as well.

PR #1426, opened 2025-01-26
[guides] provided missing information on how to enable quantization o…

itinance:
…n existing collections

as written in title. I added corresponding sections under 
- setting up scalar quantization
- setting up binary quantization
- setting up product quantization


generall (approved):

generall:
Thanks!

PR #1443, opened 2025-02-03
[article] The Gridstore

davidmyriel:
[PREVIEW](https://deploy-preview-1443--condescending-goldwasser-91acf0.netlify.app/articles/gridstore-key-value-storage/)

I have tried very hard to make this accessible to the general reader. In the process, I may have simplified/misinterpreted a few things (like the region). Please check and rephrase - but do not increase the size of the paragraph.

Its already too long.

ToDos:

- Better diagrams?

- Better description of the third component (Region, Tracker?)

- Conclusion


coszio (approved):
I am happy with the result. Is it possible to keep the final gif without the watermark?

PR #1460, opened 2025-02-12
new article on optimizing memory for bulk uploads

sabrinaaquino:


qdrant-landing/content/articles/indexing-optimization.md
```diff
@@ -0,0 +1,199 @@
...
+}
+```
+
+This approach immediately places all incoming vectors on disk. You can configure a threshold so that small segments are kept in RAM, but large segments automatically move to memmap storage once they exceed that threshold. This is controlled by the `memmap_threshold` in the collection’s `optimizers_config`.
+
+```json
+PATCH /collections/your_collection
+{
+    "optimizers_config": {
+        "memmap_threshold": 20000
```
generall, line 96: `memmap_threshold` is going to be deprecated, there are no benefits in using it

sabrinaaquino: yeah, removed this part

qdrant-landing/content/articles/indexing-optimization.md
```diff
@@ -0,0 +1,188 @@
...
+
+Let’s take a look at the best practices and recommendations to help you optimize memory usage during bulk uploads in Qdrant. We'll cover scenarios with both **dense** and **sparse** vectors, helping your deployments remain performant even under high load and avoiding out-of-memory errors.
+
+---
+
+### Indexing for dense vs. sparse vectors
+
+**Dense vectors**  
+
+Qdrant employs an **HNSW-based index** for fast similarity search on dense vectors. By default, HNSW is built or updated once the number of **unindexed** vectors in a segment exceeds a set `indexing_threshold`. Although it delivers excellent query speed, building or updating the HNSW graph can be **memory-intensive** if it occurs frequently or across many small segments.
```
timvisee, line 25: It's resource intensive in general. A user can expect 100% CPU consumption especially on smaller clusters. Is that something we want to cover in this article as well, or is it purely focused on memory usage?

qdrant-landing/content/articles/indexing-optimization.md
```diff
@@ -0,0 +1,188 @@
...
+
+### Disabling vs. deferring dense indexing
+
+**`indexing_threshold=0`** 
+ 
+Disables HNSW index creation for dense vectors. Qdrant will not build the HNSW graph for those vectors, letting you upload large volumes of data without incurring the memory cost of index creation.
+
+**`indexing_threshold>0`**  
+
+A positive threshold tells Qdrant how many unindexed dense vectors can accumulate in a segment before building the HNSW graph. Small thresholds (e.g., 100) mean more frequent indexing with less data each time, which can still be costly if many segments exist. Larger thresholds (e.g., 10000) delay indexing to batch more vectors at once, potentially using more RAM at the moment of index build, but fewer builds overall.
```
timvisee, line 40: It's probably good to mention the unit is kilobytes, as it's a bit arbitrary otherwise.

```suggestion
A positive threshold tells Qdrant how many kilobytes unindexed dense vectors can accumulate in a segment before building the HNSW graph. Small thresholds (e.g., 100 KB) mean more frequent indexing with less data each time, which can still be costly if many segments exist. Larger thresholds (e.g., 10000 KB) delay indexing to batch more vectors at once, potentially using more RAM at the moment of index build, but fewer builds overall. 

The following operation can be used to [update](https://qdrant.tech/documentation/concepts/collections/#update-collection-parameters) the indexing threshold in your existing collection:
```

qdrant-landing/content/articles/indexing-optimization.md
```diff
@@ -0,0 +1,188 @@
...
+
+### The `"m"` parameter
+
+For dense vectors, the `m` parameter defines how many edges each node in the HNSW graph can have. Setting `"m": 0` effectively **disables the HNSW graph**, so no dense vector index will be built, no matter the `indexing_threshold`. This can be helpful during massive ingestion if you don’t need immediate searchability.
+
+---
+
+## On-Disk Indexing in Qdrant
+
+By default, Qdrant keeps **vectors and indexes** in **memory** to ensure low-latency queries. However, in large-scale or memory-constrained scenarios, you can configure some or all of those indexes to be stored on-disk. This helps reduce RAM usage at the cost of potential increases in query latency, particularly for cold reads.
```
timvisee, line 63: We can put a few different things on disk:
- vectors itself
- payload data
- payload index
- HNSW index

Reading the bullet list below it seems to only cover payload indices, while the title makes me suggest it also covers HNSW.

Since the title of the article is about memory usage for bulk upload, maybe it's worth to mention all of the above since all of them affect memory usage.

In that case I'd reword 'When to use on-disk indexing' to just 'When to use on-disk'.

qdrant-landing/content/articles/indexing-optimization.md
```diff
@@ -0,0 +1,188 @@
...
+- You have **very large** or **rarely used** payload indexes, and freeing up RAM is worth potential I/O overhead.  
+- Your dataset doesn’t fit comfortably in available memory.  
+- You can tolerate slower queries if it ensures the system remains stable under heavy loads.
+
+---
+
+
+## Memmap storage and segmentation
+
+Qdrant uses **memory-mapped files** (segments) to store data on-disk. Rather than loading all vectors into RAM, Qdrant maps each segment into its address space, paging data in and out on demand. This helps keep the active RAM footprint lower, but each segment still incurs overhead (metadata, page table entries, etc.).
```
timvisee, line 75:
```suggestion
Qdrant uses **memory-mapped files** (segments) to store data on-disk. Rather than loading all vectors into RAM, Qdrant maps each segment into its address space, paging data in and out on demand. This helps keep the active RAM footprint lower, because data can be paged out if memory pressure is high. But each segment still incurs overhead (metadata, page table entries, etc.).
```

qdrant-landing/content/articles/indexing-optimization.md
```diff
@@ -0,0 +1,188 @@
...
+```json
+PATCH /collections/your_collection
+{
+    "vectors": {
+      "on_disk": true
+    }
+}
+```
+
+This approach immediately places all incoming vectors on disk. If you want both dense and sparse vectors to be stored on disk, you need to enable `on_disk` for each type separately.
```
timvisee, line 90:
```suggestion
This approach immediately places all incoming vectors on disk, which can be very efficient in case of bulk ingestion. If you want both dense and sparse vectors to be stored on disk, you need to enable `on_disk` for each type separately.
```

qdrant-landing/content/articles/indexing-optimization.md
```diff
@@ -0,0 +1,188 @@
...
+
+For dense vectors, set `on_disk: true` inside `hnsw_config`.
```
timvisee, line 92: This is not strictly true, and maybe a bit confusing.

For both dense and sparse vectors you can set `on_disk: true` in the vector configuration itself, that will put the vector data on disk.

But, you can also set `on_disk: true` for the HNSW and sparse vector indices. They are separate things.

I'd recommend to adjust this sentence a bit to clarify this difference. In the examples below we can probably show setting both `on_disk` flags.

qdrant-landing/content/articles/indexing-optimization.md
```diff
@@ -0,0 +1,188 @@
...
+
+---
+
+## Best practices for high-volume vector ingestion
+
+
+Bulk ingestion can lead to high memory consumption and even out-of-memory (OOM) errors. **If you’re experiencing out-of-memory errors with your current setup**, scaling up temporarily (increasing available RAM) will provide a buffer while you adjust Qdrant’s configuration for more a efficient data ingestion. 
+
+The key here is to control indexing overhead. Let’s walk through the best practices for high-volume vector ingestion in a constrained-memory environment.
+
```
timvisee, line 126: The most significant change to help with memory is probably to put vector data on disk right away, with the `on_disk: true` flag as discussed above. I'd recommend to put that in this list as first item.

One or two years ago, when we were not able to put vectors on disk yet, vector data itself definitely was the biggest problem. Vector data is huge, and your were forced to keep it in memory before optimization. Optimizers simply weren't fast enough to process all the data and move it to disk.

Directly putting vectors on disk solves this problem. As far as I know, the memory overhead of HNSW indexing isn't that bad, unless you're talking about billion scale.

qdrant-landing/content/articles/indexing-optimization.md
```diff
@@ -0,0 +1,188 @@
...
+**1. Disable HNSW for dense vectors (`m=0`)**
+
+During an **initial bulk load**, you can **disable** dense indexing by setting `"m"=0`. This ensures Qdrant won’t build an HNSW graph for incoming vectors.
+
+Keep `indexing_threshold=10000` (or another large number) so that when you re-enable HNSW, you won’t trigger immediate, frequent index builds. This avoids memory spikes from HNSW building as data arrives, but leaves you the option to enable indexing later by setting `m` to a positive value.
```
timvisee, line 131: I don't think it works this way.

I believe that setting either `indexing_threshold` or `m` back to it's original value will trigger indexing immediately on all your data.

It doesn't do it on all segments at once, but rather in a few jobs one after the other. The larger your CPU is, the more jobs it does in parallel. This is configurable using `max_indexing_threads`, but I'm not sure if it's worth to cover that here.

Ps, I personally prefer setting `indexing_threshold` instead as it's usually a bit easier to understand.

qdrant-landing/content/articles/indexing-optimization.md
```diff
@@ -0,0 +1,188 @@
...
+  "hnsw_config": {
+    "m": 0
+  },
+  "optimizer_config": {
+    "indexing_threshold": 10000
+  }
+}
+```
+
+<aside role="status">If your collection already contains a large number of vectors, changing these parameters will trigger a full index reconstruction, potentially causing downtime or performance degradation.</aside>
```
timvisee, line 145: It should only cause downtime if nodes crash

```suggestion
<aside role="status">If your collection already contains a large number of vectors, changing these parameters will trigger a full index reconstruction, potentially causing slight performance degradation.</aside>
```

qdrant-landing/content/articles/indexing-optimization.md
```diff
@@ -0,0 +1,188 @@
...
+PATCH /collections/your_collection
+{
+  "hnsw_config": {
+    "m": 16
+  }
+}
+```
+
+
+**4. Enable quantization**
```
timvisee, line 167: It's probably best to enable this before enabling indexing again, just to prevent having to do all optimizations a second time.

What would be even better is to enable indexing and quantization in the very same collection update call. But I understand that may be hard to explain in a nice and understandable way here.

qdrant-landing/content/articles/indexing-optimization.md
```diff
@@ -0,0 +1,188 @@
...
+
+If you had planned to store all dense vectors on disk, be aware that searches can slow down drastically due to frequent disk I/O. A more balanced approach is **scalar quantization**: compress vectors (e.g., to `int8`) so they fit in RAM without occupying as much space as full floating-point values.
```
timvisee, line 169: Nitpick: if a user has infinite memory, all data will just be kept in memory in which case I expect it to be fast.

I'll only become slow if a lot of memory pages are constantly swapped in and out because memory pressure is high / memory is limited.

```suggestion
If you had planned to store all dense vectors on disk, be aware that searches can slow down drastically due to frequent disk I/O while memory pressure is high. A more balanced approach is **scalar quantization**: compress vectors (e.g., to `int8`) so they fit in RAM without occupying as much space as full floating-point values.
```

qdrant-landing/content/articles/indexing-optimization.md
```diff
@@ -0,0 +1,188 @@
...
+```
+
+---
+
+## Best practices for high-volume vector ingestion
+
+
+Bulk ingestion can lead to high memory consumption and even out-of-memory (OOM) errors. **If you’re experiencing out-of-memory errors with your current setup**, scaling up temporarily (increasing available RAM) will provide a buffer while you adjust Qdrant’s configuration for more a efficient data ingestion. 
+
+The key here is to control indexing overhead. Let’s walk through the best practices for high-volume vector ingestion in a constrained-memory environment.
```
timvisee, line 125: Not strictly related to memory, but I think it's a good thing to mention and to understand in general. I'm not sure where a sentence like this fits best, and it probably needs a bit of rewording:

> The Qdrant optimizers always try to structure your data in a way that is efficient for searching. If you're bulk uploading a lot of data, Qdrant will constantly be moving all your data around to achieve the goal of efficient search even while your upload is still going on. This adds a lot of unnecessary data movement and overhead. Instead we recommend to upload all your data first, and only let the optimizers do their job after in a single go once all the data is there.

qdrant-landing/content/articles/indexing-optimization.md
sabrinaaquino: Good point. This one focuses more on memory usage, but we could mention that it's resource-intensive overall, since that also ties into memory.

qdrant-landing/content/articles/indexing-optimization.md
sabrinaaquino: Thanks!

qdrant-landing/content/articles/indexing-optimization.md
sabrinaaquino: Nice catch, this can definitely be more clear.

qdrant-landing/content/articles/indexing-optimization.md
sabrinaaquino: Yeah that got confusing, I'll adjust to cover all use cases

qdrant-landing/content/articles/indexing-optimization.md
sabrinaaquino: interesting, I need to look deeper into it later on. thanks for the clarification!

qdrant-landing/content/articles/indexing-optimization.md
sabrinaaquino: done! yeah that's a major item. I thought just leaving it as a separate section would be enough but we should definitely have it on the best practices list as well!

qdrant-landing/content/articles/indexing-optimization.md
sabrinaaquino: I'll add it as a tip!

qdrant-landing/content/articles/indexing-optimization.md
sabrinaaquino: nice observation!

qdrant-landing/content/articles/indexing-optimization.md
sabrinaaquino: I've added this as an additional point!

PR #1510, opened 2025-03-12
Add documentation for shard_number_per_node in collection

adrian0cg:
The configuration that has been added in https://github.com/qdrant/qdrant/pull/4309
but has been missing in the website documentation.


timvisee (changes requested):
Hi @adrian0cg. Thank you for your contribution.

This is already in the configuration file a bit below:

https://github.com/qdrant/landing_page/blob/1c4ba597bc664548313875050b573d19ef52c3e0/qdrant-landing/content/documentation/guides/configuration.md?plain=1#L339

I therefore don't think we should make this change.

PR #1544, opened 2025-04-01
Improve headers and fix redirect for analytics iframe

Rendez:
- The iframe was loading with 301, because it was missing the trailing slash
- Added localhost:3000 to Hugo's HTTP headers for path
- Remove domain checks, since we have CSP headers for protection
- Added development domain to headers, alongside staging and prod.


jasonbryant84 (approved):
LGTM

PR #1547, opened 2025-04-03
[blog] Qdrant and n8n, Part 2

mrscoopers:
Will add pictures from a webinar tmrw, and either generate or re-use the webinar cover slightly changed.


davidmyriel (approved):
proofread and edited content

PR #1548, opened 2025-04-03
Shortcode for rendering code snippets from separate markdown files

trean:


generall:
# Why?

- `.md` files became too huge to work with. 3k+ lines of text and it is not immediately visible where one snippets and and another starts
- Ability to re-use and inspect snippets.
- Ability to convert snippets in a "knowledge base" for potential MCP server.

PR #1560, opened 2025-04-10
docs: DeepEval integration

Anush008:
## [PREVIEW](https://deploy-preview-1560--condescending-goldwasser-91acf0.netlify.app/documentation/frameworks/deepeval/)


kritinv:
Hey @Anush008 , thanks for making the PR. A few things I'd like to mention:

- "RAGAS using local NLP models.": RAGAS is not best supported on DeepEval, so I would avoid mentioning it here. Additionally, DeepEval metrics aren't powered by local NLP models, but any LLM of your choice.
- Would love if you could have a section before Using Qdrant with DeepEval to define the metrics and describe them so users know what these metrics are and why they're necessary.

PR #1562, opened 2025-04-10
[Update] BQ article - Updating recommendations to defer indexing

sabrinaaquino:


sabrinaaquino:
@generall only a couple small changes, should I include anything else?

PR #1599, opened 2025-04-25
docs: Use local inference agentic-rag-crewai-zoom

Anush008:


joein (changes requested):
I made a brief review of the article, and it seems to be confusing
How are we using OpenAI here?
As far as I understand, we wanted to use embeddings from OpenAI, but later replaced it with sentence-transformers/all-MiniLM-L6-v2, however, it seems that we're still trying to use OpenAI in the code https://github.com/qdrant/examples/blob/26df22fc0f6585f32a2572ba24a5b3d1862bf332/agentic_rag_zoom_crewai/vector/data_loader.py#L217

Could we make this tutorial more coherent @davidmyriel ? Or did I get something wrong?

PR #1602, opened 2025-04-25
docs: Local inference data-ingestion-beginners.md

Anush008:


qdrant-landing/content/documentation/data-ingestion-beginners.md
```diff
@@ -52,24 +52,24 @@ The documents used for ingestion can be of various types, such as PDFs, text fil
...
 unstructured
 unstructured[pdf]
 qdrant_client
+fastembed
 ```
 
 ---
 
-**Access Keys:** Store your AWS access key, S3 secret key, and Qdrant API key in a .env file for easy access. You’ll also need an **OpenAI API key**. Here’s a sample `.env` file.
+**Access Keys:** Store your AWS access key, S3 secret key, and Qdrant API key in a .env file for easy access. Here’s a sample `.env` file.
```
joein, line 68: Not sure, if we actually want to replace OpenAI example here, however, I don't mind

qdrant-landing/content/documentation/data-ingestion-beginners.md
```diff
@@ -113,87 +113,35 @@ docs = loader.load()
...
-clip_model = CLIPModel.from_pretrained("openai/clip-vit-base-patch32")
-clip_processor = CLIPProcessor.from_pretrained("openai/clip-vit-base-patch32")
-```
+### Document Processing Function
 
----
+![data-ingestion-beginners-8.png](/documentation/examples/data-ingestion-beginners/data-ingestion-8.png)
 
-The `clip-vit-base-patch32` model is specifically trained to align images and text within the same embedding space, meaning that similar images will have embeddings that are close to each other. With the models set up, you can now create a function to process documents based on their type and generate embeddings using the CLIP model, for instance, to convert image features into vectors.
+Next, we’ll create a `process_document` function that converts various file types—such as text, PDF, and image—into embeddings. This function applies different methods to extract and embed content based on the file type, making it versatile for handling multiple formats effectively.
```
joein, line 125: This description does not match what we do
We're not creating `process_document` anymore, and we're not computing embeddings here

qdrant-landing/content/documentation/data-ingestion-beginners.md
```diff
@@ -235,7 +183,7 @@ In Qdrant, data is organized in collections, each representing a set of embeddin
 
 Here’s how to create a collection in Qdrant to store both text and image embeddings:
 
-```jsx
+```python
```
joein, line 186: if we replace OpenAI with sentence-transformers/all-MiniLM-L6-v2, then we should also update the dim for `text_embedding`

joein (changes requested):
If we replace OpenAI with sentence-transformers, figure 3 and figure 4 images should also be updated, since dim is still 1536 on them

qdrant-landing/content/documentation/data-ingestion-beginners.md
```diff
@@ -235,13 +172,13 @@ In Qdrant, data is organized in collections, each representing a set of embeddin
...
 def create_collection(collection_name):
     qdrant_client.create_collection(
         collection_name,
         vectors_config={
             "text_embedding": models.VectorParams(
-                size=1536,  # Dimension of text embeddings
+                size=384,  # Dimension of text embeddings
```
joein, line 181: Code explanation coming right after the code snippet still refers to text embeddings being 1536 instead of 384

qdrant-landing/content/documentation/data-ingestion-beginners.md
```diff
@@ -298,21 +237,22 @@ if __name__ == "__main__":
...
                 id=idx,  # Unique identifier for each point
                 vector={
-                    "text_embedding": text_embedding[0],                      
-                    "image_embedding": image_embedding[0].tolist(), 
+                    "text_embedding": models.Document(
+                        text=text_review, model="sentence-transformers/all-MiniLM-L6-v2"
+                    ),
+                    "image_embedding": models.Image(
+                        image=product_image, model="Qdrant/clip-ViT-B-32-vision"
+                    ),
```
joein, line 253: We've replaced embeddings with `models.Document` and `models.Image`, however, the explanation of the code below doesn't have any details about these structures, but only about embeddings, maybe we could write a couple of lines that embedding computation happens implicitly with fastembed integration

joein (changes requested):
We still have an image of OpenAI API in the beginning of the article, I would rather replace it with fastembed logo
Also in `Prerequisites` we're still referring to OpenAI models

`data-ingestion-4.png` has 0 points in the collection which might be confusing

qdrant-landing/content/documentation/data-ingestion-beginners.md
```diff
@@ -298,33 +239,34 @@ if __name__ == "__main__":
         text_embedding, image_embedding, points, text_review, product_image = [], [], [], "", ""
```
joein, line 239: we don't need `text_embedding`, `image_embedding anymore` since we're computing them via fastembed

joein (changes requested):
And could we also replace OpenAI logo in the beginning of the article with fastembed?

joein (approved):

PR #1605, opened 2025-04-28
docs: Use local inference in code-search.md

Anush008:
NOTEBOOK UPDATE: https://github.com/qdrant/examples/pull/70


joein (changes requested):
1.
> General usage neural encoder for Natural Language Processing (NLP), in our case
   `all-MiniLM-L6-v2` from the
   [sentence-transformers](https://www.sbert.net/docs/pretrained_models.html) library

I'd rather not mention the library itself, but just write `sentence-transformers/all-MiniLM-L6-v2`, or, e.g. write `all-MiniLM-L6-v2` from sentence-transformers (referring to it as to organization).

2. We also need to update the attached colab notebook
3. In `Ingestion Pipeline` we already can mention, that we're gonna use `fastembed`
4. Replace inferring embedding dimensionalities on remove variables with `client.get_embedding_size(model_name)` for both models.

```python
client.create_collection(
    "qdrant-sources",
    vectors_config={
        "text": models.VectorParams(
            size=nlp_embeddings.shape[1],
            distance=models.Distance.COSINE,
        ),
        "code": models.VectorParams(
            size=code_embeddings.shape[1],
            distance=models.Distance.COSINE,
        ),
    }
)
```
5. Would also be nice to format `models.Document(text=query, model="sentence-transformers/all-MiniLM-L6-v2")` and the other document in `query_batch_points`, so scroll bar would not appear on the page

Other than this - lgtm 👍

Anush008:
@joein Notebook PR: https://github.com/qdrant/examples/pull/70

joein:
We would also need to update colab
We could update code-search example later if we decide so

joein (approved):
lgtm, the only remaining parts to update are colab notebooks and code search demo

PR #1635, opened 2025-05-10
Typo in explore.md Context Search

fcossio:
Match the behavior of the python client. Score 0.0 is the best possible score. Points are sorted in ascending order.


fcossio:
Scores are negative...

PR #1649, opened 2025-05-16
Resurface-hidden-case-studies

daniel-azoulai:


daniel-azoulai:
@sabrinaaquino I noticed that case studies were featured on the blog post, but then they were removed after more featured blog posts were added. I thought they would just go to the general list below the feature section, but they didn't. So I changed the metadata block for "Featured" from "true" to "false". In the [preview](https://deploy-preview-1649--condescending-goldwasser-91acf0.netlify.app/blog/), they're in the right spot. Example below
image

sabrinaaquino (approved):

sabrinaaquino:
@daniel-azoulai interesting, this may have happened to other blogs as well. thanks for the observation.

PR #1689, opened 2025-06-08
wrong explanation for "must_not" filter.

arshamalh:


kacperlukawski (approved):
LGMT, thanks @arshamalh

PR #1711, opened 2025-06-17
Qdrant DSPy medical bot

mwitiderrick:


mwitiderrick:
> It doesn't belong to the articles. It is more of the `build` section.
> 
> Also, please improve images

@generall how can the images be improved? And which images specifically

PR #1731, opened 2025-06-24
Tutorial: using multivectors - set the batch size for points upload

kacperlukawski:
As requested on Slack, I'm changing the tutorial to suggest batching. We don't have much data, but it's generally a good practice.


mwitiderrick (approved):
lgtm

PR #1742, opened 2025-06-30
Remove chatgpt plugin

mwitiderrick:


generall (changes requested):
We can't simply remove content from pages, as it is still indexed and searched by google 

image


Dead links will compromise our search relevance score.

Instead we need to de-list it and remove from sitemap.

generall:
Alternatively, we can come up with the page of where we want to redirect this one to.

PR #1747, opened 2025-07-02
update neural search  - Merge article with tutorial

mwitiderrick:


qdrant-landing/static/articles_data/neural-search/encoders.png
kacperlukawski, line 1: I think this file should not be under `static/articles_data` as it no longer belongs to an article. It's part of the documentation, so that's where it should be put.

kacperlukawski (changes requested):
Please move the image to a different directory. Other than that, LGTM.

PR #1761, opened 2025-07-04
Update reranking-hybrid-search Images

mwitiderrick:


mwitiderrick:
> PREVIEW



> Hey, I'd appreciate having another look at the images and centering all the texts. If you check how the elements positioning look like for "Dense Embeddings" and "Sparse Embeddings", you'll see they're not consistent. Also, the images don't need to bee that big.

Please check now

PR #1780, opened 2025-07-14
Update optimization image

mwitiderrick:
[PREVIEW](https://deploy-preview-1780--condescending-goldwasser-91acf0.netlify.app/documentation/concepts/optimizer/)


qdrant-landing/static/docs/optimization.png
kacperlukawski, line 1: I don't think it's a 1:1 remake of the previous image. The old one clearly shows that segments are parts of the collection, but here it's just connected through register.

image

kacperlukawski (changes requested):

PR #1797, opened 2025-07-16
time-based boosting example

mwitiderrick:


generall (changes requested):
We don't need another "bould" tutorial for that. It would be enough to have a section next to geo-based boosting in the main docs.

PR #1802, opened 2025-07-17
Update quickstart.md

Nothing-Works:
There is a problem for windows docker user, where if you mount to a local folder then the next time to start the container it will not work. Using a named mount solves the problem so I thought it's better to notify in the docs. related link https://github.com/qdrant/qdrant/issues/6679#issuecomment-2983254015


timvisee (approved):

Nothing-Works:
@timvisee who is going to merge this PR?

generall:
But I agree that it is better to be mentioned

PR #1840, opened 2025-08-07
docs: Clean up integration docs

Anush008:
## Description

- Removed some integration pages as discussed.
- Updated some pages with the latest info.


kacperlukawski (approved):
LGTM, but honestly, I expected more projects to be discontinued

kacperlukawski (changes requested):
On second thought, it would lead to lots of 404s as URLs won't work anymore. Maybe we can just create aliases pointing to the integrations section, so people can at least find the alternatives.

Anush008:
I had that thought too. But assumed a 404 would be better than redirecting to an unrelated page.

Will redirect to the the index page.

kacperlukawski (approved):

PR #1849, opened 2025-08-12
Extend Private Cloud Setup guide

bashofmann:
* Example for upgrade 
* Example for changing the configuration 
* Example for StorageClass and VolumeSnapshotClass modification 
* Example for image mirroring


leflambeur (changes requested):
A number of changes, mostly trying to be concise or less wordy or more declarative

Some of the comments in the configuration use a lot of filler but this is consistent throughout so it might be idiomatic - feel to disregard if it is idiomatic but it doesn't read well in situ for me

PR #1857, opened 2025-08-14
Automate llms.txt generation

kacperlukawski:
This PR contains a script that automatically generates llms.txt and llms-full.txt, so whenever we change anything in the content, it's automatically reflected in these files.

### Key characteristics

- GitHub Models are used to summarize the files for the llms.txt
- The summarization is not applied to the URLs that already exist in LLMS.txt
- The llms-full.txt is always regenerated

The current state of both files was generated with this script. I also added a GitHub action that should automatically update the state of them.


automation/generate-llms-txt.py
```diff
@@ -0,0 +1,208 @@
...
+    :param content:
+    :return:
+    """
+    # Truncate the content to a maximum of 8192 characters
+    content = content[:8192]
+
+    # Call the GitHub Models API to generate a summary
+    client = openai.OpenAI(
+        api_key=os.environ.get("GITHUB_TOKEN"),
+        base_url="https://models.github.ai/inference",
```
generall, line 143: Do you propose to call language model as a part of CI process?

PR #1864, opened 2025-08-18
troubleshooting windows docker mount

generall:
Preview https://deploy-preview-1864--condescending-goldwasser-91acf0.netlify.app/documentation/guides/common-errors/#incompatible-file-system

### DIFF
--- a/qdrant-landing/content/documentation/guides/common-errors.md
+++ b/qdrant-landing/content/documentation/guides/common-errors.md
@@ -32,6 +32,63 @@ ulimit -n 10000
 
 Please note, the command should be executed before you run Qdrant server.
 
+## Incompatible file system
+
+Qdrant have a [set of requirements](https://qdrant.tech/documentation/guides/installation/#storage) for persistent file storag.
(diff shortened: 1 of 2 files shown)


qdrant-landing/content/documentation/guides/common-errors.md
```diff
@@ -32,6 +32,69 @@ ulimit -n 10000
...
+"called `Result::unwrap()` on an `Err` value: OutputTooSmall { expected: 4, actual: 0 }"
+```
+
+It might be also possible that vector data will be lost (set to all zeros) after service restart.
+
+
+### How to avoid Incompatible file system?
+
+Most common used configuration of incompatible file system is usage of WSL-baced Docker containers in Windows.
+When you mount Windows folder into Qdrant docker container, the Windows hyper visor creates a shared mount, which is not fully POSIX-compatible.
```
timvisee, line 81: AFAIK this isn't FUSE, because it's mounted by the kernel directly? FUSE is for user space mounts, not for special mounts like this.

Or do you see the 'FUSE' error message in that specific scenario?

In any case, I changed the text. If you don't agree we can change it back 👍

timvisee (approved):

timvisee:
Also added this banner in the installation/storage section:

image

PR #1865, opened 2025-08-18
Add snippets to disable quantization in guide

KShivendu:
It was missing from the [quantization guide](https://qdrant.tech/documentation/guides/quantization) and the API design is non-obvious. So better to document it clearly.

```http
PATCH /collections/{collection_name}
{
    "quantization_config": "Disabled"
}
```


timvisee:
> ```httpspec
> "quantization_config": "Disabled"
> ```

IIRC we use the lowercase `"disabled"` in other places

KShivendu:
> in other places

Where exactly? Here I've clearly verified that Qdrant needs "Disabled" instead of "disabled"

Anush008 (approved):

PR #1882, opened 2025-08-29
Add article about MUVERA embeddings

kacperlukawski:
This article describes the process of creating MUVERA embeddings out of multi-vector representations. The most recent version of FastEmbed brings the support of them, but we also want to cover the algorithm itself.


generall:
Hey! Could you please share a link to image sources in figma?

kanungle:
> Hey! Could you please share a link to image sources in figma?

@generall we are attempting to try excalidraw with Qdrant style defaults, as discussed. They excalidraw json is included with each article.

kacperlukawski:
> @kacperlukawski i reviewed. It generally looks like it's on a good path but I think it needs a little more.
> 
> What are the RPS comparisons with single vector representations? Can we quantify? When should I use MUVERA? Do we have practical examples? What are the closing thoughts? It feels to end abruptly.
> 
> Also might be use full to add a simple diagram of variable length vs fixed length vectors

I like the idea of benchmarking it. I need to run the experiments.

However, I didn't fully get the diagram of variable-length vs fixed-length vectors. I only mention "variable-length sequences of vectors" in the article, as multi-vector representations usually have more vectors for documents than for the queries. I don't know how to visualize that, so hopefully the first diagram shows that more or less.

Btw, I updated the diagrams.

kacperlukawski:
@kanungle I made some changes to the article and added some metrics to show an impact on the quality and latency. Would appreciate you having another look.

kanungle:
@kacperlukawski I like the additions a lot. Very helpful.

I do think that we should show bar charts when possible to complement tables, for easier interpretability. Okay this time but something for the future.

I think the graphics still need a little bit of work due to small fonts and line styles. I will try to edit your excalidraw and see if I can improve it.

I also believe every section needs a header, including the intro and conclusion. You could start with "What are MUVERA Embeddings" and close with header "Try it out today"

Lastly, I think we need a more descriptive title for people that don't know what MUVERA is. Maybe "MUVERA: Making Multivector Embeddings More Feasible/Practical/Accessible/Performant"

kanungle:
Hey @kacperlukawski this looks really good!

One minor change: can you make the article title in 'Title Case' not 'Sentence case'

Meaning just capitalize the first character of words greater than 4 characters

After that I will approve

kanungle:
Great work, thanks!

PR #1941, opened 2025-10-06
separate branch for content updates that are not related to the cours…

holyMolyTolli:
…e but come from the enable-course branch

This only contains changes that replace "localhost" with a managed cloud URL and key. All changes were/are part of the enable-course branch.

Created this branch to divide the enable-course branch into smaller parts. However, I'm not sure if we should merge this. I'll let @kacperlukawski decide.


holyMolyTolli:
created [new PR](https://github.com/qdrant/landing_page/pull/1944) because I renamed the branch to match it's actual content.

PR #1962, opened 2025-10-22
Fill Course home page

kanungle:
initial content for /courses page. To be refined.


kanungle:
@generall could you change the button text to "Register Interest" for all cards except the first one?

PR #1978, opened 2025-11-04
[doc] Add limit parameter to query configuration

BrodaNoel:
It took me forever to understand why the hell it was only retrieving 10 results if I was properly set `limit:20` inside each prefetch.

Then I saw the HTTP example: https://github.com/qdrant/landing_page/blob/master/qdrant-landing/content/documentation/headless/snippets/query-points/hybrid-basic/http.md

So, I'm fixing for the future fellas.

PS.: I ❤️ qdrant

I'm using it on [ArchivosYa.com - AI for lawyers, in argentina](https://archivosya.com) (soon to go abroad!)


abdonpijpelink (approved):
LGTM. Thanks for making this more clear @BrodaNoel !

I've taken the liberty of also updating the other snippets, so it's consistent across all languages.

PR #1996, opened 2025-11-17
[v1.16] Update configuration snippet

timvisee:
Update configuration snippet to the latest version.

Sourced from: <https://github.com/qdrant/qdrant/blob/v1.16.0/config/config.yaml>


qdrant-landing/content/documentation/discord.md
coszio: Where did this come from?

PR #2010, opened 2025-11-21
Remove init_from from docs

abdonpijpelink:
`init_from` was removed in version 1.16 (https://github.com/qdrant/qdrant/pull/7454). This PR removes it from the documentation.


qdrant-landing/content/documentation/concepts/collections.md
```diff
@@ -59,22 +59,6 @@ will enable the use of
 [memmaps](/documentation/concepts/storage/#configuring-memmap-storage),
 which is suitable for ingesting a large amount of data.
 
-### Create collection from another collection
```
timvisee: Maybe we can hint users to use our [migration](https://github.com/qdrant/migration/) tool instead. I believe it can be used for this purpose.

PR #2020, opened 2025-12-03
Docs: remove payload filter from JWT

timvisee:
This feature has been removed in [Qdrant 1.16.0](https://github.com/qdrant/qdrant/releases/tag/v1.16.0)


timvisee:
> I see there's also a couple of mentions of payload filtering in this article: https://qdrant.tech/articles/data-privacy ([here](https://qdrant.tech/articles/data-privacy/#payload-configuration) and [here](https://qdrant.tech/articles/data-privacy/#creating-role-based-access-control-rbac-using-jwt)), but we can also change that article in a separate PR, because articles are not strictly docs.

Great observation.

Would you be okay with taking care of that? I'd be totally fine with including it in this PR.

abdonpijpelink:
I've updated the article.

PR #2022, opened 2025-12-04
Docs: text search guide

abdonpijpelink:
## [Preview](https://deploy-preview-2022--condescending-goldwasser-91acf0.netlify.app/documentation/guides/text-search/)

This PR adds a new guide that provides an overview of Qdrant's text search capabilities: semantic search, filtering, full-text search, and hybrid search.

### DIFF
--- a/qdrant-landing/content/documentation/guides/text-search.md
+++ b/qdrant-landing/content/documentation/guides/text-search.md
@@ -268,7 +268,7 @@ For a tutorial on using SPLADE++ with FastEmbed, refer to [How to Generate Spars
 
 ### miniCOIL
 
-[miniCOIL](/articles/minicoil/) strikes a balance between the flexibility of BM25 and the performance of SPLADE++. Like SPLADE++, miniCOIL is a transformer-based model that generates sparse vectors for text. Instead of using word expansion to capture the context and meaning of terms, the model generates a four-dimensional vector for each term. miniCOIL does not use a fixed vocabulary, making it an effective model for lexical search that ranks results based on the contextual meaning of keywords.
+[miniCOIL](/articles/minicoil/) strikes a balance between the flexibility of BM25 and the performance of SPLADE++. miniCOIL is a transformer-based model that generates sparse vectors for text. Unlike SPLADE++, it doesn't use a vocabulary expansion mechanism. To capture the context and meaning of terms, the model generates a four-dimensional vector for each term. miniCOIL does not use a fixed vocabulary, making it an effective model for lexical search that ranks results based on the contextual meaning of keywords.
 
 miniCOIL can be [used with the FastEmbed library](/documentation/fastembed/fastembed-minicoil/).
 
@@ -284,7 +284,7 @@ After ingesting data with both vectors, you can use the prefetch feature to run 
 
 {{< code-snippet path="/documentation/headless/snippets/text-search/hybrid-prefetch-rrf/" >}}
 
-This query searches for an ISBN, for which only the lexical search returns a result. The `score_threshold` for the semantic query prevents low-scoring results to be returned. So in this case, only the lexical result is returned to the user. If a user had searched for "time travel", only the semantic search would return results, and those would be returned to the user. If a user would search for a term that matched both the semantic and lexical vectors, the results from both searches would be combined to provide a more comprehensive set of results. 
+This query searches for an ISBN, for which only the lexical search returns a result. The `score_threshold` for the semantic query prevents low-scoring results to be returned (0.5 is just an example threshold; you need to tune what a good threshold is for your data and model). So in this case, only the lexical result is returned to the user. If a user had searched for "time travel", only the semantic search would return results, and those would be returned to the user. If a user would search for a term that matched both the semantic and lexical vectors, the results from both searches would be combined to provide a more comprehensive set of results. 
 
 You are not limited to prefetching just two queries. Examples include, but are not limited to:
 


qdrant-landing/content/documentation/guides/text-search.md
```diff
@@ -0,0 +1,533 @@
...
+  "limit": 10,
+  "with_payload": true
+}
+```
+
+### SPLADE++
+
+The SPLADE (Sparse Lexical and Dense) family of models are transformer-based models that generate sparse vectors out of text. These models combine the benefits of traditional lexical search with the power of transformer-based models by generating homonyms and synonyms.
+
+The advantage of using SPLADE models is that they [perform better](/articles/sparse-vectors/#splade) than traditional BM25. The downside is that they use a fixed vocabulary, which means that you can't use them to find terms that are not in the vocabulary, such as product IDs.
```
mrscoopers: <and they're far heavier on inference than BM25>
<and they don't work out of domain/language like BM25>

qdrant-landing/content/documentation/guides/text-search.md
```diff
@@ -0,0 +1,533 @@
...
+        "match": {
+          "value": "Jerry Pournelle"
+        }
+      }
+    ]
+  }
+}
+```
+
+Note that both filter clauses must be true for a point to be included in the results, because a `must` clause operates like a logical `AND`. If you want to find books written by either author (as well as both), use a `should` clause, which operates like a logical `OR`:
```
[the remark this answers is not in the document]
abdonpijpelink: I wanted to give an overview of all the ways you can search for text with Qdrant, including all the filtering and ranking options.

qdrant-landing/content/documentation/guides/text-search.md
abdonpijpelink: [Added](https://github.com/qdrant/landing_page/pull/2022/commits/e1fb21cc1b9bf4d8e77d5fb30622d2c3c508f427). Thanks!

qdrant-landing/content/documentation/guides/text-search.md
```diff
@@ -0,0 +1,533 @@
...
+        "match": {
+          "text_any": "space war"
+        }
+      }
+    ]
+  }
+}
+```
+
+To filter on phrases, use a `phrase` condition. This requires enabling [phrase searching](/documentation/concepts/indexing/#phrase-search) when creating the full-text index.
```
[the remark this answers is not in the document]
abdonpijpelink: [Done](https://github.com/qdrant/landing_page/pull/2022/commits/bceed406d936cba2f0ea4a0271f886a5b5ebbabf)

qdrant-landing/content/documentation/guides/text-search.md
```diff
@@ -0,0 +1,533 @@
...
+  },
+  "using": "title-bm25",
+  "limit": 10,
+  "with_payload": true
+}
+```
+
+### SPLADE++
+
+The SPLADE (Sparse Lexical and Dense) family of models are transformer-based models that generate sparse vectors out of text. These models combine the benefits of traditional lexical search with the power of transformer-based models by generating homonyms and synonyms.
```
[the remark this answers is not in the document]
abdonpijpelink: I have [expanded the SPLADE section](https://github.com/qdrant/landing_page/pull/2022/commits/81551cdd3eab9fcccb331a385ea6b7b7af6dfc3c).

qdrant-landing/content/documentation/guides/text-search.md
```diff
@@ -0,0 +1,533 @@
...
+    "fusion": "rrf"
+  },
+  "limit": 10,
+  "with_payload": true
+}
+```
+
+This specific query searches for an ISBN, for which only the lexical search returns a result, which is then returned to the user. If a user had searched for "time travel", only the semantic search would return results, and those would be returned to the user. If a user would search for a term that matched both the semantic and lexical vectors, the results from both searches would be combined to provide a more comprehensive set of results. 
+
+You're not limited to prefetching two queries. For example, you could run multiple lexical queries across the `title`, `author`, and `isbn` fields alongside a semantic query to achieve a comprehensive search across all data.
```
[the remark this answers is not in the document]
abdonpijpelink: I have [included](https://github.com/qdrant/landing_page/pull/2022/commits/456a683493a2d224118fd5f9d1b1112802999ad4) your examples.

qdrant-landing/content/documentation/guides/text-search.md
```diff
@@ -0,0 +1,760 @@
...
+- By default, each word is then [converted to lowercase](/documentation/concepts/indexing/#lowercasing). Lowercasing the tokens allows Qdrant to ignore capitalization, making full-text filters case-insensitive.
+- Optionally, Qdrant can remove diacritics (accents) from characters using a process called [ASCII folding](/documentation/concepts/indexing/#ascii-folding). This ensures that diacritics are ignored. As a result, filtering for the word "cafe" matches "café".
+- Optionally, tokens can be reduced to their root form using a [stemmer](/documentation/concepts/indexing/#stemmer). This ensures that filtering for "running" also matches "run" and "ran". Because stemming is language-specific, if enabled, it must be configured for a specific language.
+- Certain words like "the", "is", and "and" are very common in text and do not contribute much to the meaning of text. These words are called [stopwords](/documentation/concepts/indexing/#stopwords) and can optionally be removed during indexing.
```
mrscoopers: Non-customized stopwords lists are also language-specific afaik

qdrant-landing/content/documentation/guides/text-search.md
```diff
@@ -0,0 +1,760 @@
...
+      "ascii_folding": true
+    }
+  },
+  "using": "author-bm25",
+  "limit": 10,
+  "with_payload": true
+}
+```
+
+#### Configuring BM25 Parameters
```
mrscoopers: Perhaps it should be before the language-specific part? I feel like `avg_len` is pretty important & not intuitive at all

qdrant-landing/content/documentation/guides/text-search.md
```diff
@@ -0,0 +1,285 @@
...
+
+{{< code-snippet path="/documentation/headless/snippets/text-search/create-hybrid-collection/" >}}
+
+After ingesting data with both vectors, you can use the prefetch feature to run both semantic and lexical queries in a single request. The results of both queries are then combined using a fusion method like Reciprocal Rank Fusion (RRF).
+
+{{< code-snippet path="/documentation/headless/snippets/text-search/hybrid-prefetch-rrf/" >}}
+
+This query searches for an ISBN, for which only the lexical search returns a result, which is then returned to the user. If a user had searched for "time travel", only the semantic search would return results, and those would be returned to the user. If a user would search for a term that matched both the semantic and lexical vectors, the results from both searches would be combined to provide a more comprehensive set of results.
```
mrscoopers: > This query searches for an ISBN, for which only the lexical search returns a result, which is then returned to the user. If a user had searched for "time travel", only the semantic search would return results, and those would be returned to the user.

I suspect that in reality, the dense model will pick up some irrelevant information for ISBN (or is the threshold against this?). Then we need to be accurate, many people think absolute values of similarity are universal, when actually everything is very domain (collection/retriever)-dependent, and what matters is relative similarity.

qdrant-landing/content/documentation/guides/text-search.md
```diff
@@ -0,0 +1,760 @@
...
+      "language": "spanish"
+    }
+  },
+  "using": "title-bm25",
+  "limit": 10,
+  "with_payload": true
+}
+```
+
+To configure only a stemmer or a stopword set, rather than both, set `language` to `none` and specify the configuration for the desired stemmer or stopwords.
```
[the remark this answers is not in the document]
abdonpijpelink: The default word stemmer should work fine for Spanish (or any language that uses spaces to separate words).

qdrant-landing/content/documentation/guides/text-search.md
abdonpijpelink: Sure. I've moved it before the Language-specific Settings section.

qdrant-landing/content/documentation/guides/text-search.md
```diff
@@ -0,0 +1,760 @@
...
+        "match": {
+          "text_any": "space war"
+        }
+      }
+    ]
+  }
+}
+```
+
+Qdrant also supports phrase filtering, enabling you to search for multiple words in the exact order they appear in the original text, with no other words in between. For example, a phrase filter for "time machine" matches against the title "The Time Machine" but would not match "The Time Travel Machine" (there's a word between "time" and "machine") nor "Machine Time" (the word order is incorrect).
```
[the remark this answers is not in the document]
abdonpijpelink: I've added a table that should make it more clear. What do you think?

image

qdrant-landing/content/documentation/guides/text-search.md
mrscoopers: > So in this case

So in most of the cases? Otherwise there's a risk making people believe this threshold is universal & will always work:)

PR #2052, opened 2025-12-19
Update fonts

trean:
Replacing fonts: Satoshi -> Mona Sans, Source Code Pro -> Geist Mono + small typography tweaks


generall (changes requested):
I don't like how new font looks in articles

generall (approved):

PR #2059, opened 2025-12-22
Tutorials Restructuring

kanungle:
# [PREVIEW](https://deploy-preview-2059--condescending-goldwasser-91acf0.netlify.app/documentation/tutorials-lp-overview/)

PR #2102, opened 2026-01-27
docs: add new tutorial for Golang E-Commerce

nleroy917:
This adds a new tutorial for building a search engine in Golang (no python!). It uses Qdrant cloud inference. I'm using OpenAI `text-embedding-3`. This costs money, so I'm willing to swap it. its a more intermediate tutorial, so it might be fine. Open to opinions on it.

Its intended to be read in tandem with this example code: https://github.com/nleroy917/ecommerce-search-golang

### Wait for:
- [x] [Example repository](https://github.com/qdrant/examples/pull/93#pullrequestreview-3718260123) to get merged
- [x] Tutorials restructure to get merged (#2059), sync with `upstream/master`


nleroy917:
> Why not use free cohere? I think nomic might have free too

I agree. I started with OpenAI to get myself off the ground, but want to switch to a free one

abdonpijpelink:
This looks great @nleroy917 !

> free cohere? 

Cohere trial keys are limited to 1000 requests per month, so it won't work for ingesting 100K points, but you could pre-embed all the data and only use a Cohere trial key for the queries.

PR #2219, opened 2026-03-20
"Things to Check Before Taking Qdrant into Production" guide

abdonpijpelink:
# [Preview](https://deploy-preview-2219--condescending-goldwasser-91acf0.netlify.app/documentation/operations/production-checklist/)

Adds a "Things to Check Before Taking Qdrant into Production" guide to the Operations docs.

The goal is not to add any new information, but rather to help the discovery of existing information by collecting it in a single concise guide that users can use as a checklist.


qdrant-landing/content/documentation/operations/production-checklist.md
```diff
@@ -0,0 +1,118 @@
...
+Distribute incoming requests evenly across cluster nodes to ensure consistent performance. Without load balancing, a single overloaded node can cause timeouts across your entire application. Qdrant Cloud includes a load balancer, but self-managed deployments need to configure this separately.
+
+---
+
+## 3. Quantization
+
+Compress vectors to reduce memory footprint. [Quantization](/documentation/manage-data/quantization/) is one of the most impactful changes you can make before going to production.
+
+- **Evaluate whether [Scalar Quantization](/documentation/manage-data/quantization/#scalar-quantization) fits your use case.**
+Scalar quantization converts `float32` to `uint8`, reducing memory by a factor of 4. Accuracy loss is typically below 1%. The right default for most production workloads, especially with high-dimensional vectors.
```
brian-ogrady: Recommend evaluating based on specific dataset and embedding model. Not all embedding models perform well with quantization, even scalar.

qdrant-landing/content/documentation/operations/production-checklist.md
```diff
@@ -0,0 +1,118 @@
...
+
+- **Consider [Binary Quantization](/documentation/manage-data/quantization/#binary-quantization) for maximum compression.**
+Binary quantization reduces memory by a factor of 32 and can significantly speed up searches. Best suited for compatible high-dimensional embedding models (for example, OpenAI `text-embedding-ada-002` or Cohere `embed-english-v2.0`).
+
+- **Benchmark retrieval quality after applying quantization.**
+Some models produce embeddings that can't be quantized efficiently. Verify that error rates stay within your acceptable threshold for your specific dataset and query patterns. Rescoring adds latency, so ensure it meets your performance targets.
```
brian-ogrady: Rescoring adds latency because disk reads are slower than RAM, and HNSW requires random reads, ability to rescore and depth of rescoring depends on the `limit` and `ef_search` search at query time, in addition to number of segments in the cluster and `indexed_only`. These parameters, and others (such as whether or not you decide to `prefetch` the scalar quantized vectors w/o rescoring and then rescore the full-resolution vectors at the shard level rather than the segment level) will affect the number of IOPS required per query, which may lead to cache thrashing in memory-constrained environments, even with quantization.

qdrant-landing/content/documentation/production-checklist.md
```diff
@@ -0,0 +1,118 @@
...
+- **Enable inline storage.**
+When storing vectors and the HNSW index on disk, improve search performance by [enabling inline storage](/documentation/operations/optimize/#inline-storage-in-hnsw-index). It makes searches faster by reducing the number of I/O operations, at the cost of increased storage usage.
+
+---
+
+## 5. Query Optimization
+
+Ensure your search is fast, accurate, and efficient under production load.
+
+- **Create [payload indexes](/documentation/manage-data/indexing/#payload-index) on fields used for filtering.**
```
brian-ogrady, line 70: We need to be clear with people that Qdrant will not leverage payload fields for filterable HNSW if they create the payload fields after creating the HNSW graph. So if they want to leverage a payload field for filterable HNSW BUT they already ingested their dataset, they must index the payload field and then reindex HNSW separately.

qdrant-landing/content/documentation/operations/production-checklist.md
```diff
@@ -0,0 +1,118 @@
...
+Ensure your search is fast, accurate, and efficient under production load.
+
+- **Create [payload indexes](/documentation/manage-data/indexing/#payload-index) on fields used for filtering.**
+Payload indexes speed up filtering and reduce load on the system. Identify which fields are commonly used in filters and create indexes on them before ingesting data.
+
+- **Apply payload filters to narrow the search space.**
+Searching every data point is inefficient at scale. [Filtering](/documentation/search/filtering/) on specific payload fields can reduce computational load and focus queries on relevant data subsets.
+
+- **Evaluate whether [hybrid search](/documentation/search/hybrid-queries/) fits your use case.**
+Combine dense vector search (semantic similarity) with sparse vector search (keyword matching) for higher precision.
```
brian-ogrady: Do we have benchmarks showing that this actually improves accuracy?

qdrant-landing/content/documentation/operations/production-checklist.md
```diff
@@ -0,0 +1,118 @@
...
+
+- **Rerank for maximum result precision.**
+After initial hybrid retrieval, apply late interaction embeddings to [rerank](/documentation/tutorials-search-engineering/reranking-hybrid-search/) results. Reranking can be computationally expensive, so aim for a balance between relevance and speed.
```
brian-ogrady: Ensure `m = 0` specifically for the late interaction embeddings to avoid large HNSW graph overhead. Also, ensure you are capacity planning correctly and are making the correct tradeoffs between RAM and disk access.

qdrant-landing/content/documentation/operations/production-checklist.md
```diff
@@ -0,0 +1,118 @@
...
+
+- **Implement batch processing for inserts and queries.**
+[Group vector inserts into larger batches](/documentation/manage-data/points/?q=batch#upload-points) rather than individual transactions to reduce write overhead. [Batch multiple queries together](/documentation/search/search/#batch-search-api) to cut round trips to the database.
+
+- **Reduce tail latency with [delayed fan-outs](/documentation/search/low-latency-search/#use-delayed-fan-outs).**
+Use the `read_fan_out_delay_ms` parameter to automatically query a second replica if the first one doesn't respond within the required latency threshold.
```
brian-ogrady: Can only be used when `replication_factor > 1` should clarify here.

qdrant-landing/content/documentation/operations/production-checklist.md
```diff
@@ -0,0 +1,118 @@
...
+
+- **[Query indexed data only](/documentation/search/low-latency-search/#query-indexed-data-only).**
+Under heavy write loads, large amounts of data may not be indexed immediately, leading to slower searches. Use `indexed_only` to ensure queries only consider indexed data and `prevent_unoptimized` to prevent the creation of large unoptimized segments.
```
brian-ogrady: I am hearing we may be deprecating `index_only`. We should recommend it for 1.17.0 and prior versions, unclear if we should still be recommending it 1.17.1 +

qdrant-landing/content/documentation/operations/production-checklist.md
abdonpijpelink: I think this is covered by the third bullet?

qdrant-landing/content/documentation/operations/production-checklist.md
abdonpijpelink: A bit too much detail for an overview page like this perhaps? I've added a link to "[Memory and speed tuning](https://qdrant.tech/documentation/manage-data/quantization/#memory-and-speed-tuning)" in the quantization docs instead, and made a note to include your guidance in the upcoming refresh of the quantization docs.

qdrant-landing/content/documentation/production-checklist.md
abdonpijpelink: Added. Also added an extra notice to the [filterable hnsw index section of the docs](https://qdrant.tech/documentation/manage-data/indexing/#filterable-hnsw-index) for visibility.

qdrant-landing/content/documentation/operations/production-checklist.md
abdonpijpelink: It'll depend on the data/queries, so I've reworded as: _"Hybrid search casts a wide retrieval net, maximizing recall by using multiple retrieval methods, such as combining dense vector search (semantic similarity) with sparse vector search (keyword matching). Evaluate its effectiveness for your specific dataset and queries."_

qdrant-landing/content/documentation/operations/production-checklist.md
abdonpijpelink: I've added this advice to the hybrid query docs, where we talk about rescoring.

qdrant-landing/content/documentation/operations/production-checklist.md
abdonpijpelink: Instead of mentioning specific settings here, I've updated this paragraph so it only links to the "search indexed data only" section of the docs, which we can updated when `prevent_unoptimized` is no longer experimental.

brian-ogrady (approved):

PR #2269, opened 2026-04-09
Update advanced-search page

nastyapash:


trean (approved):

trean:
`force push` after rebasing to the `master`. Restored the previous name of one card, since we don't yet have a suitable link for it.

PR #2284, opened 2026-04-17
Tweak prevent unoptimized docs

abdonpijpelink:
Two changes to the docs for `indexed_only` and `prevent_unoptimized`:

- Use stronger language not to use `prevent_unoptimized` with `wait=true`
- Position `indexed_only` and `prevent_unoptimized` as alternative mechanisms: there's no need to use both.


qdrant-landing/content/documentation/operations/optimizer.md
```diff
@@ -133,31 +133,30 @@ To address this, Qdrant supports [querying indexed data only](/documentation/sea
 
 To mitigate this, the optimizer supports a `prevent_unoptimized` mode. When enabled, points written to an unindexed segment that is larger than `indexing_threshold` are accepted and durably stored but are not visible in search results until the optimizer has indexed the segment. These are called deferred points. Not until the optimizer finishes indexing a segment containing deferred points, do those points become visible.
 
-Set `prevent_unoptimized` to `true` when creating or updating a collection:
+Enabling `prevent_unoptimized` only affects newly created segments. Existing segments are not changed retroactively. Similarly, changing `indexing_threshold` does not affect existing segments. Only new segments will use the updated threshold.
```
timvisee: I find these sentences rather confusing. We may be better of removing it all together.

> prevent_unoptimized only affects newly created segments

Not true.

It is mostly true in that the setting enables the deferred points mechanism. That is only enabled in new segments. The caveat is that restarting the Qdrant node enables it in **all** segments (including existing ones).

> Similarly, changing `indexing_threshold` does not affect existing segments. Only new segments will use the updated threshold.

Not true. Existing segments are also updated.

qdrant-landing/content/documentation/operations/optimizer.md
```diff
@@ -133,31 +133,30 @@ To address this, Qdrant supports [querying indexed data only](/documentation/sea
...
-
 ### Effect on `wait=true`
 
+Write requests support a [`wait` parameter](/documentation/manage-data/points/#awaiting-result) that, when set to `true`, causes the request to return only after the update has been applied and is visible for search. With `prevent_unoptimized` enabled, setting `wait=true` is not recommended without understanding the implications.
+
 Qdrant processes updates in strict order: each update is written to the write-ahead log and then applied sequentially by the update worker, preserving this order.
 
 Under normal conditions, setting `wait=true` on a write request returns after the update has been applied to a segment. After enabling `prevent_unoptimized`, the response is held until every deferred point, including the current update, has been indexed and is visible for search. Depending on the volume of updates and the speed of the optimizer, this can take a significant amount of time and may lead to timeouts on the client side. If the client times out, the update can be expected to be durably stored and eventually indexed, but the client will not receive a confirmation for that specific request.
```
timvisee:
```suggestion
Under normal conditions, setting `wait=true` on a write request returns after the update has been applied to a segment. After enabling `prevent_unoptimized`, the response is held until every deferred point, including the current update, has been indexed and is visible for search. Depending on the volume of pending updates in the update queue and the speed of the optimizer, this can take a significant amount of time and may lead to timeouts on the client side. If the client times out, the update can be expected to be durably stored and eventually indexed, but the client will not receive a confirmation for that specific request.
```

qdrant-landing/content/documentation/search/low-latency-search.md
```diff
@@ -47,10 +45,20 @@ Shards store their data in [segments](/documentation/manage-data/storage/). Writ
...
 
-Enabling `indexed_only` can cause recently updated data to temporarily disappear from search results until it is indexed again. To mitigate this, set the `prevent_unoptimized` optimizer setting to `true` when [creating or updating a collection](/documentation/manage-data/collections/#update-collection-parameters), or globally in the [configuration file](/documentation/operations/optimizer/). It prevents creating segments with a large amount of unindexed data for searches. Instead, once a segment reaches the so called `indexing_threshold`, all additional points will be added in 'deferred state'. Deferred points are not yet visible in reads but are still handled in writes. Deferred points will be promoted to visible points once the segment is optimized.
+With the [search parameter](/documentation/search/search/#search-api) `indexed_only` set to `true`, search operations only consider indexed data, ensuring more consistent and lower response times. The tradeoff is that the most recent data might not be included in search results until it has been indexed. It may also cause recently updated data to temporarily disappear from search results until it is indexed again.
```
timvisee:
```suggestion
With the [search parameter](/documentation/search/search/#search-api) `indexed_only` set to `true`, search operations only consider indexed data, ensuring more consistent and lower response times. The tradeoff is that the most recent data might not be included in search results until it has been indexed. It may also cause recently updated data to temporarily disappear from search results until it is indexed again. You may use `prevent_unoptimized` instead to mitigate this.
```

qdrant-landing/content/documentation/search/low-latency-search.md
```diff
@@ -47,10 +45,20 @@ Shards store their data in [segments](/documentation/manage-data/storage/). Writ
...
+### `prevent_unoptimized` Optimizer Setting
+
+*Available as of v1.17.0*
 
 <aside role="alert"><code>prevent_unoptimized</code> is an experimental feature; its behavior may change slightly in future releases and it must be used with care.</aside>
+
+Because updates in Qdrant are implemented as a delete followed by an insert, a side effect of searching indexed data only is that it can cause recently updated data to temporarily disappear from search results until it is indexed again. This is because the delete operation immediately removes the old point from the index, while the insert operation adds the new point to an unindexed segment that is not yet visible to searches.
```
timvisee: While this is true, it isn't the primary case for 'blinking' points.

Imagine our current indexing threshold is just one point.

If we now insert point A in a new segment, it is visible in search with `indexed_only=true` because the segment is small enough to be considered indexed. If we point insert point B into the same segment it goes over the indexing threshold, both point A and B now disappear from searches. So inserting data may make unrelated points disappear. `prevent_unoptimized` protects against this.

qdrant-landing/content/documentation/operations/optimizer.md
abdonpijpelink: I've removed the sentences.

qdrant-landing/content/documentation/search/low-latency-search.md
abdonpijpelink: I've made several changes to make the wording more precise.

qdrant-landing/content/documentation/ops-optimization/optimizer.md
```diff
@@ -121,43 +121,40 @@ In addition to the configuration file, you can also set optimizer parameters sep
...
+To address this, Qdrant supports [querying indexed data only](/documentation/search/low-latency-search/#query-indexed-data-only), by setting `indexed_only` to `true`. A side effect of searching indexed data only is that it can cause recently updated data to temporarily disappear from search results until it is indexed again ("blinking" points).
 
-To mitigate this, the optimizer supports a `prevent_unoptimized` mode. When enabled, points written to an unindexed segment that is larger than `indexing_threshold` are accepted and durably stored but are not visible in search results until the optimizer has indexed the segment. These are called deferred points. Not until the optimizer finishes indexing a segment containing deferred points, do those points become visible.
+To mitigate this, the optimizer supports a `prevent_unoptimized` mode. When enabled, points written to an unindexed segment that is larger than `indexing_threshold` are accepted and durably stored but are not visible in search results. These "deferred" points only become visible after the optimizer has indexed the segment.
```
timvisee: > the optimizer supports a `prevent_unoptimized` mode

It's not strictly the optimizer that supports/provides this.

But I guess it's fine to word it like this. I don't have a better way to state it.

qdrant-landing/content/documentation/operations/optimizer.md
```diff
@@ -121,43 +121,40 @@ In addition to the configuration file, you can also set optimizer parameters sep
...
 
-With `prevent_unoptimized` enabled, setting `indexed_only` to `true` is not necessary to avoid slow searches, as unindexed segments do not return deferred points.
-
-| `prevent_unoptimized` | `indexed_only` | Effect |
-|----------------------|----------------|--------|
-| `false` (default) | `false` | All points are searchable, but searches may be slow if there are many unindexed points. |
-| `false` (default) | `true` | Only indexed points are searchable, but recently updated points may temporarily disappear from search results until they are indexed. |
-| `true` | `false` (default) | Reads return indexed points and unindexed points that are not deferred. Deferred points are not visible to reads until indexed. |
-| `true` | `true` | Only indexed points are visible to reads. |
+With `prevent_unoptimized` enabled, setting `indexed_only` to `true` is not necessary.
```
timvisee:
```suggestion
With `prevent_unoptimized` enabled, setting `indexed_only` to `true` is not necessary. They are mutually exclusive.
```

qdrant-landing/content/documentation/operations/optimizer.md
```diff
@@ -121,43 +121,40 @@ In addition to the configuration file, you can also set optimizer parameters sep
...
+Under normal conditions, setting `wait=true` on a write request returns after the update has been applied to a segment. After enabling `prevent_unoptimized`, the response is held until every deferred point, including the current update, has been indexed and is visible for search. Depending on the volume of pending updates in the update queue and the speed of the optimizer, this can take a significant amount of time and may lead to timeouts on the client side. If the client times out, the update can be expected to be durably stored and eventually indexed, but the client will not receive a confirmation for that specific request.
```
timvisee:
```suggestion
Under normal conditions, setting `wait=true` on a write request returns after the update has been applied to a segment. After enabling `prevent_unoptimized`, the response is held until every deferred point, including the current update, has been indexed and is visible for search. Depending on the volume of pending updates in the update queue and the speed of the optimizer, this can take a significant amount of time and will likely lead to timeouts on the client side. If the client times out, the update can be expected to be durably stored and eventually indexed, but the client will not receive a confirmation for that specific request.
```

qdrant-landing/content/documentation/search/low-latency-search.md
```diff
@@ -47,10 +45,26 @@ Shards store their data in [segments](/documentation/manage-data/storage/). Writ
...
+### `prevent_unoptimized` Optimizer Setting
+
+*Available as of v1.17.1*
 
 <aside role="alert"><code>prevent_unoptimized</code> is an experimental feature; its behavior may change slightly in future releases and it must be used with care.</aside>
+
+To mitigate "blinking" points, an alternative to using `indexed_only` is to set the `prevent_unoptimized` optimizer setting to `true`. This prevents the creation of large segments with unindexed data. Instead, once a segment reaches the `indexing_threshold`, all additional points will be added in a "deferred" state. Deferred points are not yet visible in reads but are available to write operations. Deferred points are promoted to visible points once the segment has been optimized.
```
timvisee: Nit: saying that deferred points are available in context of writes may be confusing. This might be better:

```suggestion
To mitigate "blinking" points, an alternative to using `indexed_only` is to set the `prevent_unoptimized` optimizer setting to `true`. This prevents the creation of large segments with unindexed data. Instead, once a segment reaches the `indexing_threshold`, all additional points will be added in a "deferred" state. Deferred points are not yet visible in reads but are handled in write operations. Deferred points are promoted to visible points once the segment has been optimized.
```

timvisee (commented):
Some nits:

qdrant-landing/content/documentation/ops-optimization/optimizer.md
abdonpijpelink: We can say "Qdrant supports a `prevent_unoptimized` mode"?

timvisee (approved):

PR #2290, opened 2026-04-21
Use section titles in docs breadcrumbs

abdonpijpelink:
The breadcrumbs at the top of the navigation currently use the URL slugs as display values. This is not ideal for some URLs, for example:

image

This PR changes that so the breadcrumbs use the section title instead of the URL slugs. If no section title is defined, it falls back to the URL slug. Result:

image

This PR also renames the "Home" section into "Documentation", to match the original breadcrumb. This has the side effect that page `<title>`s (and as a result, also browser tabs) now say "Documentation" instead of "Home", which IMO is an improvement.

### DIFF
--- a/qdrant-landing/content/documentation/_index.md
+++ b/qdrant-landing/content/documentation/_index.md
@@ -1,5 +1,5 @@
 ---
-title: Home
+title: Documentation
 weight: 2
 hideTOC: true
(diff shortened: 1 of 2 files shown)


kanungle (approved):
Amazing... thank you!

PR #2292, opened 2026-04-21
docs: Retrival quality guide improvement

Dylancouzon:
## Summary

Splits the Retrieval Quality tutorial into three focused pages and aligns terminology with the ANN-benchmarks convention.

## Changes

- **New:** `retrieval-quality-fundamentals.md` three evaluation levels (ANN recall, retrieval relevance, business impact), evaluation ladder, and a "which metric when" table.
- **New:** `retrieval-quality-golden-set.md` how to build ground-truth data (log mining, LLM synthetic, human annotation), a runnable "Using the Golden Set" section with `recall@k` / `MRR` / `NDCG@k` implementations and an `evaluate()` loop against `client.query_points`, and data-leakage pitfalls to avoid.
- **Updated:** `retrieval-quality.md` narrowed to HNSW-tuning content, renamed `precision@k` → `recall@k` throughout (ANN Benchmarks conventions), fixed a prose/code mismatch (50000 → 60000).
- Both nav indexes (`search-engineering.md`, `tutorials-lp-overview.md`) updated.
- Original aliases preserved so no broken links.

## Validation

All code blocks in `retrieval-quality-golden-set.md` were run end-to-end against a real Qdrant cluster and the Anthropic API. The harness is at https://github.com/Dylancouzon/golden-set-harness — it builds a small labeled corpus where each query's relevant doc is known in advance, then verifies `recall_at_k` / `mrr` / `ndcg_at_k` and the tutorial's `evaluate()` produce the expected values. Re-run before merging future edits to the tutorial.

## Why

The old page mixed concepts, ground-truth methodology, and hands-on tuning into one document. Customers couldn't find metric guidance or golden-set construction, and the `precision`/`recall` terminology was inconsistent. Splitting into three pages maps cleanly onto those three gaps.


kanungle (dismissed):
please put related repos in qdrant-labs org. Just invited you

mrscoopers:
@abdonpijpelink started reviewing this one & noticed that "search engineering" now has (...too many tutorials):
1. Semantic Search Basics (probably they should be in basics?)
2. "Hybrid Search with Reranking" & "multivectors and late interaction" with a very look-alike pipeline, even though they cover 2 different angles. Wonder if it could be optimized code flow/snippets-wise
3. Hybrid Search with FastEmbed, which probably should be in "FastEmbed" (?)
4. PDF Retrieval got renamed to "Multivector Document Retrieval," which is not very descriptive, especially already having "multivectors and late interaction.". It needs to somehow show in the title that it's about VLMs, e.g. working with images
Wdyt?

mrscoopers:
@Dylancouzon @kanungle @abdonpijpelink I wonder if we could/should compress these 3 into 1 or 2 guides?
Maybe Retrieval Quality Fundamentals could go to Basics, and Golden Query set & Retrieval Quality Evaluation become one guide, "Retrieval Quality Evaluation" in "Search Engineering"? 
As in this section, usually, less is more.

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-fundamentals.md
```diff
@@ -0,0 +1,74 @@
...
+aliases:
+  - /documentation/tutorials/retrieval-quality-fundamentals/
+---
+
+# Retrieval Quality Fundamentals
+
+| Time: 20 min | Level: Intermediate |  |    |
+|--------------|---------------------|--|----|
+
+Before measuring retrieval quality, it's worth understanding what you're measuring. Retrieval quality operates at three distinct levels, and it's easy to optimize for the wrong one.
```
mrscoopers: I'd relax the statements throughout the guide, as afaik, evals in production reality are very ugly/mixed/non-dividable
Aka more of "operates" -> "could be classified on"

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-fundamentals.md
```diff
@@ -0,0 +1,74 @@
...
+
+The first level is **ANN recall**: does the approximate search return the same results as an exact nearest-neighbor search? This is a purely algorithmic question about how faithfully HNSW approximates exhaustive search. It has nothing to do with whether those results are useful to a human.
```
mrscoopers: This is true only for dense vector search (aka search on vector index), for sparse vectors it is not (and we support them), so I'd mention it, to not trap ourselves into "always measure ANN vs KNN."
"Is your vector index as good as a full scan"

mrscoopers: > It has nothing to do with whether those results are useful to a human.

I'd avoid statements like this, as in the end, bad ANN recall affects the usability of the search pipeline

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-fundamentals.md
```diff
@@ -0,0 +1,74 @@
...
+
+The second level is **retrieval relevance**: of the results returned, how many are relevant to the query intent? This requires a labeled ground-truth dataset or human judgment. A pipeline can achieve near-perfect ANN recall and still surface irrelevant documents if the embeddings are a poor fit for the task.
```
mrscoopers: (Or LLM judgement)

mrscoopers: Keep in mind that we have sparse vectors allowing for exact match search (no ANN), and this fundamentals guide should not completely ignore them

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-fundamentals.md
```diff
@@ -0,0 +1,74 @@
...
+
+The third level is **business impact**: does better retrieval lead to better outcomes like lower hallucination rates in downstream LLMs, higher task-completion rates, or improved user satisfaction scores? This is what stakeholders care about, but it's the hardest to measure directly. The causal chain from a vector match to a user outcome is long and easily dominated by generator behavior, UI, and other confounders, so no single offline metric is a reliable proxy for a KPI. The next section describes how teams bridge this gap in practice.
```
mrscoopers: > The next section describes how teams bridge this gap in practice.

I feel like this statement might put a burden of expectation on us, as if we used actual client pipelines to write this article
Did we(?)

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-fundamentals.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Connecting the Levels in Practice
+
+The three levels aren't measured in isolation. Teams that successfully connect retrieval work to business outcomes tend to build an **evaluation ladder** that runs each layer at a different cadence and cost, and uses the result of each layer to decide whether to invest effort at the next:
+
+| # | Layer | What it measures | Cadence | Cost |
```
mrscoopers, line 29: Let's remove the cost & cadence? As usually "it depends."
We could do instead something like "Notes" or "Good to know's" with facts like "golden set is a real cost"

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-fundamentals.md
```diff
@@ -0,0 +1,74 @@
...
+|---|---|---|---|---|
+| 1 | ANN recall | `Recall@k` vs exact kNN on a sampled query set | On index or embedding changes | Low |
+| 2 | Retrieval relevance | `Recall@k` / `NDCG@k` vs a labeled golden set | Weekly, or on retrieval-stack changes | Low per run; **golden set is the real cost** (see next tutorial) |
+| 3 | End-to-end answer quality | LLM-as-judge or human rating on the golden set | Weekly, or on retrieval or generator changes | Moderate (LLM-judge cost × eval size) |
+| 4 | Business impact | Online A/B behind a flag | Per release, once offline layers pass | High (traffic, experimentation infra) |
+
+Each layer is necessary but not sufficient. A win at layer 2 that doesn't carry through to layer 3 usually means the generator or the prompt is the bottleneck, not retrieval. This is the most useful diagnostic the ladder provides, and the reason teams shouldn't collapse layers 2 and 3 into a single score.
```
mrscoopers: > A win at layer 2 that doesn't carry through to layer 3 usually means the generator or the prompt is the bottleneck, not retrieval.

This statement sounds like retrieval is only for RAG (as "End-to-end answer quality")
But RAG is only one of the use cases, LLM on top does not always exist (still), like in e-commerce, for example

mrscoopers: So I'd specify that it's one of the use cases aka (if you use retrieval for some type of RAG)

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-fundamentals.md
```diff
@@ -0,0 +1,74 @@
...
+
+**Isolate the component under test.** When end-to-end quality moves, hold one side fixed: evaluate retrieval with the generator frozen, and evaluate the generator with retrieval frozen. Without this, attribution collapses into guesswork and the ladder stops being diagnostic.
```
mrscoopers: same as above

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-fundamentals.md
```diff
@@ -0,0 +1,74 @@
...
+
+**Proxy KPIs for teams without A/B infrastructure.** Most teams don't have the traffic or tooling to run a proper A/B. Cheap production signals that correlate with business value (click position on surfaced results, answer copy or share rate, session-level task completion, thumbs-up/down) can be instrumented long before a formal experimentation platform exists, and sit usefully between the golden-set layer and the full A/B.
```
mrscoopers: Tbf it's a whole research domain, as A/B for RAG and agentic systems is basically absent & no one properly figured out how to do it. Hence, I'd mention here that it's all very grey zone & complicated.

abdonpijpelink:
> "search engineering" now has (...too many tutorials)

I agree. Your idea to move the hybrid search tutorials to the Basics section is a good one, and we should take a good look at the titles. Let's do that in a separate PR.

> compress these 3 into 1 or 2 guides

Perhaps combine "Retrieval Quality Fundamentals" and "Evaluate Retrieval Quality" into one and keep "Building a Golden Query Set" as a separate tutorial? Then we have one tutorial for layer 1, and one for layer 2.

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-fundamentals.md
abdonpijpelink: I feel we may need a bit more of an introduction before this. Why are we interested in measuring retrieval quality? Maybe some of the introduction of the "Evaluate Retrieval Quality" tutorial should move here.

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-fundamentals.md
abdonpijpelink: Write out "Approximate Nearest Neighbor (ANN)" on first use. Maybe also provide a bit more context here? Something like _"Vector search engines such as Qdrant do not perform pure k-Nearest Neighbors (kNN) search. Instead, they use **Approximate Nearest Neighbors** (ANN) algorithms, which are much faster than the exact search, but can return suboptimal results."_ from the "Evaluate Retrieval Quality" tutorial.

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-golden-set.md
```diff
@@ -0,0 +1,125 @@
...
+  - /documentation/tutorials/retrieval-quality-golden-set/
+---
+
+# Building a Golden Query Set
+
+| Time: 40 min | Level: Intermediate |  |    |
+|--------------|---------------------|--|----|
+
+Evaluating retrieval relevance requires a labeled dataset of queries paired with their expected relevant documents (commonly called a *golden query set* or *ground truth*). The [Retrieval Quality Evaluation](/documentation/tutorials-search-engineering/retrieval-quality/) tutorial measures **ANN recall** against exact kNN, which needs no relevance labels. This page covers the separate task of building labeled data to measure **retrieval relevance** against real user intent.
+
```
abdonpijpelink, line 17: I feel a link to the levels table in "Retrieval Quality Fundamentals", and an explanation that this tutorial focuses on level 2 would be helpful here.

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-golden-set.md
```diff
@@ -0,0 +1,125 @@
...
+### 3. LLM-Based Synthetic Generation (Scales Cheaply, Lowest Fidelity)
+
+When logs and reviewers aren't available, prompt an LLM to generate plausible queries for each document. This scales to thousands of pairs, but synthetic queries are easier to retrieve than what real users type.
+
+```python
+import os
+
+import anthropic
+
+client = anthropic.Anthropic()
```
abdonpijpelink:
```suggestion
client = Anthropic(
    api_key=os.environ.get("ANTHROPIC_API_KEY"),
)
```
We don't want to go into the details of the Anthropic API here, but I think the code as-is won't work without setting the `ANTHROPIC_API_KEY` as an environment variable. So maybe making it explicit like this makes that more obvious?

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-golden-set.md
```diff
@@ -0,0 +1,125 @@
...
+        messages=[{
+            "role": "user",
+            "content": (
+                f"Generate {n} short, realistic search queries that would lead a user to the "
+                f"following document. Return only the queries, one per line.\n\n{doc_text}"
+            ),
+        }],
+    )
+    return response.content[0].text.strip().splitlines()
+```
```
abdonpijpelink, line 61: Maybe mention that using the Anthropic API is just an example, and that you can use any LLM you want to use?

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality.md
abdonpijpelink: This is a problem with the existing tutorial, but as long as we're updating it, maybe consistently use title case for all headers?

abdonpijpelink: I feel a link to the levels table in "Retrieval Quality Fundamentals", and an explanation that this tutorial focuses on level 1 would be helpful here.

abdonpijpelink: This is also a problem with the existing tutorial, but as long as we're updating it, maybe we can change it such that we don't first load the 60K items in memory but rather stream the data from disk by passing the iterator directly to `upload_points` with a `batch_size`?

abdonpijpelink (commented):
This looks great!

I've left a few suggestions, none of them blockers.

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-fundamentals.md
```diff
@@ -0,0 +1,74 @@
...
+
+**Pre-register the decision rule.** Before running the A/B, write down what constitutes a win and what constitutes a no-ship, in terms of both the retrieval metric and the KPI. This is the highest-leverage discipline for avoiding "recall improved but the KPI didn't, the KPI is noisy, let's ship anyway" rationalization.
```
mrscoopers, line 42: Could we maybe leave some links to resources supporting this advice (?) For credibility

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-fundamentals.md
```diff
@@ -0,0 +1,74 @@
...
+
+**Tooling.** Qdrant owns layers 1 and 2 directly. For layer 3, the ecosystem has mature tooling like [Ragas](https://docs.ragas.io/), [Arize Phoenix](https://phoenix.arize.com/), and [DeepEval](https://docs.confident-ai.com/) that handles LLM-as-judge scoring and offline answer-quality eval.
```
mrscoopers: Qdrant provides tooling only for ANN vs KNN (layer 1) in WebUI (I'd express it more explicitly)
For layer 2, there's a lot of tooling out there, as search metrics are here since forever, we could mention `ranx` & link Qdrant Essentials, or/and multivector course, where Kacper uses it for evals, for example

For Layer 3, I'd also link here the ecosystem section (I feel like it'd be nice to somehow explicitly group tooling for evals there? cc @abdonpijpelink wdyt)

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-fundamentals.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Quality Metrics
```
mrscoopers, line 46: I think it makes sense to structure this section by layers, too?
I mean order, mainly: ANN vs KNN for dense vector search, search relevance metrics (for any kind of search, not only semantic), (agentic) RAG metrics, online metrics

mrscoopers: As now it's search relevance metrics (applicable to lexical search, too), then suddenly ANN

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-fundamentals.md
mrscoopers: Also, ANN recall is not only tied to HNSW but to other vector indexes. We support only (filterable) HNSW, though, so maybe I'd add a small explanation of 1-2 sentences or/and references on what's ANN and why ANN recall matters

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-fundamentals.md
mrscoopers: Also, I'd move tooling to the next section, maybe, where we discuss choosing metrics? Aka make it more like "choosing instruments"/"what to pick"? Then we'll cover advices per layer +-

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-fundamentals.md
```diff
@@ -0,0 +1,74 @@
...
+
+There are various ways to quantify the quality of semantic search. Some of them, such as [Precision@k](https://en.wikipedia.org/wiki/Evaluation_measures_(information_retrieval)#Precision_at_k),
+are based on the number of relevant documents in the top-k search results. Others, such as [Mean Reciprocal Rank (MRR)](https://en.wikipedia.org/wiki/Mean_reciprocal_rank),
+take into account the position of the first relevant document in the search results. [DCG and NDCG](https://en.wikipedia.org/wiki/Discounted_cumulative_gain)
+metrics are, in turn, based on the relevance score of the documents.
+
+To evaluate the ANN algorithm itself, the question is simple: of the `k` true nearest neighbors an exact search would return, how many did the approximation find? That fraction is **`recall@k`**:
```
mrscoopers: exact search -> full scan vector search

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-fundamentals.md
```diff
@@ -0,0 +1,74 @@
...
+
+`recall@k = |ANN results ∩ exact results| / k`
+
+When both searches return exactly `k` items, `recall@k` and `precision@k` are numerically identical. The ANN community uses "recall" by convention to make clear that exact kNN is the ground truth.
```
mrscoopers: Didn't get this one 
"When both searches return exactly `k` items"?

"The ANN community uses "recall" by convention to make clear that exact kNN is the ground truth." Statements like this we have to reference (I didn't know there'a an ANN community:D)

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-fundamentals.md
```diff
@@ -0,0 +1,74 @@
...
+
+### Choosing the Right Metric
```
mrscoopers, line 58: Nicce

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-golden-set.md
```diff
@@ -0,0 +1,125 @@
...
+aliases:
+  - /documentation/tutorials/retrieval-quality-golden-set/
+---
+
+# Building a Golden Query Set
+
+| Time: 40 min | Level: Intermediate |  |    |
+|--------------|---------------------|--|----|
+
+Evaluating retrieval relevance requires a labeled dataset of queries paired with their expected relevant documents (commonly called a *golden query set* or *ground truth*). The [Retrieval Quality Evaluation](/documentation/tutorials-search-engineering/retrieval-quality/) tutorial measures **ANN recall** against exact kNN, which needs no relevance labels. This page covers the separate task of building labeled data to measure **retrieval relevance** against real user intent.
```
mrscoopers: I think we should send people for ANN vs KNN to our WebUI, where there's a tab which measures ANN recall, not to the tutorial(?)

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-golden-set.md
```diff
@@ -0,0 +1,125 @@
...
+
+### 1. Human Annotation (Highest Quality, Highest Cost)
+
+Domain experts assign relevance scores on a binary (relevant / not relevant) or graded (0/1/2 or 1–5) scale. This is the cleanest approach for high-stakes applications and produces the graded labels NDCG needs. Expert time is the bottleneck, so reserve it for a small, high-value subset — the hardest queries or the ones that matter most commercially — and use the other two approaches for coverage.
+
+### 2. Real User Queries from Logs (High Realism, Requires Production Traffic)
+
+If your app records queries with click or explicit-feedback signals, sample query-document pairs directly. This captures real user intent and vocabulary, and should be your first choice once production traffic exists. Stratify by query type or topic cluster — uniform sampling over-represents frequent queries and misses rare-but-important cases. As a rough heuristic, a few hundred labeled pairs detects large metric differences; small ranking differences or per-slice analysis need substantially more. Treat any number as a starting point and widen confidence intervals if the signal is noisy.
```
mrscoopers: > Stratify by query type or topic cluster

I believe this works for the old search systems with SERP, but it badly maps to RAG/agentic conversational systems using retrieval in the background. 
Hence, maybe, I'd try to once again use the tool of toning down the "we know 100% what we're advising" tone, especially if it's not backed by references

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-golden-set.md
```diff
@@ -0,0 +1,125 @@
...
+
+### 3. LLM-Based Synthetic Generation (Scales Cheaply, Lowest Fidelity)
+
+When logs and reviewers aren't available, prompt an LLM to generate plausible queries for each document. This scales to thousands of pairs, but synthetic queries are easier to retrieve than what real users type.
+
+```python
```
mrscoopers, line 36: Why not instead advise tooling for it? I feel like adding code requires us to think about its maintenance, and since it's external APIs (which change randomly, hell knows when & how), it becomes an unnecessary burden.

mrscoopers: As afaik certain frameworks provide the capability to generate synthetic sets for evals (idk maybe we even have some intergations in ecosystem tab)

mrscoopers: So to TL;DR: I'd avoid code snippets using external APIs here

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-golden-set.md
```diff
@@ -0,0 +1,125 @@
...
+        }],
+    )
+    return response.content[0].text.strip().splitlines()
+```
+
+## Using the Golden Set
+
+Once queries are labeled, run each through Qdrant and compare the returned IDs against the labels. The metric to compute depends on what the labels record (see the [metric-selection table](/documentation/tutorials-search-engineering/retrieval-quality-fundamentals/#choosing-the-right-metric)).
+
+For **binary-relevance labels** (a set of relevant doc IDs per query), compute `recall@k` and `MRR`:
```
mrscoopers: I'd rather use ranx for all of 'em metrics, as we started using it already in many places & metrics there are standartised (you can check examples in Qdrant Essentials on day 3, f.e.) OR provide code snippets for every client SDK we support (Rust, for example, too)

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-golden-set.md
```diff
@@ -0,0 +1,125 @@
+---
+title: Building a Golden Query Set
```
mrscoopers, line 2: I think it should be merged with the 3rd one, as this one mainly is about using the golden set after it's generated, anyway, so they merge well into a more in-depth guide (and 1st one is a bird view)!

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-golden-set.md
```diff
@@ -0,0 +1,125 @@
...
+        relevant = set(entry["relevant_ids"])
+        recalls.append(recall_at_k(retrieved, relevant, k))
+        mrrs.append(mrr(retrieved, relevant))
+    return {
+        "mean_recall": sum(recalls) / len(recalls),
+        "mean_mrr": sum(mrrs) / len(mrrs),
+    }
+```
+
+This produces the retrieval-relevance score (layer 2 of the [evaluation ladder](/documentation/tutorials-search-engineering/retrieval-quality-fundamentals/#connecting-the-levels-in-practice)). Re-run it whenever the retrieval stack changes: new embedding model, new index config, new reranker. To measure layer 1 (ANN recall against exact kNN), see [Retrieval Quality Evaluation](/documentation/tutorials-search-engineering/retrieval-quality/).
```
mrscoopers: > To measure layer 1 (ANN recall against exact kNN)

See WebUI?

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-golden-set.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Pitfalls to Watch For (Data Leakage and Friends)
+
+In golden query sets, **data leakage** means any setup that makes offline metrics look better than production reality. Unlike classic train/test leakage, the issue is often evaluation design. Keep source documents in the index (they are the expected relevant answers). Focus on these risks:
```
mrscoopers, line 101: Here I'd also back statements with references

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality.md
mrscoopers: > Semantic search pipelines are as good as the embeddings they use.

I don't like this statement. There're many techniques/tools to improve search pipelines beyond just a pick of the model, but model stes up a certain baseline, for sure

mrscoopers: > In the [evaluation process]

Hm, link is weird here, it's not like we're doing only RAG evals

mrscoopers: Misleading title of a section, ANN approximation != retrieval quality

mrscoopers: In general, this subsection makes sense, but I am a hater of the idea (that I get from this tutorial) that the whole vector search recipe is right embedding + right ANN
It's not true & weakens our positioning: there are many amazing techniques based on vector specifics to increase vector search relevance & also matters how retrieval pipeline is designed, aka is it hybrid, for example

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality.md
```diff
@@ -34,28 +34,17 @@ perform pure kNN search. Instead, they use **Approximate Nearest Neighbors** (AN
...
-If we treat the search pipeline as a whole, we could use them all. The same is true for the embeddings quality evaluation. However, for the 
-ANN algorithm itself, anything based on the relevance score or ranking is not applicable. Ranking in vector search relies on the distance
-between the query and the document in the vector space, however distance is not going to change due to approximation, as the function is
-still the same. 
-
-Therefore, it only makes sense to measure the quality of the ANN algorithm by the number of relevant documents in the top-k search results, 
-such as `precision@k`. It is calculated as the number of relevant documents in the top-k search results divided by `k`. In case of testing
-just the ANN algorithm, we can use the exact kNN search as a ground truth, with `k` being fixed. It will be a measure on **how well the ANN
-algorithm approximates the exact search**.
+For a broader discussion of what to measure and when (ANN recall vs retrieval relevance vs business impact, and which metric fits which scenario),
```
mrscoopers: I think it makes more sense if we have an overview in /documentation/tutorials-search-engineering/retrieval-quality-fundamentals/ (and put it in the basics section), and here we'll actually go deeper into each layer (ANN vs KNN, then golden set & retrieval metrics comparison & then recommendation of tooling for layers 2/3/4)

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality.md
```diff
@@ -34,28 +34,17 @@ perform pure kNN search. Instead, they use **Approximate Nearest Neighbors** (AN
...
+see [Retrieval Quality Fundamentals](/documentation/tutorials-search-engineering/retrieval-quality-fundamentals/). This tutorial focuses on the
+ANN-algorithm layer and measures it with `recall@k`: the fraction of the true top-k items returned by exact search that the approximate search
+recovers. When both ANN and exact search return exactly `k` items, `recall@k` and `precision@k` are numerically identical; we use "recall" to
+match the ANN-benchmarks convention.
 
 ## Measure the quality of the search results
```
mrscoopers: I think it'd be much more beneficial if we explain here how the WebUI tool works & how to use it to measure ANN vs KNN, based on the same example & flow, and not use python snippets

mrscoopers: and use sample datasets in WebUI
Afaik, the rule of thumb: The more SDK agnostic & the less dependent on external libraries we can get in tutorials on this page, the better

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality.md
```diff
@@ -155,37 +144,37 @@ def avg_precision_at_k(k: int):
...
-avg(precision@5) = 0.9935999999999995
+avg(recall@5) = 0.9935999999999995
 ```
 
-As we can see, the precision of the approximate search vs exact search is pretty high. There are, however, some scenarios when we
-need higher precision and can accept higher latency. HNSW is pretty tunable, and we can increase the precision by changing its parameters.
+As we can see, the recall of the approximate search vs exact search is pretty high. There are, however, some scenarios when we
+need higher recall and can accept higher latency. HNSW is pretty tunable, and we can increase the recall by changing its parameters.
   
 ## Tweaking the HNSW parameters
```
mrscoopers: Afaik, we have something like this in Qdrant Essentials? I'd link and recycle

mrscoopers (changes requested):

mrscoopers:
> Perhaps combine "Retrieval Quality Fundamentals" and "Evaluate Retrieval Quality" into one and keep "Building a Golden Query Set" as a separate tutorial? Then we have one tutorial for layer 1, and one for layer 2.
@abdonpijpelink @Dylancouzon @kanungle 

I read all 3, and I am of the strong opinion that it rather makes sense as
- Overview of all 4 layers in Basics (aka tutorial 1) - it's even more like an article or blog to me, aka an opinion piece

- Merging 2 and 3 into one tutorial "Evaluate Retrieval Quality", where we a) Explain how to use WebUI to evaluate ANN vs KNN and the motivation behind it + HNSW tuning reference to another tutorial/guide or Qdrant Essentials b) Golden Set + retrieval eval classical metrics (recall, precision, NDCG, MRR)

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-golden-set.md
Dylancouzon: I’ll review this with Neil in our 1:1. Agreed the first one doesn’t belong in the tutorial section. I’d keep the other two separate to make each concept easier to work through, and to avoid digging through unrelated content when building a golden dataset. That said, I see both sides.

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-golden-set.md
Dylancouzon: Fixed in 6eba7e2
Thanks

Dylancouzon: Addressed in 6eba7e2. The intro now opens by anchoring the tutorial at layer 2 of the evaluation ladder and links directly to the levels table in the Fundamentals page.

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-golden-set.md
Dylancouzon: Addressed in acc85e6. Swapped the retrieval-quality tutorial link for the **Search Quality** tab in the Qdrant Web UI. One thing worth flagging: `/documentation/web-ui/` doesn't currently document the Search Quality tab, so readers land on the general page rather than a specific anchor. Worth a short follow-up PR to add a section there.

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-golden-set.md
Dylancouzon: Addressed in 6eba7e2. Removed the duplicate layer-1 pointer from the end of this section since the intro now handles it (and, as of acc85e6, sends readers to the Web UI Search Quality tab).

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-golden-set.md
Dylancouzon: Addressed in a47c57eb. Kept the Anthropic snippet since it's a generic `messages.create` call unlikely to rot, and added a pointer to Ragas for readers who want a maintained testset generator. Didn't swap the snippet out: Ragas is RAG-opinionated and heavier than the prompt shape warrants here.

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-golden-set.md
Dylancouzon: Addressed in a47c57eb. Switched the metrics section to ranx. One `Qrels` / `Run` construction, one `evaluate()` call handles recall@k, MRR, NDCG@k, and others for both binary and graded labels. Replaced the three hand-rolled functions and the evaluate() loop.

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-golden-set.md
Dylancouzon: Great callout, done

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-golden-set.md
Dylancouzon: Still believe that keeping the snippet is beneficial so readers Can see how to practically implement the solution without having to navigate multiple pages at a time.

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-golden-set.md
Dylancouzon: Addressed in 81883cbe. Split the stratification sentence into two short use-case lines so "query type / topic cluster" applies to search-style traffic and "conversation turn / intent class" applies to RAG or agentic retrieval. Softened `detects` to `can detect` too. On the broader tone point: the tutorial is intentionally prescriptive as a starting point for teams building their first golden set, so leaving the guidance framing in place — happy to revisit specific lines that feel overstated.

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-golden-set.md
Dylancouzon: The Pitfalls bullets read as practical guidance rather than novel claims, and the tutorial is already prescriptive by design. Happy to add a reference or two if there are specific statements you'd like backed — which ones stand out?

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality.md
Dylancouzon: @kanungle What are your thoughts on that?

Dylancouzon: Addressed in fd095af8. Title-cased the six H2 headers in this tutorial.

Dylancouzon: Addressed in fd095af8. Added a layer-1 anchor under the Time/Level table linking to the levels table in Fundamentals, mirroring the golden-set tutorial's anchor.

Dylancouzon: Addressed in e5d91012. Rewrote the intro so the ANN algorithm reads as one of several levers that shape retrieval quality.

Dylancouzon: Addressed in e5d91012. Same rewrite as the line-16 thread; not sure why the original wording was written that way.

Dylancouzon: Dropped in e5d91012. The `/rag/rag-evaluation-guide/` link appeared three times in this file and all three are gone; agreed this tutorial isn't RAG-specific. Not sure why those links were included.

Dylancouzon: Renamed to "ANN Recall" in e5d91012. Matches the layer-1 anchor and is accurate about what the section measures.

Dylancouzon: Obsolete after 921b645e, the whole setup section was cut. Readers are expected to have their own collection at this point, so the 60K list materialization is gone entirely.

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality.md
Dylancouzon: Addressed in 921b645e. Trimmed the overview content here and narrowed the tutorial to layer 1 (ANN vs kNN) via the Search Quality tab. Fundamentals handles the ladder overview, and the golden-set tutorial handles layer 2.

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality.md
Dylancouzon: Addressed in 921b645e. Pivoted to the Search Quality tab: default one-click ANN vs kNN run, then advanced mode to tune HNSW inline and re-run. Kept Python in a separate "Automate in CI" section with a reusable skeleton function.

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality.md
Dylancouzon: Linked to [Optimize Performance](/documentation/operations/optimize/) in the HNSW section for the parameter reference. Didn't pull in Qdrant Essentials since the Web UI's advanced mode covers the hands-on flow inline. Happy to add an Essentials pointer if you want course-level coverage called out.

Dylancouzon:
Two changes on retrieval-quality.md:

 Title: "Retrieval Quality Evaluation" → "Measuring ANN Precision." The old title overclaimed. After the Web UI pivot the tutorial is narrowly about layer 1 (ANN vs exact kNN), not retrieval
  quality in general. URL is unchanged so existing links still work.

 Metric: recall@k → precision@k. The Search Quality tab in the Web UI labels this precision@k; aligning the tutorial to what readers see on screen. Numerically identical here (both searches return
  exactly k items), so no semantic shift. Fundamentals and the golden-set cross-reference were updated for consistency.

kanungle:
linkChecker is failing because there's a url referenced that was recently moved

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-fundamentals.md
Dylancouzon: Addressed in e2df842f. Added a lead paragraph before the taxonomy framing why retrieval quality matters (downstream hallucinations, user bounce, wrong-tool agent actions) and how measurement catches regressions from model, index, or dataset changes.

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-fundamentals.md
Dylancouzon: Addressed in e2df842f. Expanded to "approximate nearest-neighbor (ANN) precision" on first use.

Dylancouzon: Addressed in e2df842f. Added a parenthetical clarifying that sparse vectors use exact matching and skip the ANN-precision layer, so the framing isn't implicitly dense-only.

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-fundamentals.md
Dylancouzon: Addressed in e2df842f. Layer-2 ground truth now lists labeled dataset, human judgment, and LLM-as-judge scoring.

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-fundamentals.md
Dylancouzon: Addressed in e2df842f. Broke the Tooling bullet into per-layer recommendations: Web UI Search Quality tab for layer 1, [ranx](https://amenra.github.io/ranx/) for layer 2, and Ragas/Phoenix/DeepEval for layer 3.

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-fundamentals.md
Dylancouzon: Addressed in e2df842f. Quality Metrics now leads with layer 1 (ANN recall formula and the exact-kNN equivalence note), then the generic layer-2 relevance metrics (Precision@k, MRR, NDCG).

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-fundamentals.md
Dylancouzon: Addressed in fb1672c1. Softened to "common patterns for bridging this gap" so the sentence doesn't imply we pulled patterns from real client pipelines.

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-fundamentals.md
Dylancouzon: Addressed in fb1672c1. The layer-2/3 diagnostic now names the downstream consumer as "an LLM generator in RAG, a ranker in a search UI, or the prompt itself," so the framing isn't RAG-only.

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-fundamentals.md
Dylancouzon: Addressed in fb1672c1. The "Isolate the component under test" bullet now says "the downstream piece frozen (an LLM generator, a ranker, or the UI)" instead of assuming a generator.

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-fundamentals.md
Dylancouzon: Addressed in fb1672c1. Added a one-line caveat to the Proxy KPIs paragraph acknowledging that A/B design for RAG and agentic systems is still evolving.

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-fundamentals.md
Dylancouzon: Addressed in fb1672c1. Simplified the equivalence note to "Because both result sets are size `k`, this formula is numerically identical to `precision@k`." and linked [ann-benchmarks.com](https://ann-benchmarks.com/) as the citation for the "recall by convention" claim.

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-fundamentals.md
Dylancouzon: Sparse vectors are covered in e2df842f: the layer-1 paragraph now has a parenthetical noting that "sparse vectors use exact matching, so this layer doesn't apply to them." Happy to expand further if the current mention feels too brief.

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-fundamentals.md
Dylancouzon: Keeping the prescriptive voice intentionally. The guide's value is in being opinionated rather than hedging with "it depends." Happy to relax specific lines if particular claims feel overreaching; which ones stood out?

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-fundamentals.md
Dylancouzon: Keeping "exact search." It's the Qdrant API term (`SearchParams(exact=True)`) and matches the Web UI label, so readers looking up the parameter find the same word. "Full scan vector search" is more descriptive but longer without adding precision.

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-fundamentals.md
Dylancouzon: Keeping the Cadence and Cost columns. "It depends" is always true but doesn't help devs who don't know what's normal; the current framing ("Weekly, or on retrieval-stack changes," "Low per run; golden set is the real cost") gives defaults readers can adjust. Can add a preamble calling them typical rather than prescriptive if overclaiming is the concern.

Dylancouzon: I think I was hangry when writing that comment above. 
If this becomes a blog post or opinion piece rather than a tutorial, the Notes-style structure you suggested would fit better than a rigid comparison table. Keeping the table for now and will revisit once the format is settled.

qdrant-landing/content/documentation/tutorials-search-engineering/retrieval-quality-fundamentals.md
Dylancouzon: This will likely shift into a more opinionated blog post. That means less focus on citations, more on first-hand framing, like “here’s how we think about this from patterns we’ve seen.”

Dylancouzon:
There's just too much going on here. Will open a new PR.

PR #2322, opened 2026-04-30
Update FAQ

abdonpijpelink:
Updates the FAQ with questions sourced from customers, prospects, and the community.

### DIFF
--- a/qdrant-landing/content/documentation/faq/qdrant-fundamentals.md
+++ b/qdrant-landing/content/documentation/faq/qdrant-fundamentals.md
@@ -57,7 +57,7 @@ There are two possible reasons for this:
 
 ### How many vectors can I store in a point? Can a point have no vector at all?
 
-A point can hold any number of dense, sparse, and multi vectors. There's no hard limit imposed by Qdrant, though practical limits apply: each additional vector increases memory usage, so the realistic ceiling is determined by available RAM and storage. You can attach a single vector, or multiple vectors with different names (for example, a dense vector for semantic search alongside a sparse vector for keyword matching). This lets you run [hybrid queries](/documentation/hybrid-queries/) over several representations of the same data within one collection. Each vector must be defined in the collection's schema.
+A point can hold any number of dense, sparse, and multi vectors. There's no hard limit imposed by Qdrant, though practical limits apply: each additional vector increases memory usage, so the realistic ceiling is determined by available RAM and storage. You can attach a single vector, or multiple vectors with different names (for example, a dense vector for semantic search alongside a sparse vector for keyword matching). This lets you run [hybrid queries](/documentation/search/hybrid-queries/) over several representations of the same data within one collection. Each vector must be defined in the collection's schema.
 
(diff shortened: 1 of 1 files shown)


qdrant-landing/content/documentation/faq/qdrant-fundamentals.md
```diff
@@ -55,6 +55,43 @@ There are two possible reasons for this:
...
 - You used the `uint8` [datatype](/documentation/manage-data/vectors/#datatypes) to store vectors. `uint8` requires a special format for input values, which might not be compatible with the typical output of embedding models.
 
+### How many vectors can I store in a point? Can a point have no vector at all?
+
+A point can hold any number of dense, sparse, and multi vectors. There's no hard limit imposed by Qdrant, though practical limits apply: each additional vector increases memory usage, so the realistic ceiling is determined by available RAM and storage. You can attach a single vector, or multiple vectors with different names (for example, a dense vector for semantic search alongside a sparse vector for keyword matching). This lets you run [hybrid queries](/documentation/search/hybrid-queries/) over several representations of the same data within one collection. Each vector must be defined in the collection's schema.
```
timvisee: I don't like this wording, but I think it's worth it to emphasize we don't support 'dynamic' vectors.

```suggestion
A point can hold any number of dense, sparse, and multi vectors, though each has to be configured in the collection's schema. There's no hard limit imposed by Qdrant, though practical limits apply: each additional vector increases memory usage, so the realistic ceiling is determined by available RAM and storage. You can attach a single vector, or multiple vectors with different names (for example, a dense vector for semantic search alongside a sparse vector for keyword matching). This lets you run [hybrid queries](/documentation/search/hybrid-queries/) over several representations of the same data within one collection. Each vector must be defined in the collection's schema.
```

qdrant-landing/content/documentation/faq/qdrant-fundamentals.md
```diff
@@ -55,6 +55,43 @@ There are two possible reasons for this:
...
+
+Yes, in most Retrieval-Augmented Generation (RAG) setups, each chunk is stored as a separate point. The chunk text (or a reference to it) goes in the payload, and the embedding of that chunk is the vector. Points can share a `document_id` payload field so you can trace results back to the source document.
+
+How you chunk matters significantly and is domain-dependent. As a starting point: paragraph-level chunking works well for books and prose; sentence-level chunking tends to work better for technical articles and Q\&A content. Plan to experiment — chunking strategy is one of the highest-leverage variables in retrieval quality.
+
+See also: [Text Chunking Strategies](course/essentials/day-1/chunking-strategies/)
+
+### How does point deletion work internally? Does Qdrant rebuild the index on every delete?
+
+Qdrant implements deletions as soft deletes using a bitmask, so the index is not rebuilt after each deletion. The bitmask is a lightweight data structure kept in RAM, enabling Qdrant to quickly determine whether a point should be excluded from an operation without accessing the deleted point's data. The Vacuum Optimizer handles physical cleanup in the background.
```
timvisee:
```suggestion
Qdrant implements deletions as soft deletes using a bitmask, so the index is not rebuilt after each deletion. The bitmask is a lightweight data structure, enabling Qdrant to quickly determine whether a point should be excluded from an operation without accessing the deleted point's data. The Vacuum Optimizer handles physical cleanup in the background.
```

timvisee (approved):

PR #2326, opened 2026-05-04
add referrer policy override for hosts listed in referrerFullDomains

trean:


trean:
Close in favor #2340 2340

PR #2334, opened 2026-05-05
Multi-representation search tutorial + supporting doc cross-links

Dylancouzon:
Tutorial URL: https://deploy-preview-2334--condescending-goldwasser-91acf0.netlify.app/documentation/tutorials-search-engineering/multi-representation-search/ 

## Summary

  - Adds a new Search Engineering tutorial covering multi-representation retrieval (title, summary, chunks, and tags) end to end, with a recommended pipeline (three named-vector prefetches + RRF + grouping) plus a formula-based boosting alternative.
  - Wires three supporting cross-links into existing reference docs so readers reach the tutorial from the relevant entry points.
  - Adds the new tutorial to both the Search Engineering tutorial index and the cross-section tutorials landing page.

  ## Scope

  Closes a content gap raised repeatedly in Discord and customer interviews: how to combine multiple representations of the same item (title vs. chunks vs. document) under one Query API call. Existing content covers dense + sparse hybrid; this tutorial adds the multi-representation pattern, document-level grouping, and formula-based score boosting as a single design progression. The Vectors page tweak also addresses the recurring multivector misuse pattern (storing unrelated representations in a multivector field, where MaxSim collapses the per-representation signal).

  ## Files
  - `qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md` — new tutorial.
  - `qdrant-landing/content/documentation/headless/content/tutorials/search-engineering.md` — sidebar entry, ordered above Static Embeddings.
  - `qdrant-landing/content/documentation/tutorials-lp-overview.md` — tutorial-landing-page entry.
  - `qdrant-landing/content/documentation/search/hybrid-queries.md` — "See also" link at the end of the Grouping section.
  - `qdrant-landing/content/documentation/manage-data/vectors.md` — clarifies in the Multivectors subsection that MaxSim returns one combined score per point and points to Named Vectors + the new tutorial.
  - `qdrant-landing/content/documentation/search/text-search.md` — short-field BM25 calibration note plus BM25F workaround pointer at the end of the BM25 parameters subsection.
  - `qdrant-landing/static/documentation/tutorials/multi-representation-search/point.png` — collection-point screenshot used in the tutorial's schema section.

  ## Notebook
  The accompanying Jupyter notebook builds the same pipeline step by step against a 2,000-paper ML/CS arXiv slice and prints qualitative top-K results at each step. It will land in a separate PR against `qdrant/examples` at `multi-representation-search/multi-representation-search.ipynb`. The header table's GitHub and Open in Colab links and the body's "accompanying notebook" reference both point at that path, so they resolve as soon as the notebook PR is merged.
  PR: https://github.com/qdrant/examples/pull/103


kanungle:
Can you put the destination preview url in the top comment so we can find what to review easier? Thanks!

Dylancouzon:
@kanungle https://deploy-preview-2334--condescending-goldwasser-91acf0.netlify.app/documentation/tutorials-search-engineering/multi-representation-search/

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
```diff
@@ -0,0 +1,270 @@
...
+| --- | ----------- | ----------- | ----------- |
+
+A document is rarely well-represented by a single embedding. A paper has a title, an abstract, body chunks, and category tags, each carrying a different signal. Treat all four as one dense vector and the title gets averaged out, keyword matches on tags disappear, and chunk-level grounding for downstream reasoning is gone.
+
+This tutorial builds retrieval that uses each representation deliberately: named vectors per representation, fused via the Query API, grouped back to the document level for presentation, with score boosting for ranking preferences. Each step targets a specific retrieval failure mode you'd hit in production.
+
+This tutorial assumes you've built hybrid (dense plus sparse) search before, that you're comfortable with [named vectors](/documentation/manage-data/vectors/#named-vectors), the [Query API](/documentation/search/hybrid-queries/), Reciprocal Rank Fusion (RRF), and Best Matching 25 (BM25). If hybrid search is new, start with [Hybrid Search with FastEmbed](/documentation/tutorials-search-engineering/hybrid-search-fastembed/) first.
```
mrscoopers: > If hybrid search is new, start with [Hybrid Search with FastEmbed](/documentation/tutorials-search-engineering/hybrid-search-fastembed/) first.

Better point to docs around hybrid search/Qdrant courses, FastEmbed is a satellite of Qdrant

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
```diff
@@ -0,0 +1,270 @@
...
+
+```bash
+pip install qdrant-client fastembed datasets
+```
+
+Use Python <3.13. Not all dependencies support the newest Python versions yet.
+
+## Dataset
+
+You'll work with 20 000 arXiv papers from the [`gfissore/arxiv-abstracts-2021`](https://huggingface.co/datasets/gfissore/arxiv-abstracts-2021) Hugging Face dataset, filtered to ML/CS categories and to papers from 2018 onward — earlier ML papers predate most of the topics queries care about. Each paper has a title, an abstract, and category tags, which gives you four natural representations once the abstract is split into chunks: title, full abstract as a summary, abstract sentences as chunks, and categories as tags.
```
mrscoopers: > which gives you four natural representations once the abstract is split into chunks

Why? Not fitting the context window?
Or we're splitting for the sake of showing an example of chunking vs full length, aka in reality it'd rather be full paper texts, but for simplicity we take abstracts?

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
```diff
@@ -0,0 +1,270 @@
...
+        "arxiv_id": arxiv_id,
+        "title": row["title"].strip(),
+        "abstract": row["abstract"].strip(),
+        "categories": cats,
+    })
+```
+
+## Collection Schema
+
+Design the collection before writing any queries. The point granularity is the chunk: every chunk of every paper becomes one point. Title and summary embeddings are stored on every chunk so the Query API can fuse across them in a single request without an extra lookup.
```
mrscoopers: > Title and summary embeddings are stored on every chunk so the Query API can fuse across them in a single request without an extra lookup.

Here important t show alternative with lookups in another collection (at least reference docs) + briefly mention pluses/minuses of each approach

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
```diff
@@ -0,0 +1,270 @@
...
+
+```python
+from qdrant_client import QdrantClient, models
+
+client = QdrantClient("http://localhost:6333") # or QdrantClient(url="https://<id>.cloud.qdrant.io", api_key="...") for Qdrant Cloud
+
+client.create_collection(
+    collection_name="arxiv_multi_repr",
+    vectors_config={
+        "dense_chunk":   models.VectorParams(size=384, distance=models.Distance.COSINE),
```
mrscoopers, line 91: 384 looks like a mgic number, needs 1-line comment where it comes from (we're using all-mini-lm6 for simplicity/demo purposes)

I'd use qdrant Free Tier for Tutorials insteas of localhost, all-mini-lm6 is on Free Tier Free Inference
It allows to drop fastembed from dependencies alltogether & simplifies ingestion & embeddings for users

cc @abdonpijpelink right?

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
```diff
@@ -0,0 +1,270 @@
...
+        ),
+    },
+)
+```
+
+Each vector covers a different signal:
+
+- `dense_chunk`: the content workhorse. Chunks are short enough that a single embedding represents them faithfully.
+- `dense_title`: a few tokens that name the topic. A title hit is a strong signal even when no chunk matches.
+- `dense_summary`: between title and chunk in length and specificity. Catches queries about the contribution rather than a passage.
```
mrscoopers: Can we super quickly explain why summary & not full abstract (aka modelling context window of embedding model limit), and why summary is not a bad idea even though it loses some information & how summary can be obtained (LLM-based, extractive, ...?)
Why important to explain -- for bm25, for example, it makes sense to use full text rather than a summary (as it was initially designed for long text search), hence we need to motivate recommended choices

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
```diff
@@ -0,0 +1,270 @@
...
+- `sparse_keywords`: BM25 over title and tags concatenated. BM25 pays off on short structured fields where exact lexical matches matter.
+
+Title and summary vectors are duplicated across every chunk of the same document. That trades storage for query simplicity: one collection, one Query API call, no `lookup_from`. If your titles or summaries are heavy, or you have millions of chunks per document, store them in a separate collection and reach them with `lookup_from`. For the typical case (a few dozen chunks per document, embeddings under a kilobyte each), inline storage is the simpler choice.
```
mrscoopers: > no `lookup_from`.

I don't think people know what that refers to, it's very unknown & undocumented feature, sadly

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
mrscoopers: I assume explanations are further in the tutorial, I'd ust move them up, so it reads linearly

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
```diff
@@ -0,0 +1,270 @@
...
+
+Ingestion produces one point per chunk and reuses the title and summary embeddings.
+
+```python
+from fastembed import TextEmbedding, SparseTextEmbedding
+
+# Dense embeddings for title, summary, and chunk content; sparse BM25 for keyword matching.
```
mrscoopers: > sparse BM25 for keyword matching

on which field? that'd be important to explain early

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
```diff
@@ -0,0 +1,270 @@
...
+dense_model = TextEmbedding("BAAI/bge-small-en-v1.5")
+sparse_model = SparseTextEmbedding("Qdrant/bm25")
+
+def chunk_sentences(text, target_len=2):
+    """Split text into ~2-sentence chunks; fall back to the full text if it doesn't split cleanly."""
+    sentences = [s.strip() for s in text.split(". ") if s.strip()]
+    return [". ".join(sentences[i:i + target_len])
+            for i in range(0, len(sentences), target_len)] or [text]
+
+def to_sparse(sparse_emb):
```
mrscoopers: BM25 has been in Qdrant's core since 1.15
That means you don't need fastembed as a proxy, you just send texts to Qdrant & it converts inside it to bm25 vectors
We need to use core BM25, so people move off to it
Examples are in the Text Search guide

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
```diff
@@ -0,0 +1,270 @@
...
+    return models.SparseVector(
+        indices=sparse_emb.indices.tolist(),
+        values=sparse_emb.values.tolist(),
+    )
+
+points = []
+for paper in papers:
+    chunks = chunk_sentences(paper["abstract"])
+
+    # Paper-level embeddings: computed once per paper, reused across every chunk below.
```
mrscoopers: > paper

paper abstract (just for clarity that we don't take full texts)

mrscoopers: but also I am confused, it says summary
I thought we summarize abstract somehow

does it mean that we're just calling sbtract a summary?

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
```diff
@@ -0,0 +1,270 @@
...
+    # next(iter(...)) extracts the single vector from FastEmbed's generator output.
+    title_vec   = next(iter(dense_model.embed([paper["title"]]))).tolist()
+    summary_vec = next(iter(dense_model.embed([paper["abstract"]]))).tolist()
+    sparse_vec  = to_sparse(next(iter(sparse_model.embed(
```
mrscoopers: Why specifically these fields for bm25 vecs?

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
```diff
@@ -0,0 +1,270 @@
...
+        [paper["title"] + " " + " ".join(paper["categories"])]
+    ))))
+
+    # Chunk-level dense embedding: one vector per chunk.
+    chunk_vecs = [v.tolist() for v in dense_model.embed(chunks)]
+
+    # One Qdrant point per chunk. dense_title, dense_summary, and sparse_keywords
```
mrscoopers: Tbf I start thinking it'd be rather nice to demo in this tutorial "with lookup."
At least briefly, as an alternative

Then we could show how to move off all of the vectors also to a separate collection, as here we take a lot of storage on copying title_vec, summary_vec, sparse_vec for each abstract to all the chunks, and vectors are actually very heavy, especially in production

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
```diff
@@ -0,0 +1,270 @@
...
+        ))
+
+client.upload_points(collection_name="arxiv_multi_repr", points=points, batch_size=64)
+```
+
+After the upload completes, opening any point in the Qdrant Cloud UI shows all four named vectors attached to one chunk. `dense_chunk` carries the chunk's own embedding, while `dense_title`, `dense_summary`, and `sparse_keywords` are the same across every chunk of this paper.
+
+![A point in the arxiv_multi_repr collection showing all four named vectors](/documentation/tutorials/multi-representation-search/point.png)
+
+A fixed-length sentence chunker is used here for clarity. Chunking strategy is its own design space; see the limitations section at the end for pointers.
```
mrscoopers: Ahso, so it's here
I'd move it up a little bit, closer to chunking code

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
```diff
@@ -0,0 +1,270 @@
...
+
+## Retrieval
+
+The recommended pipeline fuses three prefetches with Reciprocal Rank Fusion and groups the results by document. One Query API call covers everything except boosting:
```
mrscoopers: > except boosting:

didn't understand this
reads like Query API can't have boosting as a building block
Also I am not sure people know what boosting is in Qdrant terms

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
```diff
@@ -0,0 +1,270 @@
...
+def retrieve(query, limit=10, group_size=3):
+    dense, sparse = embed_query(query)
+    return client.query_points_groups(
+        collection_name="arxiv_multi_repr",
+        prefetch=[
+            models.Prefetch(query=dense,  using="dense_chunk",     limit=100),
+            models.Prefetch(query=dense,  using="dense_title",     limit=100),
+            models.Prefetch(query=sparse, using="sparse_keywords", limit=100),
+        ],
+        query=models.FusionQuery(fusion=models.Fusion.RRF),
```
mrscoopers, line 198: tbf with this big amount of documents in prefetch, we could even advise DBSF? Have you empirically checked which fusion works better?
I'd link here FAQ on "which fusion to choose" that @abdonpijpelink recently merged

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
```diff
@@ -0,0 +1,270 @@
...
+        group_by="document_id",
+        group_size=group_size,
+        limit=limit,
+    ).groups
+```
+
+BM25 has a separate `query_embed` path that uses inverse document frequency (IDF) weighting calibrated for queries, distinct from the indexing-side `embed`.
```
mrscoopers: > BM25 has a separate `query_embed` path that uses inverse document frequency (IDF) weighting calibrated for queries, distinct from the indexing-side `embed`.

I think this is complicated
I didn't fully understand
But IDF weighting is done on Qdrant side, it's not dependent on embed at all

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
```diff
@@ -0,0 +1,270 @@
...
+
+The pipeline makes four design decisions. Each is worth understanding so you can defend or adjust it.
+
+### What to Prefetch
+
+A representation only earns its own prefetch if it carries signal independent of the others. The three above are deliberately complementary:
+
+- `dense_chunk` carries the body content.
+- `dense_title` carries the topical naming. For a query like "diffusion models for high-resolution image synthesis", a paper titled "High-Resolution Image Synthesis with Latent Diffusion Models" surfaces from the title prefetch even when its abstract phrases the contribution differently. The chunk prefetch alone misses it.
+- `sparse_keywords` carries lexical hits on title and tags that the dense embedding has averaged out: rare entity names, jargon, controlled-vocabulary tags.
```
mrscoopers: I still think that categories belong better in filters (NER with LLM could prefill them) & long texts in BM25
wdyt?

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
```diff
@@ -0,0 +1,270 @@
...
+
+Adding `dense_summary` as a fourth prefetch is worth it only if summaries surface what chunks don't. If summaries paraphrase the chunks, the extra prefetch adds latency without lift.
```
mrscoopers: >only if summaries surface what chunks don't

what do we recommend for checking this hypothesis? why did we recommend then to embed & store summaries, which is heavy resource-wise?

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
```diff
@@ -0,0 +1,270 @@
...
+
+### How to Fuse
+
+[Reciprocal Rank Fusion](/documentation/search/hybrid-queries/#reciprocal-rank-fusion-rrf) works on rank, not score, so it sidesteps the calibration problem entirely: dense scores live in [0, 1], sparse BM25 scores don't, and RRF doesn't have to reconcile the two. The prefetches return overlapping but not identical candidate sets, and RRF rewards documents the paths agree on.
```
mrscoopers: > reconcile the two

We have more parallel prefetches than 2 in the example above, i'd adapt this section to our specific use case

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
```diff
@@ -0,0 +1,270 @@
...
+
+Tuning linear weights between dense and sparse scores instead of using RRF is fragile: the right weight depends on query length, model, and corpus, and a weight that helps one query class hurts another. The full argument is in [Hybrid Search Revamped](/articles/hybrid-search/). Stick with RRF unless you have a reason and a holdout set to validate the alternative.
```
mrscoopers: > Tuning linear weights between dense and sparse scores instead of using RRF is fragile

I think instead we can mention that one can uses weighted RRF API, but the right weights shouldn't be based on vibe, but on collection, queries and use case

> Stick with RRF unless you have a reason and a holdout set to validate the alternative.

I am not sure we want to keep this, RRF is a baseline, but out-of-the-box it rarely works magically good

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
```diff
@@ -0,0 +1,270 @@
...
+
+### When to Group, When Not To
+
+`query_points_groups` collapses chunks back to documents while keeping the top chunks per document attached. Each group's `hits` field carries the top-`group_size` chunks for that paper, ranked by their fused score.
```
mrscoopers: > query_points_groups

I think people reading the guide have no knowledge about this API, and here it jumps right at them as a keyword

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
```diff
@@ -0,0 +1,270 @@
...
+
+Grouping helps when the consumer wants documents: a results UI, a citation list, an LLM that wants document-level attribution. It hurts when an LLM benefits from seeing several independently ranked chunks across multiple documents in its context window. Collapsing those chunks into one group per document throws away ordering information the LLM could have used.
```
mrscoopers: > It hurts when an LLM benefits from seeing several independently ranked chunks across multiple documents in its context window.

In which cases that happens?

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
```diff
@@ -0,0 +1,270 @@
...
+
+Grouping is a presentation choice, not a relevance technique. The candidates and their fused scores don't change; only the result shape does.
+
+Two operational notes:
+
+- Increase prefetch `limit` when grouping. If a paper has three chunks worth retrieving but the prefetch only returned two, grouping doesn't have the third to consider.
+- Use the `with_lookup` parameter when document-level metadata (full title, authors, publication date) lives in a separate collection. It fetches one record per group instead of repeating it per chunk.
```
mrscoopers: We definitely need to show how
Nothing in our docs shows it properly

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
```diff
@@ -0,0 +1,270 @@
...
+
+### When to Boost, When to Rerank
```
mrscoopers: I'd rename it to "other types of fusion" for discoverability and move closer to RRF

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
```diff
@@ -0,0 +1,270 @@
...
+    formula=models.SumExpression(sum=[
+        "$score[0]",
+        models.MultExpression(mult=[0.5, "$score[1]"]),
+        models.MultExpression(mult=[0.3, "$score[2]"]),
+    ]),
+    defaults={"$score[1]": 0.0, "$score[2]": 0.0},
+),
+```
+
+`$score[i]` references the score from prefetch `i`, so order in the `prefetch=` list is load-bearing. The `defaults` map covers candidates that appeared in one prefetch but not another: without it, a missing variable would error. The formula above sums the chunk score with a half-weighted title score and a smaller sparse contribution. Unlike RRF, this is a linear combination of raw scores and is fragile across query types unless you've held the weights up against representative queries.
```
mrscoopers: Aha, so here it makes sense to explain why linear fusion is rarely a good approach & reference explanation

mrscoopers: And tell that if you want a custom weighted fusion, you need to align distributions of prefetches into one range (what RRF kind of does by throwing away scores & DBSF kind of does by normalizing scores per prefetch). For that, you could use decay functions, for example,
https://medium.com/qdrant/untangling-relevance-score-boosting-in-vector-search-ccffb4a574d0 here in "I Don’t Know All the Parameters in Advance," is a small section about it

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
```diff
@@ -0,0 +1,270 @@
...
+
+---
+
+For the step-by-step build-up that produced this design (dense baseline, plus sparse, plus title prefetch, plus grouping, plus boosting) with eval numbers showing the lift at each step, see the [accompanying notebook](https://github.com/qdrant/examples/blob/master/multi-representation-search/multi-representation-search.ipynb). For a complementary view that varies the *query* across three representations against a single document representation, see the [Universal Query for Hybrid Retrieval demo](/course/essentials/day-5/universal-query-demo/). This tutorial is the document-side equivalent: multiple document representations against a single query.
+
+## Where This Pattern Doesn't Fit
+
+Multi-representation retrieval pays off when items are long and structured. It's overkill, and sometimes worse than a single dense vector, when items are short and homogeneous. Tweets, product names, and one-line forum titles don't have a meaningful title-versus-summary-versus-body distinction; splitting them into representations adds query latency without adding signal. For those cases, a single well-chosen dense model and a sparse fallback are usually enough.
```
mrscoopers: Tbf this reads as obvious
I don't think people ever split unless it's a bad-bad need (aka very long doc)

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
```diff
@@ -0,0 +1,270 @@
...
+
+The pattern also assumes you can identify the representations cleanly. If your corpus has inconsistent metadata (some documents have summaries, others don't), the missing-representation case becomes its own design problem: empty vectors, fallback strategies, or a separate index per representation availability.
```
mrscoopers: > some documents have summaries,

What do we refer to as a summary?
abstract? cause I assume people can generate a summary when indexing

> If your corpus has inconsistent metadata

I am afraid this is literally 99% of reality use cases & they still need RAG on multirepresentations, so we'd rather give some advice on how to approach it

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
```diff
@@ -0,0 +1,270 @@
...
+
+## Open Ends
+
+**Chunking strategy.** This tutorial uses a fixed-length sentence chunker for clarity. The chunking choice has a measurable effect on retrieval quality, and the right strategy depends on document structure. Worth considering for your corpus: hierarchical chunking ([POMA-AI VST](https://arxiv.org/abs/2406.04590)), late chunking ([Jina](https://jina.ai/news/late-chunking-in-long-context-embedding-models/)), and semantic chunking ([Chonkie](https://github.com/chonkie-ai/chonkie)).
+
+**BM25F.** The technically correct extension of BM25 to multi-field text of varying length is BM25F, which weights term statistics per field. Qdrant doesn't support BM25F natively today. The workaround used in step 3, separate sparse vectors per field fused via the Query API, gives you most of the practical benefit. Default BM25 parameters (k1, b) are calibrated for document-length text and may need recalibration for short fields like titles or tags; treat them as tunable hyperparameters when working with mixed-length sparse representations.
```
mrscoopers: > Default BM25 parameters (k1, b) are calibrated for document-length text and may need recalibration for short fields like titles or tags;

Much more important is to calibrate avg_len, btw, as Qdrant doesn't calculate it for you automatically, as many other search engines with BM25 do

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
mrscoopers: Also, we need to adjust the avg_len parameter
Qdrant doesn't calculate it for users automatically, as many other search engines with BM25 do

mrscoopers (changes requested):

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
Dylancouzon: Fixed in 300397b — now points to the hybrid search section of the Text Search guide.

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
Dylancouzon: Fixed in 9725689. Switched the client to Cloud Free Tier with `cloud_inference=True`, dropped FastEmbed entirely, and added a one-line comment on where 384 comes from. Self-host path covered by an aside in the Setup section.

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
Dylancouzon: Fixed in 9725689. BM25 is now generated server-side via `models.Document(text=..., model="qdrant/bm25")` at both ingest and query time. FastEmbed sparse model and the `to_sparse` helper are gone.

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
Dylancouzon: Removed in 9725689. The paragraph was specific to FastEmbed's separate `query_embed` path; with core BM25, queries just pass text and IDF weighting is handled server-side, so the distinction no longer applies.

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
Dylancouzon: Fixed in 433cae6. Added a short paragraph to the Dataset section explaining that abstracts fit any embedding model's context window — chunking is there to keep the pipeline shape identical to the production case on full paper bodies. Your second read ("in reality it'd rather be full paper texts, but for simplicity we take abstracts") was correct.

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
Dylancouzon: Fixed in 433cae6. Renamed `dense_summary` to `dense_abstract` and removed every "summary" reference — we were just calling the abstract a summary, which is the confusion you flagged. The new Dataset paragraph also explains why we work with the abstract directly rather than a generated summary.

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
Dylancouzon: Fixed in 433cae6. The schema and prose now consistently say "abstract" everywhere, and the Dataset section says explicitly that the abstract is standing in for what would be longer body text in your own corpus.

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
Dylancouzon: Fixed in 376202a. After re-reading the docs I realized `lookup_from` is for querying by ID across collections, not for splitting representation storage, so the original reference was misleading. The paragraph now points readers to `with_lookup` (Lookup in groups) for the payload-split case, which is the cleanest cross-collection pattern Qdrant supports natively.

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
Dylancouzon: Fixed in 376202a. Removed the `lookup_from` reference. It is a feature for using a vector from another collection as the query, not for storing representations across collections, so it didn't belong in that paragraph.

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
Dylancouzon: Approached in 376202a. Splitting vectors across collections in Qdrant requires running separate `query_points` calls and fusing client-side, so it is a separate design pattern from what this tutorial covers and we kept it out. For the payload-split case, the paragraph at line 107 now links to the Lookup in groups doc, which shows the chunks-collection plus documents-collection pattern using `with_lookup`.

Dylancouzon: @mrscoopers let me know if this is a good enough solution or if you'd still rather have the alternative solution. My concern is that I don't want to make the tutorial more complex than it needs to be.

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
Dylancouzon: Fixed in 68cec30. Set `avg_len` explicitly on the BM25 Document at ingest with a comment noting that the default 256 is calibrated for document-length text and would skew scoring for short fields like title + categories.

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
Dylancouzon: Fixed in 68cec30. Set `avg_len` in the ingestion code and trimmed the parameter calibration sentence from this note. The BM25 Parameters subsection in the Text Search guide already covers `k`, `b`, and `avg_len` with a working example, so keeping this note brief avoids duplication.

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
Dylancouzon: Fixed in a9610d2. Added DBSF as one of three named alternatives in the restructured fusion section, and linked the RRF vs. DBSF FAQ entry at the bottom. No empirical comparison in the tutorial itself; we framed DBSF as worth trying once you have an eval set rather than as a blanket recommendation.

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
Dylancouzon: Fixed in a9610d2. Dropped the `Stick with RRF` line. RRF is now framed as a reasonable starting point that the alternatives often beat once you have an eval set. Weighted RRF is one of the listed alternatives, with an explicit note that weights should come from validation runs, not intuition.

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
Dylancouzon: Fixed in a9610d2. Renamed the section to `Which Fusion to Use` and moved the FormulaQuery code block from the boosting section into it. The fusion section now covers RRF (default), Weighted RRF, DBSF, and custom formulas as four named options.

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
Dylancouzon: Fixed in a9610d2. Added the distribution-alignment explanation in the custom-formula paragraph, with concrete examples of how RRF and DBSF handle it automatically and what a custom formula has to do itself. Linked the in-repo Decay Functions and Score Boosting references rather than the Medium article, since the equivalent content lives in search-relevance.md.

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
Dylancouzon: ingestion switched to Cloud Inference

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
Dylancouzon: Fixed in a9610d2. The intro line now reads "One Query API call covers retrieval, fusion, and grouping" instead of "everything except boosting." The restructured sections below also split formula-based fusion from true score boosting, so the term is no longer ambiguous.

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
Dylancouzon: Fixed in a9610d2. The RRF paragraph was rewritten in the fusion-section restructure; "reconcile the two" is gone, and the new phrasing ("combines prefetches by document rank, ignoring raw scores") covers any number of prefetches.

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
Dylancouzon: Very fair. Will rewrite

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
Dylancouzon: Fixed in 8abae12. Folded the chunking note into the Dataset paragraph that already explains why we chunk. Also dropped the separate Open Ends section that listed alternative chunking methods, since the consolidated paragraph covers what readers need.

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
Dylancouzon: Moved categories to filterable payload (keyword index on `tags`), and `retrieve()` now takes an optional `tags` parameter that builds a `query_filter` when set. BM25 covers title only. Fixed in 8ea4653.

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
Dylancouzon: Tried to keep the tutorial focused on the single-collection pattern, so I pointed at [Lookup in groups](https://qdrant.tech/documentation/search/search/#lookup-in-groups) instead of walking through `with_lookup` here. Think that pointer is enough, or want me to beef up the reference docs in a separate PR? Happy either way.

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
Dylancouzon: Added a one-line comment above the `query_points_groups` call summarizing what it does (runs prefetches, fuses with RRF, applies the filter, groups by document_id), so readers see it inline before getting to the When to Group subsection. Fixed in 479d305.

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
Dylancouzon: Agreed, dropped the paragraph. The whole 'Where This Pattern Doesn't Fit' section is gone in 479d305, since the second paragraph on inconsistent metadata wasn't pulling its weight either.

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
Dylancouzon: Removed the whole 'Where This Pattern Doesn't Fit' section in 479d305 rather than try to make the inconsistent-metadata advice actionable in a paragraph. Inconsistent metadata is its own design problem and probably deserves its own write-up rather than a hedge here.

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
Dylancouzon: Removed the hedge paragraph and added `dense_abstract` as a fourth prefetch in the recommended pipeline instead. We already ingest it on every chunk, so the pipeline may as well use it; in production (where chunks would be full-paper bodies), the abstract gives a distinct paper-level view. Fixed in b301408.

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
Dylancouzon: Removed the entire 'When to Group' section. The vague 'it hurts when an LLM benefits from chunks across documents' sentence wasn't pulling its weight, and the useful operational bits (prefetch limit, query_points_groups intro) fold into code comments on retrieve(). Fixed in b301408.

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
Dylancouzon: > I am afraid this is literally 99% of reality use cases & they still need RAG on multirepresentations, so we'd rather give some advice on how to approach it

I deleted the section without really addressing concrete strategies for inconsistent metadata anywhere else. Should we add a short practical paragraph here, or would you rather move it into a separate guide where we can ground each approach with real examples and tradeoffs? And in either case, what do you want this section to look like?

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
Dylancouzon: Closing, duplicated topic with https://github.com/qdrant/landing_page/pull/2334#discussion_r3207461263

Dylancouzon:
@mrscoopers this is ready for another review pass whenever you have time. I applied the feedback, simplified things across the board, and cut sections that didn’t feel dense enough. I still have 2 open questions:

L223: I kept the tutorial focused on the single-collection pattern and linked to the "Lookup in groups" docs instead of adding a with_lookup example inline. Is that enough, or should I still add the alternative here?

L254: I removed the hedge around inconsistent metadata, but I didn’t add practical guidance back in. Should I add a short paragraph here, or make a separate guide with concrete examples and tradeoffs?

Also: the accompanying notebook (https://github.com/qdrant/examples/pull/103) is referenced from this tutorial, so it'll need to land before this PR merges.

tysmly

mrscoopers:
> L223: I kept the tutorial focused on the single-collection pattern and linked to the "Lookup in groups" docs instead of adding a with_lookup example inline. Is that enough, or should I still add the alternative here?

To be honest, it'd be truly good to add an alternative, as this one feature is very, VERY undocumented [like we're linking docs with a minimal toy example, reads like "eh, go figure it out"]. So there are 2 options: we never needed this feature from the product side in the 1st place, OR no one can start widely using it because we never provided a good example of "when" & "how". I suspect the 2nd.
Do you see a **minimal sensible** way to show how to use this lookup from another collection?

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
mrscoopers: > would you rather move it into a separate guide where we can ground each approach with real examples and tradeoffs?

Well, we could do the same v1 -> v_pitch_perfect approach, aka just leave CTA in v1 "and in case your metadata in incosnistent & you don't know what to do, come to our Discord & let's brainstorm together."

mrscoopers:
> L254: I removed the hedge around inconsistent metadata, but I didn’t add practical guidance back in. Should I add a short paragraph here, or make a separate guide with concrete examples and tradeoffs?

answered in thread around this paragraph, I propose to just leave a short CTA for the v1

mrscoopers:
> Also: the accompanying notebook (https://github.com/qdrant/examples/pull/103) is referenced from this tutorial, so it'll need to land before this PR merges.

As long as the notebook's code is the same as the code in the markdown of this tutorial & it works end-to-end, consider it approved:)

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
```diff
@@ -0,0 +1,261 @@
...
+
+# Index 'document_id' so the Query API can group by it; index 'tags' so we can filter on category.
+client.create_payload_index(
+    collection_name="arxiv_multi_repr",
+    field_name="document_id",
+    field_schema=models.PayloadSchemaType.KEYWORD,
+)
+client.create_payload_index(
+    collection_name="arxiv_multi_repr",
+    field_name="tags",
```
mrscoopers, line 110: Which tags? Don't see them in the ingestion code above

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
```diff
@@ -0,0 +1,261 @@
...
+    field_schema=models.PayloadSchemaType.KEYWORD,
+)
+```
+
+Each vector covers a different signal: `dense_chunk` for chunk content, `dense_title` and `sparse_title` for the title (dense and lexical respectively), `dense_abstract` for the abstract as a whole.
+
+Categories live in the `tags` payload with a keyword index, so queries can pre-filter by category.
```
mrscoopers: I'd link indexing, I think some might have no idea why keyword

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
```diff
@@ -0,0 +1,261 @@
...
+
+Title and abstract vectors are duplicated across every chunk of the same document. That trades storage for query simplicity: one collection, one Query API call, every representation reachable from any point. For the typical case (a few dozen chunks per document, embeddings under a kilobyte each), inline storage is the simpler choice. To split heavy document-level payload into a separate collection and pull it back at grouping time, see [Lookup in groups](/documentation/search/search/#lookup-in-groups).
```
mrscoopers: > For the typical case (a few dozen chunks per document, embeddings under a kilobyte each), inline storage is the simpler choice.

I'd rather provide numbers (tech people how to calculate them) instead of judgment, f.e
A few dozen chunks = 24 per doc, meaning you repeat 24 times the same abstract vector and the same title vector.
all-mini-lm6 at full precision is (4 * 384) / 1024 = 1.5kb
Hence, per doc we store 24 * 1.5 * 2 = 72 kb data (if we stored it externally, it'd be 24 times more, aka 3kb)
For 1 million docs, the difference is 69 GB vs 3 GB only in embeddings (I am not counting in payloads or sparse vectors). Here, the question is, how users plan to use these embeddings (search or rerank), and if they plan to store them on disk or in RAM (it's very expensive for RAM), and how fetching from another collection affects overall latency compared to the one Query API request but with vectors pulled from disk <_that I don't even know, needs research_>.

mrscoopers (changes requested):

Dylancouzon:
> Do you see a minimal sensible way to show how to use this lookup from another collection?

@mrscoopers Not without adding quite a bit of complexity and length to the tutorial.
The reference needs a real `with_lookup` example. The current one is too sparse to teach the pattern. I'd put it on the Lookup in groups page itself, not inline here: developers searching for the feature later won't dig through a multi-representation tutorial to find it. Follow-up 
Happy to own the follow-up PR, then update this tutorial with the reference, okay with that?

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
Dylancouzon: Damn, great catch! The key name was probably edited when migrating the tags from Sparse to filters.

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
Dylancouzon: Fixed in d355db9. Linked `keyword index` to the Payload Index reference.

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
Dylancouzon: Fixed in d355db9. Swapped the judgment for headline numbers (1.4 GB inline vs 60 MB sidecar at 20 000 × 24) and pointed at Calculating RAM size so readers can price it on their own corpus.

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
Dylancouzon: Added a short Discord CTA in Wrapping Up in d355db9.

mrscoopers:
> > Do you see a minimal sensible way to show how to use this lookup from another collection?
> 
> @mrscoopers Not without adding quite a bit of complexity and length to the tutorial. The reference needs a real `with_lookup` example. The current one is too sparse to teach the pattern. I'd put it on the Lookup in groups page itself, not inline here: developers searching for the feature later won't dig through a multi-representation tutorial to find it. Follow-up Happy to own the follow-up PR, then update this tutorial with the reference, okay with that?

Agree

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
```diff
@@ -0,0 +1,263 @@
...
+        limit=limit,
+    ).groups
+```
+
+### What to Prefetch
+
+A representation only earns its own prefetch if it carries signal independent of the others. The four above are deliberately complementary:
+
+- `dense_chunk` carries the body content.
+- `dense_title` carries the topical naming. For a query like "diffusion models for high-resolution image synthesis", a paper titled "High-Resolution Image Synthesis with Latent Diffusion Models" surfaces from the title prefetch even when its abstract phrases the contribution differently. The chunk prefetch alone misses it.
```
mrscoopers: > even when its abstract phrases the contribution differently

Don't we assume that vector search should be good with paraphrasing? Aka, should still fetch the right abstract if it phrases the topic differently? That rather sounds like an argument for `sparse_title`(?)

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
```diff
@@ -0,0 +1,263 @@
...
+        models.MultExpression(mult=[0.4, "$score[2]"]),
+        models.MultExpression(mult=[0.3, "$score[3]"]),
+    ]),
+    defaults={"$score[1]": 0.0, "$score[2]": 0.0, "$score[3]": 0.0},
+),
+```
+
+In this formula, `$score[i]` is the score from prefetch `i`, so the order of your `prefetch=` list matters. The `defaults` map provides fallback values (here `0.0`) for candidates that didn't appear in every prefetch, so the formula still evaluates.
+
+This is a linear combination of raw scores, which breaks down when prefetches use different scoring scales (for example, dense scores in [0, 1] alongside unbounded BM25 scores). The other two fusion strategies handle this for you: RRF discards scores entirely, and DBSF normalizes each prefetch before summing. With a custom formula, you have to normalize the scores yourself, typically using [decay functions](/documentation/search/search-relevance/#decay-functions). The full FormulaQuery syntax lives in the [Score Boosting](/documentation/search/search-relevance/#score-boosting) reference.
```
mrscoopers: Can we please make "breaks down" more alert-style, aka more noticeable?
we can reference for clarity articles/hybrid-search/#why-not-a-linear-combination

mrscoopers (approved):

mrscoopers:
@Dylancouzon approved, left 2 lass nitp

Dylancouzon:
Notebook merged in https://github.com/qdrant/examples/pull/103

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
Dylancouzon: Reframed the bullet around field separation: the title isn't in the chunk or abstract text in this schema, so a query matching title-only vocabulary needs this prefetch to surface the document. f68067466

qdrant-landing/content/documentation/tutorials-search-engineering/multi-representation-search.md
Dylancouzon: Promoted to `<aside role="alert">` and linked /articles/hybrid-search/#why-not-a-linear-combination. f68067466

PR #2345, opened 2026-05-11
fix(docs): correct C# MatchExcept snippet (was using Match instead of MatchExcept)

### LINKED ISSUE #1656: Misleading Example MatchExcept for C#
In the section about how to use MatchExcept, if you choose to see how to implement it using C #, the example uses the function Match instead of MatchExcept.

https://qdrant.tech/documentation/concepts/filtering/#match-except

brone1323:
## Summary

Fixes a copy-paste bug in the C# code example for the `MatchExcept` filter condition.

The snippet at `qdrant-landing/content/documentation/headless/snippets/filter-condition/match-except/csharp.cs` was calling `Match("color", ["black", "yellow"])` — identical to the `match-any` example — instead of `MatchExcept("color", ["black", "yellow"])`. This is misleading for developers using the C# client who are trying to implement the NOT IN (MatchExcept) filter.

**Before:**
```csharp
Match("color", ["black", "yellow"]);
```

**After:**
```csharp
MatchExcept("color", ["black", "yellow"]);
```

`MatchExcept` is a valid static method on `Qdrant.Client.Grpc.Conditions` that accepts `(string field, IReadOnlyList<string> keywords)`.

Closes #1656

### DIFF
--- a/qdrant-landing/content/documentation/headless/snippets/filter-condition/match-except/csharp.cs
+++ b/qdrant-landing/content/documentation/headless/snippets/filter-condition/match-except/csharp.cs
@@ -4,6 +4,6 @@ public class Snippet
 {
 	public static async Task Run()
 	{
-		Match("color", ["black", "yellow"]);
+		MatchExcept("color", ["black", "yellow"]);
 	}
 }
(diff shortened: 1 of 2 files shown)


abdonpijpelink (approved):
Thanks for fixing this @brone1323 !

PR #2362, opened 2026-05-19
Add with_lookup example

Dylancouzon:
Preview: https://deploy-preview-2362--condescending-goldwasser-91acf0.netlify.app/documentation/search/search/?q=lookup#lookup-in-groups

## Summary

The Lookup in groups reference now ships a runnable end-to-end example across all seven client languages. Readers can set up the two-collection pattern, ingest, query, and match the documented JSON response without leaving the page.

## What's new

- **Setup snippet** (`query-groups/with-lookup-setup/`): creates `chunks`, the required payload index on `document_id`, and a vectorless `documents` sidecar.
- **Ingest snippet** (`query-groups/with-lookup-ingest/`): upserts documents then chunks.
- **Shorthand Python example** alongside the existing explicit `WithLookup(...)` snippet.
- **Storage trade-off paragraph** quantifying when the split pays off.

## Reference page improvements

- Intro trimmed from 5 paragraphs to 3.
- Constraint paragraph reshaped into a 3-bullet list, moved before the query, and now covers the silent-fail behavior on id type mismatch.
- JSON response example fixed: previously showed chunk id `1` in two groups, which is impossible.
- Shorthand prose corrected: `with_lookup="documents"` does not return looked-up vectors (server default `with_vectors=False`).

## Existing snippet fixes

All six already existed on master and surfaced when running the new walkthrough against the snippet.
| Bug | Fix |
|---|---|
| `with_vectors` comment claimed "true by default" | Corrected to `false` |
| `query=[1.1]` (1-dim) in Python/TS/HTTP vs 4-dim elsewhere | All seven aligned to `[0.2, 0.1, 0.9, 0.7]` |
| `{collection_name}` placeholder in non-Python tabs | All tabs use `"chunks"` |
| Rust `.limit(2u64)` duplicated | Second call now `.group_size(2u64)` |
| Go missing `Limit` field | Added |
| C# legacy `SearchGroupsAsync` | Switched to `QueryGroupsAsync` |

## Verification

- `check.py` compiles all six language sources across three snippet directories.
- `generate-md.py` regenerated all wrapper markdowns.
- End-to-end run against Qdrant 1.18.0: setup → ingest → snippet query → shorthand executes cleanly, matching the documented JSON.
- Silent-fail-on-type-mismatch verified empirically.
- Engine source cross-checked for `WithLookup` field shape, lookup semantics, and `WithVector::default()`.

### DIFF
--- a/qdrant-landing/content/documentation/search/search.md
+++ b/qdrant-landing/content/documentation/search/search.md
@@ -400,25 +400,46 @@ If the `group_by` field of a point is an array (e.g. `"document_id": ["a", "b"]`
 
 ### Lookup in groups
 
-Having multiple points for parts of the same item often introduces redundancy in the stored data. Which may be fine if the information shared by the points is small, but it can become a problem if the payload is large, because it multiplies the storage space needed to store the points by a factor of the amount of points we have per group.
+When the points in a group share large fields like titles, abstracts, or full document vectors, copying that data onto every point inflates storage and forces you to rewrite every chunk whenever a shared field changes.
 
-One way of optimizing storage when using groups is to store the information shared by the points with the same group id in a single point in another collection. Then, when using the [**groups** API](#grouping-api), add the `with_lookup` parameter to bring the information from those points into each group.
+`with_lookup` solves this. Store the shared data once in a separate collection, then attach it to each group at query time using the [groups API](#grouping-api).
 
 ![Group id matches point id](/docs/lookup_id_linking.png)
 
 <aside role="status">Store only document-level metadata (e.g., titles, abstracts) in the lookup collection, not chunks or duplicated data.</aside>
 
-This has the extra benefit of having a single point to update when the information shared by the points in a group changes.
+Set up the two collections. `chunks` holds one point per chunk with its own vector. `documents` holds one point per document with payload only. The payload index on `document_id` lets `group_by` work on that field.
 
-For example, if you have a collection of documents, you may want to chunk them and store the points for the chunks in a separate collection, making sure that you store the point id from the document it belongs in the payload of the chunk point.
+{{< code-snippet path="/documentation/headless/snippets/query-groups/with-lookup-setup/" >}}
 
-In this case, to bring the information from the documents into the chunks grouped by the document id, you can use the `with_lookup` parameter:
+Ingest both collections, with `documents` populated before any query that uses `with_lookup`.
+
+{{< code-snippet path="/documentation/headless/snippets/query-groups/with-lookup-ingest/" >}}
+
+The lookup is a plain join by point id, not a vector search:
+
+- The `documents` collection must already contain a point whose id matches the `group_by` value from each chunk.
+- The id types must match. A string `group_by` value against integer point ids returns an empty `lookup` field with no error.
+- Group ids without a matching point in `documents` get an empty `lookup` field.
+
+Query the `chunks` collection with `group_by="document_id"` and `with_lookup` pointing at the `documents` collection:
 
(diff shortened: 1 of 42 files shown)


kanungle (approved):
Thanks @Dylancouzon !

qdrant-landing/content/documentation/search/search.md
```diff
@@ -400,25 +400,46 @@ If the `group_by` field of a point is an array (e.g. `"document_id": ["a", "b"]`
...
+```python
+client.query_points_groups(
+    collection_name="chunks",
+    query=[0.2, 0.1, 0.9, 0.7],
+    group_by="document_id",
+    limit=2,
+    group_size=2,
+    with_lookup="documents",  # server defaults: with_payload=True, with_vectors=False
+)
+```
```
abdonpijpelink: How about making this a testable snippet too?

abdonpijpelink (commented):
Thanks for this @Dylancouzon ! These docs needed some love.

Some thoughts:

> JSON response example fixed: previously showed chunk id 1 in two groups, which is impossible.

This is not correct. A chunk can be shared by multiple documents, for example with a payload like `{"document_id": [1, 2]}`. It's nice to show in the example (I think it may be something people do: de-duplicate chunks for boilerplate text shared by many documents), but if you think it's better to remove it, we should mention it in the text.

Unrelated to your changes, but as long as we're updating these docs:

- would you have thoughts about how to improve the diagram? It's not great in its current form.
- I think using the same IDs for the chunks and the documents make the example a bit confusing. Given an ID like `1`, is that from a chunk or a document? How about using IDs like `200`, `201`, ... for the documents, like in the diagram?

qdrant-landing/content/documentation/search/search.md
Dylancouzon: The argument value is in the prose and the explicit form is already in the snippet above, so it's redundant. 
I dropped this code block, looks cleaner that way.

Dylancouzon:
@abdonpijpelink Thanks for the feedback

> This is not correct. A chunk can be shared by multiple documents, for example with a payload like {"document_id": [1, 2]}. It's nice to show in the example (I think it may be something people do: de-duplicate chunks for boilerplate text shared by many documents), but if you think it's better to remove it, we should mention it in the text.

Thanks for the correction, I reverted the JSON example to keep chunk id 1 in both groups + expanded prose definition in [0ead6b8](https://github.com/qdrant/landing_page/pull/2362/commits/0ead6b8f8b7e1188da00d7e23f67baf872684599)

> would you have thoughts about how to improve the diagram? It's not great in its current form.

 I agree it could be clearer, but it matches the style of the other diagrams in this section. Updating one in isolation would make the section inconsistent. Should we try to get design to do a broader pass after VSD? 

> I think using the same IDs for the chunks and the documents make the example a bit confusing. Given an ID like 1, is that from a chunk or a document? How about using IDs like 200, 201, ... for the documents, like in the diagram?

Updated in [0ead6b8](https://github.com/qdrant/landing_page/pull/2362/commits/0ead6b8f8b7e1188da00d7e23f67baf872684599)

> How about making this a testable snippet too?

(replied inline) The argument value is in the prose and the explicit form is already in the snippet above, so it's redundant.
I dropped this code block in [13690c4](https://github.com/qdrant/landing_page/pull/2362/commits/13690c49acea5f3542a1e18b821365c987604639), looks cleaner that way. 
Let me know if that's okay with you.

abdonpijpelink (approved):
LGTM. Thanks @Dylancouzon

PR #2365, opened 2026-05-21
[fix-cognee] Revamp the new cognee API i.e., recall and remember

lucifertrj:
- Cognee's new API, i.e., after cognee>1.0, it now uses recall and remember over search and cognify accordingly. 
- Also, we need to declare graph_dataset_database_handler for set_graph_db_config to avoid multiple backend tracking errors. 
- Main resource is also updated on the: [cognee-community GitHub](https://github.com/topoteretes/cognee-community/blob/main/packages/vector/qdrant/example.py)


Anush008 (approved):
Thanks!

PR #2392, opened 2026-06-01
Add blog: Branch-Aware Semantic Code Search with Qdrant (feedback wanted)

Dylancouzon:
Preview (Read PR first): https://deploy-preview-2392--condescending-goldwasser-91acf0.netlify.app/blog/branch-aware-code-search/

Reference tutorial: https://qdrant.tech/documentation/tutorials-search-engineering/branch-aware-search/?q=branch+aware#branch-aware-search-over-versioned-documents

# Add blog: Branch-Aware Semantic Code Search with Qdrant (feedback wanted)

Opening this for feedback on the approach before the final SEO polish and publish. The post itself is in good shape; what I'm mainly after is a read on the **design decisions** and one open question (blog vs article).

New post: `/blog/branch-aware-code-search/`. It teaches the branch-aware search pattern over a real codebase and points to the Branch-Aware Search tutorial for the implementation.

## Design decisions I'd like feedback on

1. **Complement the tutorial, don't repeat it.** The tutorial shows the implementation (payload model, filters, indexes, working code). This post deliberately stays at the decision level: what to chunk, how to track a chunk's identity across versions, and how to keep the index in sync with Git. Split: tutorial = how to build it, blog = how to reason about it. Does that division hold up?

2. **Lead with code, even though the pattern is general.** Branch-aware search works for any versioned corpus (docs, config, schemas), but we led with code because it's the most common version-controlled corpus and the strongest for SEO compared to the other document types. The post states that the pattern generalizes. Reasonable to lead with code?

3. **Cover best practices and caveats, not just the happy path.** We wanted the post to help someone do this properly, so it walks the real decisions and tradeoffs (chunking strategy, symbol identity across renames/moves/overloads, merges/rebases/force-push, retention). This is where it got complicated and grew into near-article territory, which leads to the open question.

## Open question: blog or article?

It currently sits between the two: blog-length, but with article-ish depth in the caveats. The two directions would look quite different:

- **Blog (where it is now):** general and informational. Concept-first, light on data, leans on the tutorial for the build. Broader reach, better for SEO and promotion.
- **Article:** much more in-depth and data-backed. Expanded benchmarks and citations, fuller worked examples (more of the real index and actual code), each caveat developed properly with tradeoffs. More of a reference piece than a promotional one, and it would live under `/articles/`.

Is the blog version enough, or is this worth the fuller article treatment? The subject is fairly niche, so I'm genuinely unsure. Feedback welcome.

## Headline options

Primary: **"Branch-Aware Semantic Code Search with Qdrant."** Alternatives: "Why Your Code Search Returns the Wrong Version" or "Search Your Codebase on the Branch You're Actually On."

## Still to do before merge

- Feedback on the design decisions and the blog-vs-article call.
- A hard SEO pass with our internal target keywords.

🤖 Generated with [Claude Code](https://claude.com/claude-code), verified by me


kanungle (approved):
> This post deliberately stays at the decision level: what to chunk, how to track a chunk's identity across versions, and how to keep the index in sync with Git. Split: tutorial = how to build it, blog = how to reason about it. Does that division hold up?

Yes

> Branch-aware search works for any versioned corpus (docs, config, schemas), but we led with code because it's the most common version-controlled corpus and the strongest for SEO compared to the other document types. The post states that the pattern generalizes. Reasonable to lead with code?

No; blogs should not really have code in them

> Open question: blog or article?

> Blog (where it is now): general and informational. Concept-first, light on data, leans on the tutorial for the build. Broader reach, better for SEO and promotion.

> Article: much more in-depth and data-backed. Expanded benchmarks and citations, fuller worked examples (more of the real index and actual code), each caveat developed properly with tradeoffs. More of a reference piece than a promotional one, and it would live under /articles/.

Keep light as a blog. I don't think we want to go through the effort of enriching the technical depth as far as an article requires

kanungle:
@generall do you have a preference on blog vs article on this one? I think it can be an article

kanungle:
It would need a little more content detail though. Or we can keep it light as a blog

generall:
> do you have a preference on blog vs article on this one? I think it can be an article

Is that an ad for tutorial page? If so - blog.

kanungle (approved):
please update the publish date before merging

PR #2408, opened 2026-06-09
Advise match any over should for multivalue filtering

abdonpijpelink:
[Preview](https://deploy-preview-2408--condescending-goldwasser-91acf0.netlify.app/documentation/search/filtering/#should)

Users have observed the undocumented behavior that, when filtering over multiple values of the same field, `match any` is faster than `should`. This is because `match any` is a single condition with a fast lookup, while `should` evaluates each value as a separate condition.

This PR adds a small note to the `should` section pointing users to `match any` when all `should` conditions target the same field.

### DIFF
--- a/qdrant-landing/content/documentation/search/filtering.md
+++ b/qdrant-landing/content/documentation/search/filtering.md
@@ -75,6 +75,8 @@ Filtered points would be:
 ]
 ```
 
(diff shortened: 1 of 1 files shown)


timvisee (approved):
Makes sense, thanks.

In the future we may make our query planner smarter to automatically do such conversion. But that won't be a short term thing.

PR #2426, opened 2026-06-16
Migration and recovery options guide

abdonpijpelink:
[Preview](https://deploy-preview-2426--condescending-goldwasser-91acf0.netlify.app/documentation/migration-recovery-options/)

When migrating data, we see users choose the migration tool when collection snapshots or backups would fit their use case better. To help customers choose the right tool for their use case, this PR adds a new guide that covers three migration and recovery options: when to use each, resource requirements, and operational trade-offs.

- Option 1: Migration tool (cross-DB, shard changes)
- Option 2: Core snapshots (env migrations, skipping index rebuilds)
- Option 3: Cloud backups (full failure recovery, cluster reversion)

### DIFF
--- a/qdrant-landing/content/documentation/capacity-planning.md
+++ b/qdrant-landing/content/documentation/capacity-planning.md
@@ -3,7 +3,7 @@ title: Capacity Planning
 short_description: "Size a Qdrant cluster by balancing RAM and disk for vectors, payload, indexes, replication, and quantization workloads."
 description: "Plan Qdrant cluster capacity: estimate RAM and disk for vectors, payloads, indexes, replication, and quantization to match your workload."
 partition: deploy
-weight: 238
+weight: 115
 aliases:
   - capacity
   - /documentation/cloud/capacity-sizing
@@ -109,6 +109,7 @@ Always remember to add 50% of the vector size. This would account for things lik
 total_vector_size = number_of_dimensions * 4 bytes * 1.5
 ```
 
-**Disclaimer**
+**Disclaimers**
 
-The above calculations are estimates at best. If you're looking for more accurate numbers, you should always test your data set in practice.
+- The above calculations are estimates at best. If you're looking for more accurate numbers, you should always test your data set in practice.
+- Migration scenarios require more headroom than normal operations. When using the Migration Tool or restoring a snapshot, the target cluster needs twice the disk of the source collection. When using the Migration Tool, it also needs twice the RAM. See [Migration and Recovery Options](/documentation/migration-recovery-options/) for details. 
--- a/qdrant-landing/content/documentation/capacity-planning.md
+++ b/qdrant-landing/content/documentation/capacity-planning.md
@@ -112,4 +112,4 @@ total_vector_size = number_of_dimensions * 4 bytes * 1.5
 **Disclaimers**
 
 - The above calculations are estimates at best. If you're looking for more accurate numbers, you should always test your data set in practice.
-- Migration scenarios require more headroom than normal operations. When using the Migration Tool or restoring a snapshot, the target cluster needs twice the disk of the source collection. When using the Migration Tool, it also needs twice the RAM. See [Migration and Recovery Options](/documentation/migration-recovery-options/) for details. 
+- [Migration scenarios](/documentation/migration-recovery-options/) require more headroom than normal operations. When using the Migration Tool or restoring a snapshot, the target cluster needs twice the disk space currently used by the source collection. When using the Migration Tool, it also needs twice the RAM currently in use. To determine the current disk and RAM usage of your collection, check the [Web UI](/documentation/ops-monitoring/memory-usage/) or use the [API](/documentation/ops-monitoring/memory-usage/#api).
--- a/qdrant-landing/content/documentation/migrate-to-qdrant/from-qdrant.md
+++ b/qdrant-landing/content/documentation/migrate-to-qdrant/from-qdrant.md
@@ -17,6 +17,10 @@ Use the `qdrant` subcommand to copy a collection from one Qdrant instance to ano
 - **Target collection name** (must be different from source if using the same instance)
 - **API keys** — for each instance, if authentication is enabled
 
+<aside role="alert">
+  The target cluster needs <strong>twice the RAM and disk</strong> of the source collection during migration. Qdrant keeps original segments alongside newly built ones during optimization. You can scale back down once the migration completes.
+</aside>
+
 ## Concept Mapping
 
 | Source Qdrant | Target Qdrant | Notes |
--- a/qdrant-landing/content/documentation/migrate-to-qdrant/from-qdrant.md
+++ b/qdrant-landing/content/documentation/migrate-to-qdrant/from-qdrant.md
@@ -7,7 +7,9 @@ weight: 110
 
 # Migrate Between Qdrant Instances
 
-Use the `qdrant` subcommand to copy a collection from one Qdrant instance to another — or between collections within the same instance. The tool automatically recreates the full collection schema (vector config, HNSW settings, quantization, sharding) on the target.
+Use the `qdrant` subcommand to copy a collection from one Qdrant instance to another or copy data between collections within the same instance.
+
+If the target collection does not exist, the tool automatically creates it using the source collection's configuration (vector config, HNSW settings, quantization, sharding). To change the target collection's configuration, create the target collection manually before running the migration.
 
 ## What You Need
 
--- a/qdrant-landing/content/documentation/migration-recovery-options.md
+++ b/qdrant-landing/content/documentation/migration-recovery-options.md
@@ -1,0 +1,122 @@
+---
+title: "Migration and Recovery"
+short_description: "Compare the migration tool, core snapshots, and cloud backups to choose the right approach for moving or recovering your Qdrant data."
+description: "Decide between Qdrant's three data movement options — migration tool, collection snapshots, and cloud backups — based on use case, resource requirements, and operational trade-offs."
+partition: deploy
+weight: 130
+---
+
+# Migration and Recovery Options
+
+Qdrant gives you three mechanisms for moving or recovering data, each suited to a different scenario. The right choice depends on what you're migrating from, whether you need to change the shard count, and whether you're on Qdrant Cloud.
+
+**Quick guide:**
+
+- Use the **[Qdrant Migration Tool](#qdrant-migration-tool)** when moving data from other vector search engines (for example, from Pinecone, Weaviate, or Chroma to Qdrant) or when you need to change shard counts, apply filters, or migrate a subset of data.
+- Use **[Snapshots](#snapshots)** when moving a Qdrant collection between environments (for example, self-hosted to Cloud) and you want to skip re-indexing by transferring the existing index.
+- Use **[Cloud Backups](#cloud-backups)** for full cluster failure recovery or reverting a cluster to a previous state. This is the fastest restore option when you're on Qdrant Cloud.
+
+## Qdrant Migration Tool
+
+The [Qdrant Migration Tool](/documentation/migrate-to-qdrant/) streams points from a source database directly to a target Qdrant cluster over the network. Because data arrives as individual upserts, Qdrant builds the HNSW index from scratch. This is the slowest option for indexing but the most flexible in terms of source system and target shape.
+
+**When to use it:** Migrating from any of the [supported sources](/documentation/migrate-to-qdrant/#supported-sources), including other Qdrant clusters; changing the number of shards; or migrating a filtered subset of vectors.
+
+<aside role="alert">
+During migration, the target cluster needs <strong>twice the RAM and disk</strong> of the source collection. You can scale RAM back down once the migration is complete.
+</aside>
+
+See the [Data Migration tutorial](/documentation/tutorials-operations/migration/) for step-by-step instructions.
+
+## Snapshots
+
+Snapshots are `tar` archives of a collection's data, payload, and its pre-built index. Restoring a snapshot to a new cluster skips the indexing phase entirely, which can save hours on large collections.
+
+**When to use it:** Moving a Qdrant collection between environments (for example, self-hosted to Cloud or Managed Cloud to Private Cloud). For self-hosted deployments, snapshots also let you recover from a full cluster failure or revert to a known-good state after a failed deployment or data corruption.
+
+The target cluster still needs approximately two times the collection's disk size during the restore window, because the snapshot file and the restored collection exist simultaneously on disk until the restore is complete.
+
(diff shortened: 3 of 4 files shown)


qdrant-landing/content/documentation/migrate-to-qdrant/from-qdrant.md
```diff
@@ -17,6 +17,10 @@ Use the `qdrant` subcommand to copy a collection from one Qdrant instance to ano
 - **Target collection name** (must be different from source if using the same instance)
 - **API keys** — for each instance, if authentication is enabled
 
+<aside role="alert">
+  The target cluster needs <strong>twice the RAM and disk</strong> of the source collection during migration. Qdrant keeps original segments alongside newly built ones during optimization. You can scale back down once the migration completes.
+</aside>
```
timvisee, line 24: There are some cases where this isn't necessary. For example when using a snapshot targeting an empty cluster and all data was already optimized.

But let's keep this, it's a good and very clear rule of thumb.

qdrant-landing/content/documentation/migration-recovery-options.md
```diff
@@ -0,0 +1,122 @@
...
+
+## Snapshots
+
+[Snapshots](/documentation/snapshots/) are `tar` archives of a collection's data and its pre-built index. Restoring a snapshot to a new cluster skips the indexing phase entirely, which can save hours on large collections.
+
+**When to use it:** Moving a Qdrant collection between environments (for example, self-hosted to Cloud, or Managed Cloud to Private Cloud). Snapshots also let you restore collections to a known-good state after accidental deletion, data corruption, or a failed data update.
+
+The target cluster still needs approximately two times the collection's disk size during the restore window, because the snapshot file and the restored collection exist simultaneously on disk until the restore is complete.
+
+Snapshots require the target to run the same or a newer version of Qdrant.
```
timvisee: It can be at most one minor version higher.

For example: 1.17.1 to 1.18.1 is allowed. 1.16.0 to 1.18.1 is not allowed.

It is entirely possible that it 'just works'. But we don't guarantee it and thus it's very version specific.

qdrant-landing/content/documentation/migration-recovery-options.md
```diff
@@ -0,0 +1,122 @@
...
+| **Scope** | Per collection (optional filter) | Per collection or shard | Entire volume |
+
+### Performance & Cost
+
+How each option handles indexing on the target, data transfer speed, network cost, and load placed on the source cluster.
+
+| | Migration Tool | Snapshots | Cloud Backups |
+|---|---|---|---|
+| **Indexing on target** | Full re-index (slow) | None (index transfers with data) | None (direct disk clone) |
+| **Transport speed** | High (network stream) | Moderate (disk I/O + upload) | Very fast (metadata clone) |
```
timvisee: Transport speed: snapshots are much faster than the migration tool.

Both write data to disk. But for migration there is a lot more overhead: serialize data, scroll, transform into upserts, deserialize data, write to disk, optimize

qdrant-landing/content/documentation/migration-recovery-options.md
```diff
@@ -0,0 +1,122 @@
...
+| **Network cost** | Medium (streamed data) | High (potential cloud egress) | High / None (stays within region) |
```
timvisee: Both the migration tool and snapshot have (potentially) cloud egress.

It's likely true that snapshots are a bit larger, yes.

qdrant-landing/content/documentation/migration-recovery-options.md
```diff
@@ -0,0 +1,122 @@
...
+| **Impact on source** | Medium-High (CPU/IO intensive) | Moderate (CPU/IO for compression) | Minimal (handled by storage) |
```
timvisee:
```suggestion
| **Impact on source** | Medium-High (CPU/IO intensive) | Moderate (CPU/IO) | Minimal (handled by storage) |
```

Snapshots are a tar archive, but a tar has no compression.

qdrant-landing/content/documentation/migration-recovery-options.md
```diff
@@ -0,0 +1,122 @@
...
+
+### Operations & Mechanics
+
+How much effort each option requires, which APIs or tools you'll use, and where data lands during the operation.
+
+| | Migration Tool | Snapshots | Cloud Backups |
+|---|---|---|---|
+| **Operational effort** | Moderate (CLI setup) | High (manual API scripting) | Low (automated, cluster-wide) |
```
timvisee: I'd categorize migration tool vs snapshot to be similar.

One API call on the target with the streaming URL from the source.

But yes, this is subjective.

qdrant-landing/content/documentation/migration-recovery-options.md
```diff
@@ -0,0 +1,122 @@
...
+| **API / tooling** | CLI Tool | Qdrant REST API | Cloud infrastructure|
+| **Snapshot location** | N/A | External (S3 or compatible) | Cloud infrastructure |
+| **Restore target** | Any existing cluster | Any existing cluster | Restore within same region |
```
timvisee, line 86: Maybe something we can add:

Migration tool requires both the source and target to be online at the same time.

A snapshot can be taken offline first and uploaded later without the source being online. In cloud backups we also don't need the source to be online.

timvisee (changes requested):
Some nits. Take them with a grain of salt.

qdrant-landing/content/documentation/migrate-to-qdrant/from-qdrant.md
holyMolyTolli: > copy a collection [...] between collections within the same instance.

Makes no sense

qdrant-landing/content/documentation/migrate-to-qdrant/from-qdrant.md
holyMolyTolli: @timvisee 

> when using a snapshot targeting an empty cluster

But migration tool doesn't use snapshots(?) I think this is about the migration tool only.

qdrant-landing/content/documentation/migrate-to-qdrant/from-qdrant.md
holyMolyTolli: > The tool automatically recreates the full collection schema (vector config, HNSW settings, quantization, sharding) on the target.

Doesn't add something. Of course if I copy a collection it will recreate it's schema. More notable is that the schema can also be changed, so I would write that instead.

abdonpijpelink: Good catch. I've changed it into [...] or copy ***data*** between collections within the same instance.

qdrant-landing/content/documentation/capacity-planning.md
```diff
@@ -109,6 +109,7 @@ Always remember to add 50% of the vector size. This would account for things lik
 total_vector_size = number_of_dimensions * 4 bytes * 1.5
 ```
 
-**Disclaimer**
+**Disclaimers**
 
-The above calculations are estimates at best. If you're looking for more accurate numbers, you should always test your data set in practice.
\ No newline at end of file
+- The above calculations are estimates at best. If you're looking for more accurate numbers, you should always test your data set in practice.
+- Migration scenarios require more headroom than normal operations. When using the Migration Tool or restoring a snapshot, the target cluster needs twice the disk of the source collection. When using the Migration Tool, it also needs twice the RAM. See [Migration and Recovery Options](/documentation/migration-recovery-options/) for details.
```
holyMolyTolli: > Migration scenarios require more headroom than normal operations.

Bulk upsert scenarios e.g. with the migration tool.

> twice the disk [...] twice the RAM

Important to make it clear here that we are talking about the actual usage, not the provisioned disk/ram. SO one needs to check usage in the UI, not just the size of the nodes.

qdrant-landing/content/documentation/migration-recovery-options.md
```diff
@@ -0,0 +1,122 @@
...
+weight: 130
+---
+
+# Migration and Recovery Options
+
+Qdrant gives you three mechanisms for moving or recovering data, each suited to a different scenario. The right choice depends on what you're migrating from, whether you need to change the shard count, and whether you're on Qdrant Cloud.
+
+**Quick guide:**
+
+- Use the **[Qdrant Migration Tool](#qdrant-migration-tool)** when moving data from other vector search engines (for example, from Pinecone, Weaviate, or Chroma to Qdrant), when you need to change shard counts, or migrate a subset of data.
```
holyMolyTolli: > or Chroma to Qdrant)

")" needs to move to the left behind "Chroma"

qdrant-landing/content/documentation/migrate-to-qdrant/from-qdrant.md
abdonpijpelink: I believe this sentence came from user confusion about whether they need to create the target collection first or whether the tool does it for you. I've changed it into:

_If the target collection does not exist, the tool automatically creates it using the source collection's configuration (vector config, HNSW settings, quantization, sharding). To change the target collection's configuration, create the target collection manually before running the migration._

qdrant-landing/content/documentation/capacity-planning.md
abdonpijpelink: Good call. I've added a link to https://qdrant.tech/documentation/ops-monitoring/memory-usage/

qdrant-landing/content/documentation/migration-recovery-options.md
abdonpijpelink: Done

qdrant-landing/content/documentation/migration-recovery-options.md
abdonpijpelink: I've changed it into:

_The target cluster must match the source cluster's minor version, or be at most one minor version higher. For example, you can restore a 1.17.1 snapshot to a 1.18.1 cluster, but not a 1.16.0 snapshot._

qdrant-landing/content/documentation/migration-recovery-options.md
abdonpijpelink: I've dropped the "Transport" and simplified the row to:

```
| **Speed** | Moderate | Fast | Very fast |
```

qdrant-landing/content/documentation/migration-recovery-options.md
abdonpijpelink: Changed it into:

```
| **Network cost** | Medium (potential cloud egress from streamed data) | High (potential cloud egress) | High / None (stays within region) |
```

qdrant-landing/content/documentation/migration-recovery-options.md
abdonpijpelink: Done

qdrant-landing/content/documentation/migration-recovery-options.md
abdonpijpelink: Changed it so  both are "Moderate":

```
| **Operational effort** | Moderate (CLI setup) | Moderate (manual API scripting) | Low (automated, cluster-wide) |
````

File: qdrant-landing/content/documentation/migration-recovery-options.md (hunk 7 above)
Comment (abdonpijpelink) [reply qdrant-landing/content/documentation/migration-recovery-options.md:86]: Good call. I've added a row "Source availability":

```
| **Source availability** | Must be online throughout | Online during snapshot creation only; not required for upload or restore | Not required |
```

**Review by abdonpijpelink — COMMENTED** [review]
Thanks @timvisee and @holyMolyTolli ! I've applied your suggestions.

**Review by timvisee — APPROVED** [review]

File: qdrant-landing/content/documentation/migration-recovery-options.md (hunk 6 above)
Comment (holyMolyTolli) [reply qdrant-landing/content/documentation/migration-recovery-options.md]: @timvisee I think for snapshots we need to loop through the nodes which we do not do natively in qdrant (yet), so I would argue it's a bit more effort (looping script) than the migration tool (just one cli command)

File: qdrant-landing/content/documentation/migration-recovery-options.md [hunk 10]
```diff
@@ -0,0 +1,122 @@
...
+| **Target cluster shape** | Flexible | Moderate: same shard count | Strict: same config & node count |
+| **Target Qdrant version** | Any | Same or newer | Exact same version |
+
+### Resource Requirements
+
+Disk and RAM headroom needed on the target cluster, and how each option handles an interrupted operation.
+
+| | Migration Tool | Snapshots | Cloud Backups |
+|---|---|---|---|
+| **Target disk headroom** | 2× source | 2× source | ≥ source size |
```
holyMolyTolli, line 107: Same as above, maybe we should be clear here that this is about the data usage at source, not the provisioned space. maybe 

| **Target disk size** | 2× source data | 2× source data | ≥ source data size |

qdrant-landing/content/documentation/migration-recovery-options.md
```diff
@@ -0,0 +1,122 @@
...
+| **Target RAM headroom** | 2× source | Same as source | Same as source |
```
holyMolyTolli, line 108: See above

qdrant-landing/content/documentation/migration-recovery-options.md
```diff
@@ -0,0 +1,122 @@
...
+| **Resilience to interruption** | Checkpoint resume | Atomic restore | Atomic clone |
+
+## Resource Requirements at a Glance
```
holyMolyTolli: This the same table as above but transposed?

holyMolyTolli (commented):
Comments added

qdrant-landing/content/documentation/migration-recovery-options.md
abdonpijpelink: It's funny how sometimes you can read something many times and still be blind to the obvious. Removed the 2nd table.

qdrant-landing/content/documentation/migration-recovery-options.md
abdonpijpelink: Added a note below the table:

_“Source” refers to actual collection usage, not provisioned capacity. Check your current usage in the [Web UI](/documentation/ops-monitoring/memory-usage/) or via the [API](/documentation/ops-monitoring/memory-usage/#api)._

qdrant-landing/content/documentation/migration-recovery-options.md
timvisee: @holyMolyTolli you make a very good point.

If all is replicated everywhere it isn't necessary. But it's much more likely a node doesn't hold all replicas.

abdonpijpelink: I've changed the operational effort for snapshots into `High (per-node API scripting)`

holyMolyTolli (approved):

PR #2430, opened 2026-06-18
Add query decomposition for multi-hop questions

Dylancouzon:
Preview: https://deploy-preview-2430--condescending-goldwasser-91acf0.netlify.app/documentation/improve-search/query-decomposition/

Adds **Query Decomposition for Multi-Hop Questions** as a standalone tutorial under Improve Search (`improve-search/query-decomposition.md`) and links it from the section index.

### Why

Multi-stage queries and reranking only improve results already retrieved. For multi-hop questions, the needed information may never be fetched.

Decomposition solves this by iteratively retrieving, having an LLM generate the next sub-question, repeating up to a hop limit, and fusing results with Qdrant's RRF.

I tested this in a retrieval workshop and it improved answers on complex multi-hop queries that single-pass retrieval missed.

  ### Files
  - `improve-search/query-decomposition.md` — new tutorial
  - `improve-search/_index.md` — adds it to the section table

  Workshop: https://github.com/qdrant-labs/self-correcting-loops-workshop

### DIFF
--- a/qdrant-landing/content/documentation/improve-search/query-decomposition.md
+++ b/qdrant-landing/content/documentation/improve-search/query-decomposition.md
@@ -1,6 +1,6 @@
 ---
 title: Query Decomposition for Multi-Hop Questions
-short_description: "Answer multi-hop questions by retrieving in steps: an LLM asks each follow-up sub-question, then Qdrant fuses the results with RRF."
+short_description: "Answer multi-hop questions by retrieving in steps: an LLM asks each follow-up sub-question, then fuse the per-hop results with RRF."
 description: "Answer multi-hop questions in Qdrant: decompose the query into retrieval steps, let an LLM ask each follow-up, and fuse results with RRF."
 weight: 8
 partition: ecosystem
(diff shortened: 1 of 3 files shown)


abdonpijpelink (changes requested):
This is a great use case example, but I don't think it belongs on the hybrid queries docs page in its current form. 1) it uses an external API (OpenAI in this case) which makes it more of an "Ecosystem" topic, and 2) the snippets are Python-only.

How about turning this into a short tutorial under "[Improve Search](https://qdrant.tech/documentation/improve-search/)" and then we can link to it from the hybrid queries page?

Dylancouzon:
@abdonpijpelink appreciate the feedback. I updated the PR accordingly

qdrant-landing/content/documentation/improve-search/query-decomposition.md
```diff
@@ -0,0 +1,112 @@
...
+    return response.points
+```
+
+After each hop, the LLM reads the results so far and names the one fact still missing, or replies `DONE`. Picking the next sub-question is a light task, so a small, fast model handles it well:
+
+```python
+def next_subquestion(question, hops):
+    """Ask the LLM what to retrieve next, or return None when nothing is missing."""
+    # top 3 chunks from every hop so far
+    context = "\n".join(hit.payload.get("text", "") for hits in hops for hit in hits[:3])
```
generall: this line is hard to understand, double inside-out loops in a single line with not that descriptive names

qdrant-landing/content/documentation/improve-search/query-decomposition.md
```diff
@@ -0,0 +1,112 @@
...
+```python
+def fused_retrieve(queries, limit=10):
+    """Retrieve for every sub-question and let Qdrant merge them with built-in Reciprocal Rank Fusion (RRF)."""
+    response = client.query_points(
+        collection_name="{collection_name}",
+        prefetch=[
+            models.Prefetch(query=embed(q), using="dense", limit=limit)
+            for q in queries
+        ],
+        query=models.RrfQuery(rrf=models.Rrf()),
```
generall: If we use RRF for fusing independent queries, we should at least tie limit to number of sub-queries. Otherwise some of the queries responses might be just lost. And as a more correct approach I would propose to use batched query instead of prefetches.

qdrant-landing/content/documentation/improve-search/query-decomposition.md
```diff
@@ -0,0 +1,112 @@
...
+hops = [retrieve(question)]  # their results, used to steer the LLM
+
+for _ in range(MAX_HOPS):
+    follow_up = next_subquestion(question, hops)
+    if follow_up is None:
+        break
+    queries.append(follow_up)
+    hops.append(retrieve(follow_up))
+
+pool = fused_retrieve(queries)  # Qdrant fuses all sub-questions with built-in RRF
```
generall: We have already made retrieval for all sub-queries in `hops.append(retrieve(follow_up))`, why do we query database again? Is it only for fusion? If so, it doesn't seem very efficient.

generall (changes requested):
An example of real-live executio for the `Where was the director of the film Inception born?` might be interesting to demonstrate

qdrant-landing/content/documentation/improve-search/query-decomposition.md
Dylancouzon: Rather than tying the limit or batching, I fused each hop's full result list client-side, so no sub-query's results get cut off. Seemed more elegant. 
Thanks for finding the issue.

qdrant-landing/content/documentation/improve-search/query-decomposition.md
Dylancouzon: Applied.

qdrant-landing/content/documentation/improve-search/query-decomposition.md
Dylancouzon: Applied. Now Fusing hop results we already have instead of re-querying.

Dylancouzon:
@generall thanks for the feedback. 2 resolved inline thread and one waiting for your confirmation.
I added a real run of the Inception question showing the generated follow-up and fused chunks. Also rewrote a few sections to keep the tutorial lean and readable without losing the reader.

kanungle (approved):
@abdonpijpelink i think we need to find a better place for these tutorials. Ecosystem feels like the wrong place

PR #2435, opened 2026-06-18
Add article: Route Retrieval by Evidence

Dylancouzon:
Direct link: https://deploy-preview-2435--condescending-goldwasser-91acf0.netlify.app/articles/route-retrieval-by-evidence/

**WIP open for DevRel review.**

New articles piece: route retrieval by evidence (cheap signals + a gate), not question shape. Numbers from the [self-correcting loops workshop](https://github.com/qdrant-labs/self-correcting-loops-workshop).

Link checker failed due to reference to https://github.com/qdrant/landing_page/pull/2430


generall:
I like the idea of the article - predicting if the result is meaningful or just random noize. However, the `for Self-Correcting RAG` seems redundant.

The decision we make is trivial - either ask more questions or not. Does it really need 5 sections to explain?
The proposed classification logic is something we call a "decision stump", based on very basic statistics - the first thing people will try anyway. Dedicating a full article to something as trivial as this doesn't present us as experts in the area. 

My recommendation would be:

- make the "rag" part minimal, keep the motivation part, but drop the `What Routing by Evidence Buys` and below.
- Concentrate more on the confidence estimation approach - this is the most complex and interesting question the original article highlights. Here are a few ideas: 
  - Can multiple dense embeddings help? Article estimates correlation between sparse and sense results, but what if we have multiple dense vectors?
  - Can rank changes after rescoring can help? E.g. measure if ranks stays the same VS change a lot after rescoring
  - Can embedding configence be a signal (see https://github.com/qdrant/block-embeddings, and feel free to explore other methods)

kanungle:
The technical article category names have been updated. Please update this article frontmatter accordingly

Dylancouzon:
@generall Thanks for the feedback. Did a good amount of rewriting. Changes applied:

> the `for Self-Correcting RAG` seems redundant / make the "rag" part minimal

Gone. Changed the title, and cut out unnecessary sections

> The decision we make is trivial - either ask more questions or not. Does it really need 5 sections to explain?

It does not, shortened. 

> The proposed classification logic is a "decision stump"... doesn't present us as experts.

Fair. I benchmarked the signals across three corpora. Which signal fires on which failure mode is the point now, not the threshold.

> Concentrate more on the confidence estimation approach

Done.

> Can multiple dense embeddings help?

Added a multi-dense agreement signal

> Can rank changes after rescoring help?

Tested rank churn after a ColBERT rescore a few ways. No conclusive results. 

> Can embedding confidence be a signal (block-embeddings)?

Not sure how this one fits the concept? Seems like a whole different approach

Other methods explored:
- Small classifier over all signals (logistic regression, gradient boosting): no lift over the single best signal
- Entity/evidence coverage: near chance on all three corpora, dropped.
- 20 signals were tested for the v1, so I already kept the most conclusive categories without having to list all of them

Other changes:
- Added two more corpora (NFCorpus and SciFact via BEIR) so the claims aren't from one dataset.
- New figure: signal separation (AUC) per signal, per corpus.
- Positioned against the relevant prior art (QPP, CRAG, Self-RAG, Adaptive-RAG, sufficient-context).
- Trimmed throughout.

kanungle (approved):

PR #2454, opened 2026-06-25
Increase maximum side navigation depth by one level

abdonpijpelink:
[Preview](https://deploy-preview-2454--condescending-goldwasser-91acf0.netlify.app/documentation/search/text-search/)

This PR increases the maximum documentation side navigation depth by one level.

Motivation: 
- Our docs have several very long pages (for example, the [text search guide](https://qdrant.tech/documentation/search/text-search/)), which hurts both readability and agent accuracy. For instance, agents recommend full-text payload index settings (for filtering) when users ask about BM25 configuration, because both concepts are very similar and live on the same page. 
- Breaking up pages at this level requires one extra nav depth level: text search logically belongs under "Search", but the sidebar currently can't render further sub-pages at that depth.

Having one additional depth enables us to break up more long pages into smaller sub pages, while working in the current docs structure.

https://github.com/qdrant/landing_page/commit/975366a611d0388b92c4a786912b38d6d04d5cde implements the max side navigation depth increase.
https://github.com/qdrant/landing_page/commit/3527e9c53afc2455c96b0fb763ed8173b1ed596a applies this to the text search guide as a concrete example to review ([preview](https://deploy-preview-2454--condescending-goldwasser-91acf0.netlify.app/documentation/search/text-search/)).

Screenshot of the additional side nav depth:
image

/cc @mrscoopers @szabosteve because we've discussed the need to have one more level in the past


generall:
Would it be possible to keep page names short, so they fit a single line in a navbar?

abdonpijpelink:
> Would it be possible to keep page names short, so they fit a single line in a navbar?

I've shortened the long title to "Hybrid Search"

kanungle (approved):
I'm hesitant on this because having too many layers in docs can be overwhelming, but agree the current model is overwhelming too. Going ahead and approving, but we should use the tertiary level sparingly imho.

generall (approved):

PR #2457, opened 2026-06-26
Fix `ServiceMonitor` resource config for Prometheus

szabosteve:
### [Preview](https://deploy-preview-2457--condescending-goldwasser-91acf0.netlify.app/documentation/ops-monitoring/hybrid-cloud-prometheus/)

This PR increases the `interval` param of the `ServiceMonitor` config from `30s` to `60s`, so `scrapeTimeout` (55s) is less than the `interval` which is required by Prometheus.


abdonpijpelink (approved):
LGTM. Thanks!

PR #2484, opened 2026-07-08
Default rescoring docs are outdated

nleroy917:
I was running some tests with TurboQuant recently, and I noticed that the [docs for "Searching with quantization"](https://qdrant.tech/documentation/manage-data/quantization/#searching-with-quantization) are missing information about TurboQuant. 

Specifically, they say:
> By default, rescoring is only enabled for binary quantization. Other quantization methods do not rescore by default.

which (I believe) is not true now. I [found that](https://github.com/qdrant/qdrant/blob/44ad62f8cd69642be5afa6441612525e24a0d063/lib/segment/src/vector_storage/quantized/quantized_vectors.rs#L2276) rescoring is set to `true` by default for `TQBits::Bits1`, `TQBits::Bits1_5`,  and `TQBits::Bits2` in addition to binary quantization. So I updated that wording.

I'm not sold on how I articulated it, but was trying to be as clear as possible:
> By default, rescoring is only enabled for four quantization methods: binary quantization TurboQuant 1 Bit, TurboQuant 1.5 Bit, and TurboQuant 2 Bit. Other quantization methods do not rescore by default.

Open to changing that.

### DIFF
--- a/qdrant-landing/content/documentation/faq/qdrant-fundamentals.md
+++ b/qdrant-landing/content/documentation/faq/qdrant-fundamentals.md
@@ -205,7 +205,7 @@ By default, `hnsw_ef` equals `ef_construct` (default: 100). `ef_construct` is a 
 
 ### What are the default `rescore` values for different quantization methods?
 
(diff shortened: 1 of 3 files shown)


abdonpijpelink (approved):
LGTM. Thanks!

PR #2487, opened 2026-07-08
add tutorial: Indexing Payloads of Random Shape

Dylancouzon:
Preview: https://deploy-preview-2487--condescending-goldwasser-91acf0.netlify.app/documentation/tutorials-search-engineering/index-dynamic-payloads/

Adds a search-engineering tutorial for collections where the payload keys are data (per-vendor product feeds, arbitrary file metadata), so indexing each key separately doesn't scale.

**Pattern:** reshape variable keys into values in a fixed `{key, value}` array, index it once, query with a nested filter. Constant index count instead of one per key.

Index memory, measured on a clean instance at 10,000 points:

| Distinct keys | One index per key | Array + nested |
|---|---|---|
| 300 | +377 MiB | +24 MiB |
| 3,000 | +3.2 GiB | +24 MiB |

The array stays flat regardless of key count. Query cost barely moves: array+nested runs within ~0.2–0.5 ms of a dedicated per-key index (1.54 vs 1.01 ms at 10k points, 2.22 vs 2.03 ms at 100k). A per-key index is the fastest way to filter, but it's exactly what blows up memory across open-ended keys, so array+nested nearly matches that speed without the cost.

The article also covers a concatenated `key=value` variant (one index, 30–40% faster) for exact-match-only attributes, and a routing rule for which shape to use when.

**Files:** new `tutorials-search-engineering/index-dynamic-payloads.md`; reference pointer in `manage-data/indexing.md`; two tutorial-listing rows.


qdrant-landing/content/documentation/tutorials-search-engineering/index-dynamic-payloads.md
```diff
@@ -0,0 +1,242 @@
...
+
+| Points | Distinct keys (= indexes) | Build time | Memory added |
+|---|---|---|---|
+| 10,000 | 300 | 19.7 s | +377 MiB |
+| 10,000 | 1,000 | 63.4 s | +1,190 MiB |
+| 10,000 | 3,000 | 220.0 s | +3,203 MiB |
+
+A 40,000-key catalog would extrapolate to tens of GB of index memory.
+
+Two costs compound the memory. Every index created after ingest triggers a rebuild of the HNSW graph, repeated for each new key. Because user uploads define the schema, no fixed field list exists to index against, so the set of indexes grows with the data instead of staying fixed.
```
generall: > Every index created after ingest triggers a rebuild of the HNSW graph

It doesn't trigger a rebuid, but it does contribute to HNSW index time.

> Two costs compound the memory

Which one is the second?

qdrant-landing/content/documentation/tutorials-search-engineering/index-dynamic-payloads.md
```diff
@@ -0,0 +1,242 @@
+---
+title: Indexing Payloads of Random Shape
+short_description: "Index open-ended payloads whose keys are data in Qdrant, using a nested filter over a fixed key-value array instead of one index per key."
+description: "Model payloads with arbitrary keys in Qdrant. Reshape open-ended fields into a key-value array, index it at setup, and query exact matches and ranges with a nested filter."
+weight: 13
+aliases:
+  - /documentation/tutorials/index-dynamic-payloads/
```
generall: we only need aliases for migrating lages to a new location. New pages should avoid it.

generall (commented):
Overall LGTM

qdrant-landing/content/documentation/tutorials-search-engineering/index-dynamic-payloads.md
Dylancouzon: Claude slop, my bad. Fixed

qdrant-landing/content/documentation/tutorials-search-engineering/index-dynamic-payloads.md
Dylancouzon: My understanding was wrong, I thought that late index creation did trigger a rebuild. This was the second cost on top of the memory usage. 

Fixed, thanks for catching

generall (approved):

kanungle (approved):

PR #2492, opened 2026-07-10
Beginners course module 0

manas95826:
[PREVIEW](https://deploy-preview-2492--condescending-goldwasser-91acf0.netlify.app/course/beginners/)

PR #2494, opened 2026-07-10
Beginners course module 1

manas95826:
[PREVIEW](https://deploy-preview-2494--condescending-goldwasser-91acf0.netlify.app/course/beginners/module-1/)


qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,244 @@
...
+description: "Understand why keyword search breaks and how vectors solve it. Learn about embeddings, distance metrics, and hybrid search systems."
+isLesson: true
+weight: 20
+---
+
+{{< date >}} Module 1 {{< /date >}}
+
+# Let's Understand Search
+
+Understand why traditional search fails, how semantic (vector) search solves it.
```
kanungle: I think this is misleading. I think it should be something like:

> Understand why traditional search struggles and how modern semantic search improves it.

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,244 @@
...
+
+This approach works for predictable, structured queries. It breaks immediately on the language real users actually write.
+
+### Real-World Failure Examples
+
+| Query | Document in the index | Result |
+|-------|----------------------|--------|
+| car repair | automobile maintenance guide | ❌ Missed |
+| cheap flights NYC | affordable airfare to New York | ❌ Missed |
+| Apple stock | fruit company disambiguation? | ✔ Match |
```
kanungle: This line doesn't make sense to me

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,244 @@
...
+
+### The Three Core Failure Modes
```
kanungle: There are 4 in the list

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,244 @@
...
+- **Synonyms**: "car" ≠ "automobile" to a keyword engine, even though they mean the same thing. No word overlap = no match.
+- **Paraphrasing**: Same meaning, completely different words. "cheap flights" vs. "affordable airfare" are identical in intent, invisible to grep.
+- **Polysemy**: One word, multiple meanings. "apple" is a fruit company, a fruit, a music label. Context determines meaning that keywords can't.
+- **Word order**: "dog bites man" and "man bites dog" use identical words. Keyword search treats them as equivalent.
+
+## 2. How Traditional Search Improved
+
+Over time, search systems became more diverse. However, they all shared the same fundamental ceiling: they work on words, not meaning.
+
+### Evolution of Search Techniques
```
kanungle: is this actually a chronological evolution? I didn't look it up

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,244 @@
...
+
+## 3. Enter Semantic Search
+
+Semantic search changes the question from:
+
+**Keyword search asks:**
+"Does this document contain the same words?"
+
+**Semantic search asks:**
+"Does this document mean the same thing?"
```
kanungle: > mean the same concept

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,244 @@
...
+| Query | Matched document |
+|-------|------------------|
+| "How to fix a car" | "Repairing an automobile" ✔ |
+
+Zero shared words. Semantic search retrieves it anyway - because the meaning is equivalent.
+
+## 4. How It Works: Embeddings
+
+Semantic search works by converting text into vectors - lists of numbers that capture meaning. Similar meanings produce vectors that are close together in high-dimensional space. Different meanings produce vectors that are far apart.
+
```
kanungle, line 100: would be good to have a diagram here

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,244 @@
...
+Each dimension in the vector captures some aspect of the text's meaning. A 384-dimension model has 384 such aspects. No single dimension maps cleanly to a human concept like "color" or "emotional tone" - it's the combination of all dimensions together that encodes meaning.
+
+### Model Types
+
+- **Small models (128–384 dims)**: Fast, low memory. Good for well-scoped domains like product search or FAQ retrieval.
+- **Large models (768–1536 dims)**: More nuanced. Better for open-domain question answering and long-document retrieval.
+- **Domain-specific models**: Fine-tuned on legal, medical, or code corpora. Outperform general models on specialized content.
+- **Multimodal models**: Project text and images into the same vector space. Used by Tripadvisor and others (see Module 4).
+
```
kanungle, line 133: maybe add something like:

> Vector embeddings aren't limited by these models, however. They are theoretically capable of capturing any data into a transformed structured format

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,244 @@
...
+def cosine_similarity(a, b):
+    return np.dot(a, b) / (np.linalg.norm(a) * np.linalg.norm(b))
+
+query_vec = model.encode("car repair")
+doc_vec   = model.encode("automobile maintenance")
+
+score = cosine_similarity(query_vec, doc_vec)
+print(f"Similarity: {score:.3f}")   # e.g. 0.847
+```
```
kanungle, line 176: Replace this with a formula

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,244 @@
...
+
+### Distance Metric Comparison
+
+| Metric | Best for | Notes |
+|--------|----------|-------|
+| Cosine | Text similarity, NLP models | Robust to different vector magnitudes. Most common default. |
+| Dot product | When embeddings are normalized | Faster than cosine if vectors are unit-normalized at index time. |
+| Euclidean (L2) | Image embeddings, spatial data | Sensitive to magnitude - works best with models trained for it. |
+
```
kanungle, line 190: We have Manhattan too

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,244 @@
...
+#   SKU-48292  (score: 0.91)  ← wrong product
+#   SKU-48291  (score: 0.89)  ← correct product
+#   SKU-48290  (score: 0.87)  ← wrong product
+
+# With an exact filter applied:
+# must: { key: "sku", match: { value: "SKU-48291" } }
+#   SKU-48291  (score: 0.89)  ← only correct result returned
+```
+
+### Key insight
```
kanungle: needs `Title Case`

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,244 @@
...
+
+## 7. Modern Search = Hybrid Systems
+
+Production search today combines multiple retrieval signals in a single pipeline. Each signal handles a different class of query. Together, they cover the full spectrum of how real users search.
+
+### Hybrid Search Components
+
+- **Dense**: Semantic / vector - Intent, vibe, meaning
+- **Sparse**: BM25 / keyword - Exact terms, rare tokens
+- **Filters (Optional)**: Payload constraints - Hard rules, access control
```
kanungle: I'm not aware of any of our content that mentions "filtering" as a component of Hybrid Search

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,244 @@
...
+
+### Where Hybrid Search Is Used
+
+- **Multimodal RAG** - retrieve relevant text, images, audio, and video for LLM context windows
+- **Agentic AI systems** - multi-step agents that query different data sources sequentially
+- **E-commerce** - find semantically similar products, then filter by price, brand, and availability
+- **Knowledge bases** - semantic over documents, keyword for exact references and code snippets
+
```
kanungle, line 231: > - and more...

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,244 @@
...
+| Approach | Strength | Limitation |
+|----------|----------|------------|
+| Keyword / Grep | Fast, exact matching | No semantic understanding - misses synonyms and paraphrases |
+| BM25 / TF-IDF | Great for rare or specific terms | No synonym handling - relies entirely on word overlap |
+| Semantic / Dense | Understands meaning and intent | Can miss exact tokens - 'SKU-48291' may drift to similar IDs |
+| Hybrid | Best of both worlds | More complex to build, tune, and operate |
+
+## 8. References & Further Reading
+
+- **Qdrant Documentation** - [Qdrant Overview](https://qdrant.tech/documentation/concepts/)
```
kanungle: Name it "Qdrant Concepts"

kanungle (changes requested):
This is a good start. Left some comments. Not sure why Module 0 changes are in this PR

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,236 @@
...
+  - Overview of Qdrant's vector search engine - collections, points, payloads, and APIs.
+
+- **Distance Metrics Deep Dive** - [Distance Metrics - Qdrant](https://qdrant.tech/documentation/concepts/#distance-metrics)
+  - Cosine, dot product, Euclidean, and Manhattan - when to use each.
+
+- **Filtering & Hybrid Search** - [Filtering - Qdrant](https://qdrant.tech/documentation/concepts/filtering/)
+  - Payload filter syntax, indexed fields, and combining filters with vector queries.
+
+- **RAG Tutorials** - [RAG Tutorials - Qdrant](https://qdrant.tech/rag)
+  - End-to-end retrieval-augmented generation tutorials using Qdrant as the retriever.
```
kanungle: These links need to be href'd on the text itself, instead of beside it.

Example, just:

[Filtering & Hybrid Search]((https://qdrant.tech/documentation/concepts/filtering/)

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,236 @@
...
+## What's Next - Module 2
+
+In the next module, we'll break down:
+
+- What is a vector and why does it have hundreds of dimensions?
+- How do dimensions actually represent meaning?
+- How similarity really works under the hood - and when it fails.
+- Your first Qdrant collection: points, payloads, and your first query.
+
+End of Module 1. Continue to Module 2: First Principles of Vector Search.
```
kanungle: This needs to be a button link, like in the essentials course

kanungle (changes requested):

qdrant-landing/content/course/beginners/module-1/_index.md
kanungle: make this change on all modules

manas95826: I didn't add it intentionally because it might lead to PR runs failure because of disjoint modules, will add once everything is approved

kanungle (changes requested):
Progress since last round is real (formula, Manhattan row, concepts naming, intro line all fixed), but several items from my previous reviews are still open, so requesting changes again.

Still open from my last reviews:
1. **Module 0 files are still in this PR.** Please remove `course/beginners/_index.md` and the `module-0/` files so this PR only adds module 1; those shared files belong to #2492. Same request going out on every module PR.
2. **Line 202: `- and more...`** Cut it or name a real use case.
3. **Lines 213–226: reference links.** Href the text itself: `[Qdrant Concepts](https://qdrant.tech/documentation/concepts/)`, drop the link-beside-text pattern for all four.
4. **Title Case on headings** throughout: "Today's Path", "Core Limitation", "What You'll Learn", and so on.
5. **Syllabus still says "Hybrid Search: Dense + Sparse + Filters"** in `_index.md`. Same point I made on the module body: filters aren't a hybrid-search component. "Dense + Sparse". (Moves to #2492 once the shared files come out of this PR.)

New items:
6. **Line 141 is a broken sentence:** "A score of 0.0 means on the other hand can be interpreted as…". Suggest: "A score of 1.0 means the vectors point in the same direction. A score near 0.0 means the texts are semantically unrelated." Also drop "(identical meaning)"; cosine 1.0 is same direction, not identical meaning.
7. **Line 149: "around 0.847".** Did you run this? Three decimals plus "around" don't mix, and the score depends on the model version. Run it and quote the real number, or say "around 0.85".
8. **Line 83 table row is mislabeled and the example is broken:** typo tolerance is fuzzy matching, not "keyword matching", and "(receive → receive)" is the same word twice. → "Fuzzy matching | Tolerance for typos (recieve → receive)".
9. **Line 71: SPLADE doesn't belong in the pre-semantic evolution list.** It's a learned sparse model that came after dense retrieval. Drop it here; it's covered properly in Module 3. (This answers my earlier chronology question: no, it wasn't chronological.)
10. **Diagrams are off-brand.** The colors aren't from the Qdrant color palette and the shapes aren't from our diagram system. Rebuild them from the Docs/Diagrams Figma component library using the Qdrant palette.
11. **No exercise.** The module is all reading. Add one small task: embed three query/document pairs with the section 4 snippet and compare cosine scores.
12. "Best of both worlds" (line 211) → say what it does: "Covers both exact terms and intent".
13. `_index.md`: "vector databases" → "vector search engines" (hard rule), and "Duration: Multiple modules" is vague, this course has 6. The Path and syllabus should list all six modules.

qdrant-landing/content/course/beginners/module-1/_index.md
ESzyszka: Added a button, although now its a 404. The merge order is important here after the corrections to avoid conflicts: 

--> Merge Module 0 first 
--> Module 1 to 5 chronologically next

kanungle (changes requested):
### Must fix (blocks shipping)

- [x] **No video at the top of the module page.** Every course module page needs one — see the Essentials course pages for the embed pattern. Outro bumper yes, Intro bumper no.
- [x] **Diagrams are off-brand.** In `cosine-similarity.png` (and the pattern holds across the set): colors are outside the Qdrant palette (olive/dark yellow labels, non-ramp greens), and headings are set in mono instead of Mona Sans. Please rebuild from the Docs/Diagrams Figma library (https://www.figma.com/design/dlnrdaJ6pGlKyXoZquVhtZ/Docs---Diagrams?node-id=8137-14624) using palette tokens (Violet 50 `#8547FF`, Teal 50 `#038585`, P50 `#DC244C`, neutrals), Mona Sans headings, and Geist Mono values.
- [x] **Dead/outdated doc links.** The docs were restructured; these targets no longer exist: `/documentation/concepts/` and `#distance-metrics`, `/documentation/concepts/filtering/`, and `qdrant.tech/rag`. Use: filtering → `/documentation/search/filtering/`, RAG → an existing article such as `/articles/what-is-rag-in-ai/`. Also use relative paths (`/documentation/...`), not absolute `https://qdrant.tech/...`, matching the rest of the repo.
- [x] **Hand-rolled next button.** The closing `<a>` with inline styles (`background:#dc244c`) is off-system and targets `/course/beginners/module-2/`, which 404s until #2495 merges. Drop it and rely on the course layout's built-in next navigation (the `content.nextButton`/`nextDay` frontmatter pattern Essentials uses).
- [x] **Broken markdown:** `3. **TF-IDF / BM25 ` — unclosed bold, renders as literal asterisks.

### Should fix

- [x] **Unmeasured numbers presented as measurements.** "yields a similarity score around 0.85" for "car repair" vs. "automobile maintenance" on all-MiniLM-L6-v2 is asserted, not measured (expect meaningfully lower). Same question for the diagram's 0.96 / 0.62 / -0.28. Run the snippet in the lesson, use the actual printed scores in both prose and diagram. Also: the text says "a score near 0.0 means unrelated" while the diagram shows -0.28 — add one sentence that cosine can go negative.
- [x] **Nonsense table row.** `Apple stock | fruit company disambiguation? | ❌ Missed` — the middle cell isn't a document, and polysemy is a wrong-match failure, not a miss. Suggest: document = "apple harvest season guide", result = "wrong sense matched".
- [x] **Vague closing paragraph in §4.** "Vector embeddings aren't limited by these models, however. They are theoretically capable of capturing any data into a transformed structured format." says nothing concrete. Suggest: "Embeddings aren't text-only: the same idea works for images, audio, and code, which is how multimodal search in Module 4 works."
- [x] **Style sweep.** Spaced hyphens used as em-dash substitutes throughout ("no matter how closely related the meaning is", "Recency, popularity, personalization - payload values…") — replace with commas, colons, or two sentences. Heading case: "Today's Path" here vs. "Today's path" in Modules 0/2 and Essentials — match Essentials. Several images have label-only alt text ("Cheap Flights Example") — write descriptive alt text.

### Nice to have

- "(see Module 4)" appears twice — fine if Module 4's scope is locked, otherwise soften to "later in this course".
- The PR description says images will be added soon, but they're already in the diff — please update it to reflect the actual state.

ESzyszka:
All Fix and Must Fix comments addressed

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,271 @@
...
+
+Traditional search works by matching exact words. That's it. If the query string appears in the document, it's a hit. If it doesn't, it's a miss, no matter how closely related the meaning is.
+
+```python
+# Simple keyword search
+if "car repair" in document:
+    return document
+```
+
+![Keyword search only matches documents that contain the exact words "car" and "repair"](/courses/beginners/module-1/car-repair.png)
```
kanungle, line 65: This image is inconsistent because it shows a search containing "car" but not "repair" being returned, but not the inverse

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,271 @@
...
+
+print(len(query_vec))   # 384 dimensions
+print(query_vec[:5])    # [-0.12456684  0.03955902  0.08732592 -0.02361243 -0.0705409 ]
+```
+
+![An embedding model turns the text "car repair" into a fixed-length list of 384 numbers](/courses/beginners/module-1/generating-vector.png)
+
+### What Are Dimensions?
+
+Each dimension in the vector captures some aspect of the text's meaning. A 384-dimension model has 384 such aspects. No single dimension maps cleanly to a human concept like "color" or "emotional tone." It's the combination of all dimensions together that encodes meaning.
```
kanungle: Write something like:

> Vector embeddings are instead generated by complex neural networks

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,271 @@
...
+
+### Model Types
+
+- **Small models (128–384 dims)**: Fast, low memory. Good for well-scoped domains like product search or FAQ retrieval.
+- **Large models (768–1536 dims)**: More nuanced. Better for open-domain question answering and long-document retrieval.
```
kanungle: These are more like medium-sized models. Large models go even higher.

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,271 @@
...
+- **Domain-specific models**: Fine-tuned on legal, medical, or code corpora. Outperform general models on specialized content.
+- **Multimodal models**: Project text and images into the same vector space. Used by Tripadvisor and others (see Module 4).
```
kanungle: Not just text and images

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,271 @@
...
+print(query_vec[:5])    # [-0.12456684  0.03955902  0.08732592 -0.02361243 -0.0705409 ]
+```
+
+![An embedding model turns the text "car repair" into a fixed-length list of 384 numbers](/courses/beginners/module-1/generating-vector.png)
+
+### What Are Dimensions?
+
+Each dimension in the vector captures some aspect of the text's meaning. A 384-dimension model has 384 such aspects. No single dimension maps cleanly to a human concept like "color" or "emotional tone." It's the combination of all dimensions together that encodes meaning.
+
+### Model Types
```
kanungle: There are specific image-only embedding models, etc

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,271 @@
...
+
+### Cosine Similarity
+
+The most common metric for text. It measures the angle between two vectors, ignoring their magnitude (length) and focusing purely on direction. A score of 1.0 means the vectors point in the same direction, and a score near 0.0 means the texts are semantically unrelated. Cosine similarity can in principle range from -1 to 1, where negative values mean the vectors point in opposite directions. In practice, normalized text-embedding models like the one used here rarely produce negative scores, so unrelated text usually lands as a small positive number rather than below zero.
+
+$$
+\text{cosine\_similarity}(A, B) = \frac{A \cdot B}{\lVert A \rVert \, \lVert B \rVert}
+$$
+
+![Cosine similarity measures the angle between two vectors: a smaller angle gives a score closer to 1, and unrelated vectors score near 0](/courses/beginners/module-1/cosine-similarity.png)
```
kanungle: Opposite relation would be totally opposing directions

kanungle: consider adding

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,271 @@
...
+
+Dense similarity finds the neighborhood. Filters, exact matches, and payload constraints find the right point within that neighborhood. You need both.
+
+## 7. Modern Search = Hybrid Systems
+
+Production search today combines multiple retrieval signals in a single pipeline. Each signal handles a different class of query. Together, they cover the full spectrum of how real users search.
+
+### Hybrid Search Components
+
+- **Dense**: Semantic or vector search, for intent, vibe, and meaning.
```
kanungle: remove "vibe"

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,271 @@
...
+- **Sparse**: BM25 or keyword search, for exact terms and rare tokens.
```
kanungle: this reads as if there are only two types of sparse embeddings: BM25 and keyword

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,271 @@
...
+
+### Where Hybrid Search Is Used
```
kanungle: These examples are all modern search examples, not specific to Hybrid. Consider changing the heading

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,271 @@
...
+- [Filtering](/documentation/search/filtering/)
+  - Payload filter syntax, indexed fields, and combining filters with vector queries.
+- [What Is RAG in AI?](/articles/what-is-rag-in-ai/)
+  - How retrieval-augmented generation works and where a vector search engine fits in.
+
+## What's Next: Module 2
+
+In the next module, we'll break down:
+
+- What is a vector, and why does it have hundreds of dimensions?
```
kanungle: > hundreds to thousands of dimensions

kanungle (changes requested):

qdrant-landing/content/course/beginners/module-1/_index.md
kanungle: Do not mention customers either, and Module 4 has changed I believe

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,273 @@
...
+
+### What Are Dimensions?
+
+Vector embeddings are not hand-crafted; they're generated by complex neural networks trained on large amounts of data. Each dimension in the resulting vector captures some aspect of the text's meaning, but no single dimension maps cleanly to a human concept like "color" or "emotional tone." A 384-dimension model has 384 such aspects, and it's the combination of all of them together, not any one dimension, that encodes meaning.
+
+### Model Types
+
+- **Small models (128–384 dims)**: Fast, low memory. Good for well-scoped domains like product search or FAQ retrieval.
+- **Medium models (768–1536 dims)**: More nuanced. Better for open-domain question answering and long-document retrieval.
+- **Large models (3072+ dims)**: Generally higher fidelity at the cost of speed and memory, though quality depends more on training than dimension count alone.
```
kanungle: 4096 is the max

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,273 @@
...
+
+### Cosine Similarity
+
+The most common metric for text. It measures the angle between two vectors, ignoring their magnitude (length) and focusing purely on direction. A score of 1.0 means the vectors point in the same direction, and a score near 0.0 means the texts are semantically unrelated. Cosine similarity can in principle range from -1 to 1: a score near -1 means the vectors point in totally opposing directions, the opposite relation. In practice, normalized text-embedding models like the one used here rarely produce negative scores, so unrelated text usually lands as a small positive number rather than below zero.
+
+$$
+\text{cosine\_similarity}(A, B) = \frac{A \cdot B}{\lVert A \rVert \, \lVert B \rVert}
+$$
+
+![Cosine similarity measures the angle between two vectors: a smaller angle gives a score closer to 1, unrelated vectors score near 0, and opposite vectors score near -1](/courses/beginners/module-1/cosine-similarity.png)
```
kanungle, line 148: show:
1) similar
2) unrelated
3) opposite

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,273 @@
...
+1. The Problem: Why Keyword Search Struggles
+2. How Traditional Search Improved
+3. Enter Semantic Search
+4. How It Works: Embeddings
+5. Comparing Meaning: Distance Metrics
+6. Why Similarity Alone Is Not Enough
+7. Modern Search = Hybrid Systems
+8. References & Further Reading
+
+**Follow-along code**: [Module 1 notebook](https://github.com/qdrant/examples/blob/master/Beginner-course/Module1.ipynb)
```
kanungle: add colab notebook as a badge (see Tutorials for example)

kanungle (changes requested):
minor changes to text + a video to review is needed

Dylancouzon:
My review of the course and Notebook:

Disclaimers: I am not aware of what were the original instructions for this course, ignore the findings that contradict. Review was written manually and cleaned up with AI, this isn't review slop. Some changes are opinionated, feel free to disagree.

## Must fix

- [x]  1.Start with "What Is Search" before jumping into failure modes. This is a beginners course and some online feedback already points out that we expect people to know everything about HNSW
- [x] 2. Page isn't runnable. Fix code and installs, or keep pointing to the Notebook. Needs perfect parity.
- [x] 3. Use FastEmbed, not sentence-transformers. Same model. No reason to skip our own library. No `util.cos_sim`, so numpy cosine only. Touches all six notebooks
- [x] 4. Dense and sparse appear without setup. We say keyword and semantic for six sections, then switch. Sparse is also doing double duty: sparse data in the Manhattan row, sparse retrieval in section 7.
- [x] 5. We tell beginners semantic fixes word order. It doesn't. "dog bites man" vs "man bites dog" is 0.907 vs 0.570 for a real paraphrase. "safe for kids" vs "harmful to kids" is 0.779.
- [x] 6. SKU scores are inaccurate, use the ones from the Notebook.
- [ ] 7. Cosine diagram "opposite" is wrong. That pair is +0.78, not -0.99. Could use plain vectors instead of text labels.

## Should fix

- [x] 1. Cover environment setup. Assume some builders have little or no coding experience.
- [x] 2. Narration is dense and repetitive. "How Traditional Search Improved" repeats the table below it.
- [x] 3. Some AI writing signs: "X, not Y" eight times, rule-of-three, takeaway headings like "Core Limitation" and "Key Insight".
- [x] 4. "Model Types" are sizes, not types. Real axes are multilingual, static, cross-encoder, bi-encoder... Dimensions don't equal quality.
- [x] 5. Reorder section 4: embedding, embedding model, kinds of models, use cases, dimensions.
- [x] 6. `#distance-metrics` doesn't exist on `/documentation/manage-data/collections/`.
- [x] 7. Explain SKU. This is a vertical specific term
- [x] 8. Expand dot product around the choice, not the math. Cosine is direction only, dot product is direction times both magnitudes. Same number when vectors are normalized
- [x] 9. Section 4 snippet computes `doc_vec` and never uses it.
- [x] 10. References should go deeper: whitepapers, full articles, Wikipedia definitions.

## Nice to have

- [x] 1. Section 4 title says "Embeddings" and body says "vectors". Confusing.
- [x] 2. Dot product definition is inaccurate for our implementation. We normalize on upload and run cosine as dot product over normalized vectors, so no speed gain. Though it is technically how it should work. Not sure what's the best definition to use.
- [x] 3. Drop "Where Modern Search Is Used". Not hybrid-specific.
- [x] 4. Move "Try It Yourself: Compare Cosine Scores" under cosine.
- [x] 5. "Apple stock" row says wrong sense matched, the Notebook prints `NO MATCH` for the same example. Both are right for different engines, so say which one we mean.
- [x] 6. Expand acronyms on first use: NLP, SPLADE, RAG, LLM.

## Section 7 specifically

- [x] 1. Should land on semantic complementing keyword, not replacing it. Sections 1 and 2 argue keyword search is broken, then 7 says we use both.
- [x] 2. The SKU example is really the keyword-wins case. Probably a better anchor for that conclusion than the current framing.
- [x] 3. Quick Comparison repeats rows from earlier sections, could go too.

## Subjective, my opinions, feel free to disagree

1. Most of this module should be an intro to the concepts and vocabulary needed for the course. We shouldn't assume familiarity with any of the concepts
2. Section 8 references and resources should have whitepapers, full length articles, Wikipedia definitions for anyone who wants to dig into a subject.

Dylancouzon (changes requested):
See above comment

Dylancouzon:
- [x] Keep the token-index caveat out. It adds implementation detail before readers need it.
- [x] Make “How Traditional Search Improved” a short bridge into semantic search rather than another survey of retrieval techniques.

- [x] Keep “Enter Semantic Search” as the transition at the end of the keyword section, not a separate section.

- [x] Remove “that’s what the rest of this section is about.” The code and explanation already establish the purpose.

- [x] Add short comments to the embedding example explaining what the model call returns and why we inspect the vector length and first few values.

- [x] Cut the broad model taxonomy. For this course, explain the useful tradeoff: larger models can capture more context, but use more compute and memory.

- [x] Replace “Choosing a Model for Your Use Case” with why we chose this model for the course. Point readers to the embedding-model documentation when they start their own project.

- [x] Fold the definition of dimensions into the embedding explanation, rather than giving it its own subsection.

- [x] Add expected output to the cosine exercise so readers can compare their result with the lesson.

- [x] Simplify “What to look for” into a couple of plain takeaways tied directly to the output.

- [x] Cut the cosine versus dot-product explanation down substantially. The key point is enough: for normalized vectors, they produce the same ordering.

- [x] Simplify section 7 around the two useful lessons: embeddings still struggle with word order and negation, and exact identifiers need exact matching. Keep filtering and access control for the module where readers can use them.

- [x] Make “Exact Matching” a short continuation of the SKU example, and establish what keyword matching is earlier in the module.

- [x] Reframe section 8 as a brief preview: search can combine semantic and exact retrieval when the use case needs both; hybrid search is next module’s topic, not a universal requirement.

- [x] Fix the cosine diagram before merge. Its displayed scores do not match the examples: the “similar,” “unrelated,” and “opposite” pairs shown are not 0.96, 0.10, and -0.99.

- [x] Align the page and companion notebook on FastEmbed. The page says the Colab notebook runs its snippets, but the notebook still uses sentence-transformers; that leaves readers following two different implementations.

- [x] Replace the comment-only payload-filter block with either a runnable example in the appropriate module or plain prose. As written, it looks like code readers can use but has no imports, context, or result.

- [x] Remove the committed TODO in section 7 and include the real paraphrase comparison from the notebook. It gives the word-order lesson its most useful contrast: reversed meaning can score higher than a genuine paraphrase.

- [x] Make these cuts as a group. The module has grown to roughly 2,800 words, and the duplicated model setup, model-selection guidance, metric detail, and hybrid overview make it feel longer without adding a beginner-sized lesson.

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,292 @@
...
+  allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture; web-share"
+  referrerpolicy="strict-origin-when-cross-origin"
+  allowfullscreen>
+</iframe>
+</div>
+-->
+
+# Let's Understand Search
+
+Understand why traditional search struggles and how modern semantic search improves it.
```
Dylancouzon:
```suggestion
Traditional search retrieves documents by matching the words in a query to the words in an index. It’s fast and still useful for precise terms, but it can miss relevant results when people express the same intent differently. In this module, you’ll see how embeddings let search compare meaning, how similarity is measured, and why modern search often combines both approaches.
```

Added a quick intro

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,292 @@
...
+Choose how you want to follow this module:
+
+- **Use Colab:** [Open the companion notebook in Colab](https://colab.research.google.com/github/qdrant/examples/blob/master/Beginner-course/Module1.ipynb) to run the examples in your browser.
+- **Run locally:** Use Python 3.9 or later, then install the dependencies:
+
+```bash
+pip install fastembed numpy
+```
+
+The embedding model runs on your CPU. You don't need a GPU or API keys. The first time you run it, fastembed downloads the model file, so you'll need an internet connection and it may take a few minutes; after that it's cached locally and loads instantly, so don't mistake that first pause for an error.
```
Dylancouzon:
```suggestion
The embedding model runs on your CPU. You don't need a GPU or API keys. The first run downloads the model, so it may take a few minutes and requires an internet connection. Later runs use the cached model.
```

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,292 @@
...
+## 1. What Is Search?
+
+Search is the act of finding the right information out of everything you have, given a question. You type "car repair" into a box, and something has to decide which of your thousands of documents, products, or messages actually answers that.
+
+Every search system, no matter how it's built internally, does the same two things:
+
+1. **Retrieve**: narrow a huge collection down to a shortlist of documents that might be relevant.
+2. **Rank**: order that shortlist so the best answer ends up near the top.
+
+This module is about one question: how does a system decide what "relevant" means? We'll start with the simplest possible answer, watch it fail, and build up from there. No prior knowledge of vector search engines or indexing algorithms is assumed.
```
Dylancouzon:
```suggestion
At the heart of search is one question: what makes a result relevant? <br>
We'll start with the simplest possible answer, watch it fail, and build up from there. No prior knowledge of vector search engines or indexing algorithms is assumed.
```

Personal preference, feel free to deny

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,292 @@
...
+
+## 2. The Problem: Why Keyword Search Struggles
+
+In its simplest form, keyword search retrieves documents by exact word match: it checks whether the literal terms in your query appear in the document, with no understanding of what those terms mean. That works when people use the same terms as the content they need, as the example shows. (Real keyword systems add stemming, typo tolerance, and ranking on top, section 3 covers those, but none of them teach the system what words mean, which is the gap this module is really about.)
```
Dylancouzon:
```suggestion
In its simplest form, keyword search retrieves documents by matching query terms against terms in the document. In other words, it matches the words themselves, not what they mean. It works when the query and document use the same vocabulary, but it can miss relevant results when the same intent is expressed differently.
```

Less dense, removed parenthesis

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,292 @@
...
+
+That creates a few common problems:
+
+- **Different words, same meaning:** "car repair" misses "automobile maintenance."
+- **Same word, different meaning:** "Apple stock" can retrieve fruit content instead of financial information.
+- **Same words, different order:** "dog bites man" and "man bites dog" contain the same terms, but mean very different things.
+
+## 3. How Traditional Search Improved
+
+Keyword search picked up real upgrades over the years: faster lookups, relevance ranking, tolerance for typos, matching on word roots. Each made matching faster or more forgiving, but none of them taught the system what words mean. A keyword system can't know "car" and "automobile" are synonyms unless someone hard-codes that fact, and you can't hard-code an entire language.
```
Dylancouzon:
```suggestion
Traditional search has evolved beyond exact word matching. Techniques such as stemming, typo tolerance, and relevance ranking make it faster and more forgiving. But they still rely on words: the system cannot tell that “car repair” and “automobile maintenance” mean the same thing unless that connection is explicitly defined.
```

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,292 @@
...
+
+That's the gap **semantic search** closes. Instead of asking "Does this document contain the same words?" it asks "Does this document mean the same thing?" Nobody hand-codes the fact that "car" and "automobile" are related, the model learns it from the text it was trained on, and sentences with related meaning end up as vectors that sit close together, even when they share no words.
```
Dylancouzon:
```suggestion
That's the gap **semantic search** closes. Instead of asking "Does this document contain the same words?" it asks "Does this document mean the same thing?" Nobody hand-codes the fact that "car" and "automobile" are related, the embedding model learns it from the text it was trained on, and sentences with related meaning end up as vectors that sit close together, even when they share no words.
```

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,292 @@
...
+
+# check both vectors are the same length (384 dimensions each), then peek at the first 5 floats of each
+print(len(query_vec), len(doc_vec))
+print(query_vec[:5])
+print(doc_vec[:5])
+```
+
+![An embedding model turns the text "car repair" into a fixed-length list of 384 numbers](/courses/beginners/module-1/generating-vector.png)
+
+### Model Size: A Tradeoff
```
Dylancouzon:
```suggestion
### Model Size and Vector Dimension
```

We're listing 2 tradeoffs

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,292 @@
...
+
+Embedding models vary along two mostly separate axes. The model's parameter count and architecture (how "big" the model itself is) drive how much compute it takes to generate an embedding, and how much nuance it can capture. The output's dimension count (128–384 for smaller models, 1024+ for larger ones, like the 384 above) drives how much memory each vector takes to store and how expensive it is to compare at search time. The two don't always move together: a bigger model doesn't automatically output bigger vectors, and more dimensions alone isn't a quality signal, a well-trained small model can beat a poorly trained large one.
```
Dylancouzon:
```suggestion
When choosing an embedding model, two sizes matter:

- **Model size** is the number of parameters in the model. Larger models often capture more nuance, but need more compute to create embeddings.
- **Vector dimension** is the number of values in each embedding. Higher-dimensional vectors use more storage and take more work to compare during search.

These are related, but they are not the same. A larger model can produce a short vector, and a higher-dimensional vector does not automatically produce better search results
```

Much easier to read

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,292 @@
...
+
+This module uses `sentence-transformers/all-MiniLM-L6-v2` because it's small enough to run on a CPU with no API keys or GPU, and accurate enough to demonstrate semantic search clearly. When you start your own project, see [Points, Vectors and Payloads](/course/essentials/day-1/embedding-models/) for how to weigh size, language, and domain fit when picking a model.
+
+## 5. Comparing Meaning: Distance Metrics
+
+Once we have vectors, we need a way to measure how similar two of them are. Different metrics suit different situations.
+
+### Cosine Similarity
+
+Cosine similarity measures the angular similarity between two vectors. It focuses on whether vectors point in the same direction rather than on their length. This aligns well with many text embeddings, where the angle encodes meaning and the length is less important.
```
Dylancouzon:
```suggestion
Cosine similarity measures the angle between two vectors. Vectors that point in a similar direction receive a score closer to 1. Vectors that point in different directions receive a lower score. <br>
It ignores vector length, which is the overall size of a vector. For text embeddings, the direction often carries more useful information about meaning than this size. This makes cosine similarity a common choice for semantic search.
```

Can be a hard concept to grasp at first, so I'd simplify the definition as much as possible.

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,292 @@
...
+
+$$
+\text{cosine\_similarity}(A, B) = \frac{A \cdot B}{\lVert A \rVert \, \lVert B \rVert}
+$$
+
+Here $A \cdot B$ is the **dot product**: multiply each pair of matching positions in the two vectors and sum the results. $\lVert A \rVert$ is the vector's length (magnitude). Dividing the dot product by both lengths is what turns a raw dot product into a length-independent angle measurement.
```
Dylancouzon:
```suggestion
Here, $A \cdot B$ is the **dot product**: multiply each pair of matching values in the two vectors and add the results. $\lVert A \rVert$ is the vector’s length, also called its magnitude. Dividing by both vector lengths removes the effect of vector size, so the score measures their angle instead.
```
Small simplification

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,292 @@
...
+    doc_vec = list(model.embed([document]))[0]
+    score = cosine_similarity(query_vec, doc_vec)
+    print(f"{score:.3f}  |  {query!r}  vs  {document!r}")
+```
+
+Does `"apple stock"` score higher against the finance sense or the fruit sense, and does that match the sense you meant? Then compare that first score to the third row: does spelling out `"Apple Inc."` instead of `"apple"` pull the score toward the finance sense, and by how much?
+
+### Distance Metric Comparison
+
+| Metric | Best for | Notes |
```
Dylancouzon:
```suggestion
| Metric | Common use | Notes |
```
Less absolute

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,292 @@
...
+|--------|----------|-------|
+| Cosine | Text similarity, NLP (Natural Language Processing) models | Robust to different vector magnitudes. Most common default. |
```
Dylancouzon:
```suggestion
| Cosine | Text similarity, NLP (Natural Language Processing) models | Compares vector direction and ignores vector length. A common default for text embeddings. |
```

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,292 @@
...
+| Dot product | Vectors already normalized to unit length | Numerically identical to cosine similarity once vectors are unit length, same score, not a separate metric. |
```
Dylancouzon:
```suggestion
| Dot product | Vectors already normalized to unit length | Produces the same ranking as cosine similarity when every vector has unit length. |
```

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,292 @@
...
+| Euclidean (L2) | Image embeddings, spatial data | Sensitive to magnitude; works best with models trained for it. |
```
Dylancouzon:
```suggestion
| Euclidean (L2) | Image embeddings, spatial data | Measures the straight-line distance between vectors. Both direction and length affect the score. |
```

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,292 @@
...
+| Manhattan (L1) | Grid-like or count-based data | Sums absolute differences per dimension rather than squaring them first, making it less affected by extreme values in any single dimension. |
```
Dylancouzon:
```suggestion
| Manhattan (L1) | Grid-like or count-based data | Adds the absolute difference for each dimension. Less common for text embeddings. |
```

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,292 @@
...
+
+**Cosine vs. dot product:** for vectors normalized to unit length, dot product produces the exact same ranking as cosine similarity, it's a cheaper way to compute the same result, not a different metric. That's why Qdrant normalizes vectors on upload and computes a "Cosine" collection as a dot product internally.
```
Dylancouzon:
```suggestion
**Cosine vs. dot product:** When vectors are normalized to unit length, cosine similarity and dot product produce the same ranking. In that case, dot product is a simpler way to calculate the same result. Qdrant normalizes vectors when you use cosine distance, then uses dot product internally during search.
```

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,292 @@
...
+
+## 6. Why Similarity Alone Is Not Enough
+
+Sections 1 and 2 showed keyword search failing on synonyms, paraphrasing, polysemy, and word order. It's tempting to read that as "semantic search replaces keyword search." It doesn't, each is strong exactly where the other is weak, as the next two cases show. 
+
+In Qdrant, each thing you store is a [**point**](https://qdrant.tech/documentation/manage-data/points/): a vector plus an optional [**payload**](https://qdrant.tech/documentation/manage-data/payload/), arbitrary metadata like a timestamp or a permission list, and a [**collection**](https://qdrant.tech/documentation/manage-data/collections/) is the set of points you search over. [Filtering](https://qdrant.tech/documentation/search/filtering/) by recency, permissions, or other payload values, and combining that with ranking signals, is a separate layer on top of similarity, later modules cover it once you have a collection to filter.
```
Dylancouzon:
```suggestion
In Qdrant, each item you store is called a **point**. A point contains a vector and can also include a **payload**, which is metadata such as a timestamp or permission list. A **collection** is the group of points you search.

You can use a **filter** to restrict results by payload values, such as only returning documents a user can access. Filtering narrows the results before similarity search ranks them. Later modules show how to combine them in a Qdrant query.
```

Removed the word “thing,” and linking to four different documents in a single sentence is excessive. We'll define those later, so there’s no need to direct readers elsewhere now.

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,292 @@
...
+
+| Pair | Cosine similarity |
+|------|-------------------|
+| "dog bites man" vs "man bites dog" | 0.907 |
+| "safe for kids" vs "harmful to kids" | 0.779 |
+| "dog bites man" vs "a canine attacked a person" | 0.570 |
+
+The first two rows score high, even though each pair means something different: one flips who's doing the biting, the other flips safe into dangerous. For this model and this kind of phrasing, high lexical overlap still produces a high score, even though a person would read these pairs as opposites right away.
+
+The third row is the sharpest version of the problem: "a canine attacked a person" is a genuine paraphrase of "dog bites man," meaning the same thing in different words, yet it scores lower (0.570) than the reversed, opposite-meaning sentence (0.907). Shared words move the score more than shared meaning does.
```
Dylancouzon:
```suggestion
The third pair is a paraphrase: “a canine attacked a person” means nearly the same thing as “dog bites man.” Yet it receives a lower score (0.570) than the reversed sentence (0.907), which changes the meaning.
```

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,292 @@
...
+
+So: semantic search is great at synonyms and paraphrasing, but shaky on word order and negation. Don't rely on a similarity score alone anywhere it actually matters whether something is "safe" or "not safe."
```
Dylancouzon:
```suggestion
Semantic search works well for many synonyms and paraphrases, but similarity scores can still miss important details such as word order and negation. Do not rely on similarity alone when those details matter.
```

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,292 @@
...
+
+### Exact Matching: When You Need a Filter, Not Similarity
+
+Not every query needs semantic understanding. A query for an exact product code, like "SKU-48291" (SKU stands for Stock Keeping Unit, the retail term for this kind of code), needs an exact match instead. Try it yourself: embed the query and three candidate product codes, then compare their cosine scores.
```
Dylancouzon:
```suggestion
Not every query needs semantic understanding. A query for an exact product code, like "SKU-48291", needs an exact match instead. Try it yourself: embed the query and three candidate product codes, then compare their cosine scores.
```
Already defined right before

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,292 @@
...
+
+# Real output:
+#   0.730  |  SKU-48292  (wrong product)
+#   0.734  |  SKU-48291  (correct product)
+#   0.765  |  SKU-48290  (wrong product, scores HIGHEST)
+```
+
+The wrong product code coming out on top is a real failure, and the fix is a payload filter, not keyword search: restrict the search to points where the `sku` field equals `"SKU-48291"`, and you get back exactly that product, no embedding or similarity score involved. Module 2 shows how to write that filtered query for real, once we have a Qdrant collection to run it against.
+
+Keyword search (like BM25, covered in the hybrid search section below) is a separate tool from a payload filter: it ranks documents by shared vocabulary, rather than checking for one exact match. Similarity, filters, and keyword search each solve a different problem, none of them replaces the others.
```
Dylancouzon:
```suggestion
A payload filter is different from keyword search. A filter checks whether a field has an exact value. Keyword search, such as BM25, ranks documents by the words they share with the query. Similarity search, keyword search, and filters solve different retrieval problems, and production systems often combine them.
```

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,292 @@
...
+
+## 7. When a System Needs Both
+
+The product code example above is why some production systems combine semantic search with exact filters or keyword search rather than picking one: a single collection can serve queries that need meaning, queries that need one precise token, and queries that need term-based relevance ranking.
```
Dylancouzon:
```suggestion
The previous examples show why no single retrieval method works for every query. A search system may need semantic search to find related meaning, filters to match exact values such as a SKU, and keyword search when matching terms should affect the ranking.

**Hybrid search** specifically combines semantic retrieval with keyword retrieval. It uses **dense vectors** to compare meaning and **sparse vectors** to match and rank terms, often with an algorithm such as BM25.
```

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,292 @@
...
+
+Keyword search (like BM25, covered in the hybrid search section below) is a separate tool from a payload filter: it ranks documents by shared vocabulary, rather than checking for one exact match. Similarity, filters, and keyword search each solve a different problem, none of them replaces the others.
+
+## 7. When a System Needs Both
+
+The product code example above is why some production systems combine semantic search with exact filters or keyword search rather than picking one: a single collection can serve queries that need meaning, queries that need one precise token, and queries that need term-based relevance ranking. 
+
+This isn't a universal requirement, plenty of systems only ever need one of these, but when a use case needs both meaning-based and term-based retrieval, that's called **hybrid search**: combining **dense** (semantic/vector) retrieval with **sparse** (keyword-style, e.g. BM25, a decades-old algorithm that ranks documents by how much of the query's vocabulary they contain) retrieval.
```
Dylancouzon:
```suggestion
Not every search system needs hybrid search. But when users may search by either meaning or specific terms, combining both methods can return more relevant results.
```

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,292 @@
...
+- [Vector Embeddings Explained](/articles/what-are-embeddings/)
+  - A longer walkthrough of how embedding models are trained and how embeddings are used in ML and search.
+- [FastEmbed: Qdrant's Efficient Python Library for Embedding Generation](/articles/fastembed/)
+  - The library used in this module's code samples, and how it differs from running models directly.
+
+**Definitions:**
+
+- [Cosine similarity](https://en.wikipedia.org/wiki/Cosine_similarity)
+- [Dot product](https://en.wikipedia.org/wiki/Dot_product)
+- **Upsert**: to insert a point if its ID doesn't exist yet, or update it in place if it does. It's how you add and modify data in a Qdrant collection. See [Upload Points](/documentation/manage-data/points/#upload-points).
```
Dylancouzon:
```suggestion
## 8. Further Reading

- [Distance Metrics](/course/essentials/day-1/distance-metrics/) Learn more about cosine similarity, dot product, and Euclidean distance.
- [Vector Embeddings Explained](/articles/what-are-embeddings/) A deeper introduction to how embedding models turn data into vectors.
- [FastEmbed](/articles/fastembed/) Learn more about the library used to generate embeddings in this module.
```

I’d keep only resources that help them continue from this lesson

qdrant-landing/content/course/beginners/module-1/_index.md
```diff
@@ -0,0 +1,292 @@
...
+#   0.821  |  'cheap flights to New York'  vs  'affordable airfare to NYC'
+#   0.332  |  'cheap flights to New York'  vs  'best pizza in Chicago'
+```
+
+**What to look for:**
+
+- The synonym and paraphrase pairs (0.733, 0.821) score high despite sharing almost no words.
+- The unrelated pair (0.332) scores far lower, reflecting different meaning.
+
+**Your turn:** run this block (it reuses `model` and `cosine_similarity` from above) to test a polysemy case, and see how much a little extra context changes the score:
```
Dylancouzon:
```suggestion
**Your turn:** Run this block (it reuses `model` and `cosine_similarity` from above) to test a **polysemy** case, where the same word has more than one meaning. See how extra context changes the score:
```
Defining Polysemy

Dylancouzon (approved):
Hey, the content looks good to me. I made one final pass to reduce density and simplify vocabulary, especially for learners who may not speak English as a first language or are new to these concepts.

These are all small wording suggestions, so please take, tweak, or skip them as you see fit. I like the module overall, and I’m happy to consider it accepted from my side.

kanungle (approved):

kanungle:
This is really good, thanks. Let me know when the video is ready and I'll review and approve

kanungle:
updated diagrams:


keyword-search-refined


cosine-angle-fixed-text

kanungle (approved):
Looks great. Video is excellent.

Dylancouzon:
Minor nits 
Aligned Module 1 with Modules 2–5 by adding a Follow-along code link after the intro, pointing to Module1.ipynb. “Before You Start” now focuses solely on local setup.

Also fixed:
- The incorrect claim about hybrid search being the next module’s topic.
- A “Points, Vectors and Payloads” link that led to Embedding Models.
- A missing period in “Model Size and Vector Dimension.”
- Duplicate vector-length explanation in the cosine similarity section.

kanungle (approved):

PR #2496, opened 2026-07-10
Beginners course module 3

manas95826:
[PREVIEW](https://deploy-preview-2496--condescending-goldwasser-91acf0.netlify.app//course/beginners/module-3/)


qdrant-landing/content/course/beginners/module-3/_index.md
```diff
@@ -0,0 +1,356 @@
...
+description: "Understand dense vs sparse search, when each fails, and how hybrid systems combine them. Learn about fusion strategies, multimodal search, and real-world use cases."
+isLesson: true
+weight: 40
+---
+
+{{< date >}} Module 3 {{< /date >}}
+
+# Sparse vs Dense vs Hybrid Search
+
+Understand dense vs sparse search, when each fails, and how hybrid systems combine them.
```
kanungle: Replace with:

> Understand dense versus sparse retrieval, their strengths, and how a hybrid approach can combine them.

qdrant-landing/content/course/beginners/module-3/_index.md
```diff
@@ -0,0 +1,356 @@
...
+5. Fusion Strategies
+6. Beyond Text: Multimodal Search
+7. Real-World Use Cases
+8. References & Further Reading
+
+By the end, you'll understand when to use dense, sparse, or hybrid search and how to implement them in Qdrant.
+
+## 1. Where We Left Off
+
+In Module 2, you built a complete ingestion and retrieval pipeline: raw text → vector → store → top-K query. Dense-only search works well for natural language. It breaks immediately on structured identifiers.
```
kanungle: > Dense-only search works well for natural language. It breaks immediately on structured identifiers.

Replace with:

> Dense-only retrieval is best for semantic and contextual search. It struggles on structured identifiers.

qdrant-landing/content/course/beginners/module-3/_index.md
```diff
@@ -0,0 +1,356 @@
...
+| Query | SKU-48291 |
+|-------|-----------|
+| **The user wants exactly this product. No synonyms. No paraphrasing.** | |
+
+**Dense search returns**
+
+| SKU-48292 | (0.91) | ← wrong |
+|----------|--------|---------|
+| SKU-48291 | (0.89) | ← correct |
+| SKU-48290 | (0.87) | ← wrong |
```
kanungle: Thierry used to do an "iPhone" search example where the wrong model was retrieved for a ecommerce query. I think that works better than SKUs

qdrant-landing/content/course/beginners/module-3/_index.md
```diff
@@ -0,0 +1,356 @@
...
+
+🎧
+**Audio**
+Hum a melody → matching songs
+Audio fingerprints or spectrogram embeddings
+
+📝
+**Text**
+"cheap flights NYC" → semantic docs
+Sentence transformers, OpenAI embeddings, etc.
```
kanungle: I don't think we want emojis

qdrant-landing/content/course/beginners/module-3/_index.md
```diff
@@ -0,0 +1,356 @@
...
+  - How to configure and query multiple named vectors on the same point.
+
+## What's Next - Module 4
+
+Next, we'll explore:
+
+- Real enterprise architectures - Tripadvisor, HubSpot, and OpenTable in depth
+- Production patterns: multi-tenancy, agent memory, and RAG pipelines
+- Deployment options: Cloud, Hybrid Cloud, Edge, and self-hosted
+- Formula queries - when RRF and DBSF aren't enough
```
kanungle: Do not use real customer use cases in the course

kanungle (changes requested):
Looks okay. More is needed on IVF, BM25, SPLADE, miniCOIL, as well as how sparse vectors are encoded differently: mostly zeros (only non-zero values are entered). I also think filtering needs to be broken out as a separate "optional" part of this. Filtering is applicable on dense and sparse retrieval too, not just hybrid.

kanungle (changes requested):
Please QA this more and resubmit...

Same issues as Module 2.

You likely need fresh diagrams for most of this course. Use our design system on figma and AI

Other formatting issues as well...

kanungle (dismissed):
Re-checked after the latest push. Good progress on the sparse-models depth and the filtering section, that's what I wanted. Still blocking:

1. **Technical error in Step 2:** "Pass a models.Document object and specify the model - Qdrant handles embedding on the server side." With `QdrantClient(":memory:")` there is no server; `models.Document` gets embedded client-side by FastEmbed. Server-side embedding is a separate feature (Qdrant Cloud Inference). Replace with: "Pass a `models.Document` object and specify the model. The client embeds it locally using FastEmbed before upload."
2. **Three References links 404:** `concepts/hybrid_query/` → `/documentation/concepts/hybrid-queries/`; `tutorials/sparse_search/` is gone → link `/articles/sparse-vectors/`; `tutorials/multimodal_rag/` is gone → link `/documentation/advanced-tutorials/multimodal-search-fastembed/`. Make them all site-relative while you're at it, and click every link before resubmitting.
3. **Diagrams, same note as before:** fusion.png and multimodal.png are default mermaid/tool colors, not ours. Rebuild them from the Docs/Diagrams Figma component library using the Qdrant color palette.
4. **"Where We Left Off" tables are malformed markdown** (data in the header row, mismatched columns) and will render broken. Make it a proper table, or better, one small on-brand diagram. Show, don't tell.
5. **`- etc..` bullet in "What's Next - Module 4"** and "Sentence transformers, OpenAI embeddings, etc." in the Multimodal section. We never use "etc." List what Module 4 actually covers, and "Sentence transformers or OpenAI embeddings."
6. **✔/✖ in the head-to-head table**, same note as my emoji comment. Use "Strengths" / "Weaknesses" row labels.
7. **No next-module button at the bottom.** "End of Module 3. Continue to Module 4" needs to be the actual button Essentials uses, not plain text.
8. **Syllabus drift in `_index.md`:** the Module 3 accordion still lists "Real-World Use Cases" (removed, correctly) and omits the new Filtering section. Also drop "Customer Use Cases" from the Module 4 title there.
9. **This PR still carries `_index.md` and all of module 0.** Please remove them so this PR only adds `module-3/`; same request going out on every module PR.
10. **Alt text is just filenames** (`![sparse.png]`). Every image needs descriptive alt text; accessibility requirement.

ESzyszka:
Comments addressed + cross-checked with Neil-review skill

Dylancouzon (dismissed):
Same disclaimers as previous modules. 

## Must fix

1. Section 1's iPhone table doesn't match the Notebook's numbers, needs updating. The Notebook has iPhone 15 first at 1.0000, the page has it third at 0.89.
2. Import the Notebook's catalog into Section 4 so the page and Notebook match. The current single-point example has no `40` or `size 10` in its sparse text, so sparse has nothing useful to match, and the Try It has the same issue.
3. We set up a Cloud cluster in module 2 and drop to `:memory:` here. Would be nice to keep continuity and reuse the collection the reader already has.
4. `fusion.png` could show RRF more clearly. I'd make one candidate appear in both rows so the fused row shows agreement between the two lists, and rename "Mixture" to the term we use elsewhere.
5. Rebuild `nike-example.png` without the Filters column and sync its tokens to whatever example we land on. Section 3 stops at Dense + Sparse now, and the `price <= 200` filter isn't in the module.
6. Multimodal feels out of place in Section 6, module 5 covers it as the capstone. I'd drop it and update module 2's What's Next, which promises it.
7. Section 7's snippets need the missing setup added. `dense_query_vector` and `sparse_query_vector` aren't defined anywhere, a manual sparse query needs `models.SparseVector`, and the page needs an install line dependecies 

## Should fix

1. Taxonomy could use settling. Module 1 says semantic search throughout and module 3 calls it dense search. Same thing with vectors and embeddings being used interchangeably.
2. Section 4's code could use more narration, the blocks arrive without a sentence saying what each does.
3. Quite a lot of detail on BM25 vs miniCOIL vs SPLADE for a beginners course. The prose, the table and `comparison.png` cover the same ground. I'd cut most of it
4. RRF appears in a Section 3 heading and in the diagrams before Section 5 expands it. Worth expanding on first use.
5. Wording could be friendlier for beginners. `prefetch` comes up 17 times without a plain explanation, and posting list, inverted index, term frequency, inverse document frequency and labeled test set all arrive cold. Quite a few repetitions and it's fairly lengthy overall
6. References at the end would work better as "learn more and go further". Most of the current references belong inline where we mention the concept, and a few already are. Same as modules 1 and 2.
7. Section 2 uses `SKU-48291` as a pure exact-match example, then the Key Insight says every real query carries both semantic intent and exact constraints. I'd soften that line, it contradicts the example above it.

## Nice to have

1. Notebook links differ across modules. Module 1 has the Colab badge, modules 2 and 3 use plain GitHub links, and the label changes too.
2. "Key Insight", "How It Works", "Starting Point" and "Why It Matters" are takeaway headings, same as module 1.
3. The local mode note calls the page "this notebook".
4. Alt text on `dense-search.png` and `sparse-search.png` is just the section name. The other four are good.

## Subjective, my opinions, feel free to disagree

1. The page shows no output across eight code blocks while the Notebook shows all of it. Same as module 2.
2. Dense detail in module 2 and sparse here seems like a fine split. Even the depth is uneven, module 2 keeps dense lighter

## Same feedback as earlier modules

- Today's Path is a dry bullet list. A short paragraph TL;DR reads better.
- Retrieval examples change every module. One running example across the course would be nicer.
- No module page has an install line.

ESzyszka:
Changes made to Module 3: 

Screenshot 2026-08-11 at 14 14 43

Dylancouzon:
Feedback is all minor. Structure is great.

- Give it the same messaging, jargon, narration, and verbosity pass as Module 2, using Module 1 as the anchor. Keep it controlled, not a full rewrite.
- Add the same TL;DR used in Section 2. It worked really well.
- The Nike example doesn't really show dense search failing. We explain the limitation well enough, so I'd keep it if re-recording is costly.
- Sparse models go too deep. Simplify the explanation, cut down the table, and link to deeper material. - @kanungle might disagree on that one
- Define terms at first use, e.g. "inverted index" and "query walks."
- "Head-to-Head Comparison" mostly repeats Module 1. I'd reduce it to a paragraph. Feel free to push back.
- Define fusion the first time we mention it, before RRF. Briefly name the other fusion methods and link out, but skip weights and other advanced details already covered in the docs.
- Section 4, Step 1: use `create_payload_index` twice instead of a loop. Before the code, explain what's new and why, specifically the sparse vector config. Afterward, keep only the new information instead of repeating Section 2.
- Remove comments like "Section 6 uses them." We'll explain things when they become relevant.
- Be consistent with "dense embedding" vs. "dense vector." Using both interchangeably may imply they're different things.
- Add inline comments to the hybrid query + fusion code instead of explaining everything afterward.
- Code snippets should show example outputs. The current example also doesn't really demonstrate why hybrid is better, so the payoff lands a little flat.
- The price filter is unindexed.
- Trim Further Reading to the few links most relevant at this stage.
- "Filters in the Wrong Place" repeats the placement mechanism explained just above. Move the indexing note to Step 1 and keep only the local-mode vs. real-server point.

Additional nits from Claude:
- "Not a cleanup pass" contradicts the Common Mistake section, which then describes filtering as a cleanup pass.
- The BM25 description says it "scores only tokens as written," which conflicts with the earlier explanation that BM25 tokenizes and stems.
- The DBSF guidance looks backwards. It recommends DBSF when retrievers are "well calibrated," but DBSF is useful specifically because it normalizes score distributions that aren't directly comparable.

ESzyszka:
Feedback from @Dylancouzon's last comment implemented

qdrant-landing/content/course/beginners/module-3/_index.md
```diff
@@ -0,0 +1,432 @@
...
+#### TL;DR
+```
+Module 2 showed you where your data lives and how Qdrant retrieves it. 
+In this module, you'll learn what that retrieval misses and how to cover the gap. 
+You'll explore dense and sparse vectors, BM25, and the inverted index, then see 
+why a product code defeats either one alone. You'll also learn how fusion merges 
+two ranked lists, and where a filter belongs so both retrievers respect it. 
+By the end, you'll have built a hybrid collection, run a fused query, 
+and filtered it correctly.
+```
```
Dylancouzon:
```suggestion
Module 2 showed you where your data lives and how Qdrant retrieves it. 
In this module, you'll learn what that retrieval misses and how to cover the gap. 
You'll explore dense and sparse vectors, BM25, and the inverted index, then see 
why a product code defeats either one alone. You'll also learn how fusion merges 
two ranked lists, and where a filter belongs so both retrievers respect it. 
By the end, you'll have built a hybrid collection, run a fused query, 
and filtered it correctly.
```
Remove code quotation marks

qdrant-landing/content/course/beginners/module-3/_index.md
```diff
@@ -0,0 +1,432 @@
...
+
+```python
+# BM25 vector for "Nike Pegasus 40 running shoes"
+indices = [1974139272, 24614856, 1784631546, 243905464, 303109060]
+values  = [1.67, 1.67, 1.67, 1.67, 1.67]
+```
+
+Five tokens, five weights, nothing else stored. The indices are hashes of each token rather than positions in a word list, which is why they are large. The values are uniform because BM25 produces only half the score, the part counting how often a token appears. Qdrant computes the other half at query time.
+
+Sparse retrieval does not match characters, which is a common assumption. BM25 tokenizes and stems first, so `SKU-48291` and `SKU-48292` still share the token `sku`. What it gives you is that `40` and `41` are *different tokens* with no relationship at all, where dense placed them 0.0087 apart. The distinguishing token gets its own dimension instead of being averaged away.
```
Dylancouzon:
```suggestion
Sparse retrieval does not match characters, which is a common assumption. M25 first splits text into tokens and cuts each token back to its root. (stemming), so `SKU-48291` and `SKU-48292` still share the token `sku`. <br>
What it gives you is that `40` and `41` are *different tokens* with no relationship at all, where dense placed them 0.0087 apart. The distinguishing token gets its own dimension instead of being averaged away.
```
define stemming

qdrant-landing/content/course/beginners/module-3/_index.md
```diff
@@ -0,0 +1,432 @@
...
+
+Sparse similarity in Qdrant is always the dot product, with no metric to choose, unlike the dense side where you pick Cosine, Dot, or Euclidean.
+
+#### Sparse Models
+
+This module uses **BM25**: statistical, no training needed, and it weights the tokens a text contains without adding any it does not. Two neural alternatives go further. **SPLADE** expands a text with related terms, and **miniCOIL** keeps BM25's vocabulary but weights each token using its context, which is what Qdrant recommends for new projects. Swapping between them is a one-line model change, so start here and read [Understanding SPLADE and Sparse Vectors](/articles/sparse-vectors/) or [miniCOIL](/articles/minicoil/) when you need more.
```
Dylancouzon:
```suggestion
This module uses BM25, a statistical scoring method that requires no training. It scores the terms already present in the text.

Two trained models extend BM25:

- **SPLADE** expands the text with related terms, allowing a document to match a query even when they share none of the same words.
- **miniCOIL** keeps the original terms but weights them based on their surrounding context. We recommend it for new projects.

Switching between these models requires only a one-line change. Start with BM25, then read [Understanding SPLADE and Sparse Vectors](/articles/sparse-vectors/) or [miniCOIL](/articles/minicoil/) when you need more.
```
Break down the paragraph

qdrant-landing/content/course/beginners/module-3/_index.md
```diff
@@ -0,0 +1,432 @@
...
+
+#### Indexing Sparse Vectors
+
+Because most dimensions are zero, Qdrant stores sparse vectors in an **inverted index**: rather than a row per point, it keeps one list per token recording which points contain that token and with what weight. Each list is a **posting list**.
```
Dylancouzon:
```suggestion
Because most dimensions are zero, Qdrant stores sparse vectors in an inverted index: rather than a row per point, it keeps one list per token recording which points contain that token and its weight.
```
Posting list is not a term we use anywhere

qdrant-landing/content/course/beginners/module-3/_index.md
```diff
@@ -0,0 +1,432 @@
...
+
+```text
+Token "nike"    → posting list: [point_1, point_2, point_4, ...]
+Token "pegasus" → posting list: [point_1, point_2, point_3, ...]
+```
+
+A query reads only the posting lists for the tokens it contains, summing weights as it goes, and never touches a point sharing none of them. HNSW from Module 2 is approximate, trading a little accuracy for speed. The sparse index makes no such trade: it is exact.
```
Dylancouzon:
```suggestion
A query reads only the lists for the tokens it contains, summing weights as it goes, so it scores just the points that share at least one token with the query. HNSW from Module 2 is approximate, trading a little accuracy for speed. The sparse index scores every candidate those lists hold, so its ranking is exact.
```
Remove posting list

qdrant-landing/content/course/beginners/module-3/_index.md
```diff
@@ -0,0 +1,432 @@
...
+
+### Dense and Sparse, Side by Side
+
+Module 1 covered the strengths and they hold here. Dense handles synonyms, paraphrasing, and intent, and struggles wherever one distinguishing token carries the meaning: serial numbers, rare words, invented product names. Sparse is the mirror image, strong on exact tokens and domain jargon, and blind to any rewording that shares no words. One caveat on cross-language retrieval: the model used here, `sentence-transformers/all-MiniLM-L6-v2`, is English-only, so it needs the multilingual model Module 1 discussed.
```
Dylancouzon:
```suggestion
Module 1 covered the strengths of dense vectors, and those strengths apply here too. Dense vectors handle synonyms, paraphrases, and intent well, but they can miss cases where one exact token carries the meaning, such as a serial number, rare term, or invented product name. <br>
Sparse vectors make the opposite tradeoff. They work well for exact tokens and domain-specific terms, but they cannot recognize reworded content when the query and document share no words. <br>
One caveat is cross-language retrieval. The model used here, `sentence-transformers/all-MiniLM-L6-v2`, supports English only. For cross-language retrieval, use a multilingual embedding model.
```
Break down the paragraph

qdrant-landing/content/course/beginners/module-3/_index.md
```diff
@@ -0,0 +1,432 @@
...
+
+Dense finds what a query means. Sparse finds what it says. Product search, support search, and code search usually need both at once, which is what the next section is for.
```
Dylancouzon:
```suggestion
Dense finds what a query means. Sparse finds what it says. Most real catalogs carry both, a name a shopper paraphrases and a model number they type exactly, which is what the next section builds.
```
Better to use a concrete example

qdrant-landing/content/course/beginners/module-3/_index.md
```diff
@@ -0,0 +1,432 @@
...
+
+Sparse pushes the 41 down to third, because `40` is a different token from `41`. But it now has dense's problem on a different pair: the men's and women's Pegasus 40 share every token the query contains, so it separates them by 0.0103 on a 3.8396 top score, roughly a quarter of a percent. Dense had those two 0.09 apart and no trouble at all.
+
+Each retriever ranks the right shoe first, and each leaves it a hair ahead of something wrong. Neither is safe alone.
+
+![Hybrid search for the query "Nike Pegasus 40": dense retrieval contributes semantically related running shoes while sparse retrieval locks onto the exact model number, and fusion combines both into one ranked list.](/courses/beginners/module-3/nike-example.png)
+
+### Fusion
+
+**Fusion** is the step that merges two ranked lists into one. It runs on the server once both retrievers finish, and it decides the final order. Qdrant offers two strategies, Reciprocal Rank Fusion and Distribution-Based Score Fusion, and the [Hybrid Queries documentation](/documentation/search/hybrid-queries/) covers their parameters.
```
Dylancouzon:
```suggestion
**Fusion** combines the ranked results from two retrievers into a single list. After both retrievers finish, Qdrant applies fusion on the server to determine the final ranking. <br>
Qdrant supports two fusion strategies: Reciprocal Rank Fusion and Distribution-Based Score Fusion. See the [Hybrid Queries documentation](/documentation/search/hybrid-queries/) for details and available parameters.
```
Not a fan of and, and

qdrant-landing/content/course/beginners/module-3/_index.md
```diff
@@ -0,0 +1,432 @@
...
+    field_schema=models.PayloadSchemaType.INTEGER,
+)
+client.create_payload_index(
+    collection_name="products",
+    field_name="price",
+    field_schema=models.PayloadSchemaType.FLOAT,
+)
+```
+
+Index before ingestion. An index added later still filters correctly, but the structure Qdrant uses to skip non-matching points during a search is built as the data is written, so adding one afterward means a rebuild. Qdrant Cloud also runs **strict mode** by default, a set of guardrails that rejects queries expensive enough to destabilize a cluster, and filtering an unindexed field is one of them: you get a 400 rather than a slow answer.
```
Dylancouzon:
```suggestion
**Index Before Ingestion.** Create payload indexes before you ingest data. Qdrant can add an index later and still filter correctly, but it must rebuild the index for existing points. Creating it first lets Qdrant build the index as it writes the data. <br>
Qdrant Cloud also enables **strict mode** by default. These guardrails reject queries that could be expensive enough to destabilize a cluster. Filtering on an unindexed field is one such case, so Qdrant returns a `400` error instead of running a slow query.
```
More clear

qdrant-landing/content/course/beginners/module-3/_index.md
```diff
@@ -0,0 +1,432 @@
...
+).points
+```
+
+Swap `DENSE_MODEL` for `SPARSE_MODEL` and `using="sparse"` for the sparse equivalent. The placement is identical because there is no prefetch for the filter to belong to.
+
+Whichever retriever you use, the filter admits the same four products, the ones in stock and stocked in size 11: the Pegasus 40 and 41, the Invincible 3, and the Ghost 15. Eligibility does not depend on the retriever. Only the order they come back in does, which is the whole reason the choice of retriever still matters after filtering.
+
+### One Thing to Watch: Local Mode
+
+A hybrid query accepts a filter in either place, and a real server honors both: a top-level `query_filter` does reach the prefetches, so results stay valid either way. Local mode is where they diverge. It ignores the top-level filter and raises no error, so a notebook can print results that break their own filter while looking healthy. Per-prefetch placement behaves the same everywhere, which is reason enough to make it the habit.
```
Dylancouzon:
```suggestion
<aside role="status">
The companion notebook uses local mode with `QdrantClient(":memory:")`. In this mode, Qdrant ignores a top-level filter on a hybrid query instead of returning an error, so the notebook may show results that don’t match the filter.

Filters defined inside each `Prefetch` work consistently in both local mode and Qdrant Cloud.
</aside>
```
More clear

qdrant-landing/content/course/beginners/module-3/_index.md
```diff
@@ -0,0 +1,432 @@
...
+2. Add a `models.FieldCondition` on `price` with `models.Range(lte=140)` to `shopper_filter` and rerun. Two products should drop out. Predict which before you run it.
+3. Query `comfortable shoes for long runs`, a phrase no title contains. Dense ranks the products, while sparse returns a flat tie across the five titles holding both `running` and `shoes`. Work out why sharing exactly the same query tokens produces a tie.
+
+## 7. Beyond Text: Multimodal Search
+
+The same primitive, embed data and search by similarity, applies to any modality: images with CLIP or SigLIP, video as sampled frames, audio as spectrogram embeddings. Qdrant stores whatever vectors your model produces and the retrieval mechanics do not change.
+
+Store them as named vectors on the same point, exactly as this module stored dense and sparse together. Each named vector is its own space, so a query has to be embedded by the model that produced the vectors it is searching: to find images with a text query, embed that text with CLIP's text encoder, not the sentence transformer. Module 5 builds this out as the capstone.
+
+## 8. Knowledge Check
```
Dylancouzon: Hide the answers in `<details>` blocks. Showing them immediately turns the exercise into a recap and removes the need to think.

Ask readers to make a decision, not repeat a sentence from the previous section. Three of the four questions currently test recall. Keep questions that place readers in scenarios where they must apply the module’s concepts:

qdrant-landing/content/course/beginners/module-3/_index.md
```diff
@@ -0,0 +1,432 @@
...
+
+## What's Next - Module 4
+
+Next we go from primitives to judgment, designing a complete vector search system end to end:
+
+- The five layers of a vector search stack: query, indexing, storage, knowledge, and distribution
+- A worked example: designing a news search system, decision by decision
+- Filtering in production: how the query planner combines filters with vector search, and multitenancy via payload filters
+- The production RAG pipeline, from query understanding to generation
+- Deployment options: Local, Docker, Managed Cloud, Hybrid Cloud, Private Cloud, and Edge
```
Dylancouzon:
```suggestion
Eight products can fit in one collection and one query. A multilingual news archive with strict tenant isolation and a generation step requires a more deliberate design. <br>
Module 4 works through this system end to end, decision by decision. You’ll learn how the layers of the stack work together, how filters behave when the query planner selects a strategy, and how deployment choices range from Docker to Qdrant Edge.
```
Better hook, no need to give the full details

Dylancouzon (approved):
Additional notes:

- [x] - `Fusion.png` still doesn’t show a candidate appearing in both lists, and it labels the output “Mixture.” This conflicts with the alt text, which says candidates appearing in both lists rise to the top.
- [x] - Remove `comparison.png`, `sparse.png`, and `multimodal.png` from the PR. Nothing on the page references them anymore.
- [x] - The module is a little lengthy. The margin point appears three times: in Section 1’s table, the following prose, and Section 4’s score table. The TL;DR, Today’s Path, and “By the end, you’ll know” section repeat the same information. Consider condensing Section 7 into an `<aside>` and leaving multimodal search for Module 5.

Approved pending the changes above and inline

Dylancouzon (approved):
Pushed a commit with 2 minor nits and removed an em-dash. 

Two things left, both outside this PR and now unblocked:
- The notebook can be updated
- The Module 3 accordion in module 0 can be updated. It still lists “Beyond Text: Multimodal Search” and is missing the Filtering section.

qdrant-landing/content/course/beginners/module-3/_index.md
```diff
@@ -0,0 +1,447 @@
...
+isLesson: true
+weight: 40
+---
+
+{{< date >}} Module 3 {{< /date >}}
+
+# Sparse vs Dense vs Hybrid Search
+
+<div class="video">
+  <iframe src="https://www.youtube.com/embed/9XXz21jmWes?rel=0" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture; web-share" referrerpolicy="strict-origin-when-cross-origin" allowfullscreen>
```
kanungle: video needs updating with final, branded version that is audio-leveled

qdrant-landing/content/course/beginners/module-3/_index.md
```diff
@@ -0,0 +1,447 @@
...
+  </iframe>
+</div>
+
+Understand dense versus sparse retrieval, their strengths, and how a hybrid approach can combine them.
+
+**Follow-along code**: [Module 3 notebook](https://github.com/qdrant/examples/blob/master/course/beginners/Module3.ipynb)
+
+#### TL;DR
```
kanungle: This is not a TLDR, its an Overview. Either content or section header needs to change (and be consistent across modules)

qdrant-landing/content/course/beginners/module-3/_index.md
```diff
@@ -0,0 +1,447 @@
...
+
+> Module 2 showed you where your data lives and how Qdrant retrieves it. 
+In this module, you'll learn what that retrieval misses and how to cover the gap. 
+You'll explore dense and sparse vectors, BM25, and the inverted index, then see 
+why a product code defeats either one alone. You'll also learn how fusion merges 
+two ranked lists, and where a filter belongs so both retrievers respect it. 
+By the end, you'll have built a hybrid collection, run a fused query, 
+and filtered it correctly.
```
kanungle, line 34: Not sure why all of this is on separate lines

qdrant-landing/content/course/beginners/module-3/_index.md
```diff
@@ -0,0 +1,447 @@
...
+3. Hybrid Search: Dense and Sparse
+4. Setting Up Hybrid Search in Qdrant
+5. Fusion Strategies
+6. Filtering: Works with Any Retrieval Method
+7. Knowledge Check
+8. References & Further Reading
+
+## 1. Where We Left Off
+
+In Module 2 you built a complete ingestion and retrieval pipeline: raw text, vector, store, top-K query. Dense retrieval handles meaning well. It gets shaky on the part of a query that has to be exact.
```
kanungle: add comma after "Module 2"

qdrant-landing/content/course/beginners/module-3/_index.md
```diff
@@ -0,0 +1,447 @@
...
+
+A margin that thin holds across eight products. Across eighty thousand, with every colorway and width in the catalog, something will drift into that one percent and take the top slot. Model numbers, SKUs, and part codes need matching, not neighborhood. That is the gap sparse search fills.
+
+## 2. The Two Families of Search
+
+Every retrieval system is built from one or both of these.
+
+### Dense Search
+
+![Two short phrases encoded into 384-dimensional dense vectors, positioned close together in vector space because they share meaning rather than words.](/courses/beginners/module-3/dense-search.png)
```
kanungle: remove title from diagram and fix branding

qdrant-landing/content/course/beginners/module-3/_index.md
```diff
@@ -0,0 +1,447 @@
...
+cosine("car repair",    "automobile maintenance")   # 0.7334
+cosine("cheap flights", "affordable airfare")       # 0.7241
+cosine("cheap flights", "bake a cake")              # 0.2047
+```
+
+Read the gap, not the absolute number: what makes 0.73 meaningful is the distance from the 0.20, not the value on its own.
+
+### Sparse Search
+
+![A sparse vector drawn as a long mostly-empty row, with weights on only the handful of positions matching tokens present in the text.](/courses/beginners/module-3/sparse-search.png)
```
kanungle: remove title from diagram and fix branding

qdrant-landing/content/course/beginners/module-3/_index.md
```diff
@@ -0,0 +1,447 @@
...
+)
+client.create_payload_index(
+    collection_name="products",
+    field_name="price",
+    field_schema=models.PayloadSchemaType.FLOAT,
+)
+```
+
+**Index Before Ingestion.** Create payload indexes before you ingest data. Qdrant can add an index later and still filter correctly, but it must rebuild the index for existing points. Creating it first lets Qdrant build the index as it writes the data. <br>
+Qdrant Cloud also enables **strict mode** by default. These guardrails reject queries that could be expensive enough to destabilize a cluster. Filtering on an unindexed field is one such case, so Qdrant returns a `400` error instead of running a slow query.
```
kanungle: link to strict mode docs

qdrant-landing/content/course/beginners/module-3/_index.md
```diff
@@ -0,0 +1,447 @@
...
+1.0000  Nike Pegasus 40 running shoes
+0.5833  Nike Pegasus 41 running shoes
+0.5833  Nike Pegasus 40 womens running shoes
+0.4000  Nike Pegasus Trail 4 trail running shoes
+```
+
+Every run in this module put the right shoe first, so the order is not what improved. What changed is the margin: dense and sparse alone each left the right shoe under two percent clear of a rival (Sections 1 and 3), while hybrid with RRF widens that lead to **41.7%** (1.0000 vs. 0.5833). The Pegasus 40 is the only product ranked first by *both* retrievers, and fusion turns that agreement into distance. Fusion scores come from rank position, which is why they resemble neither input scale.
+
+## 5. Fusion Strategies
+
```
kanungle, line 300: section needs a diagram

kanungle (changes requested):
Update to final video, update diagrams, add a diagram, remove `<br>` tags, and see other comments

Content is good, polish is somewhat missing.

kanungle:
Updated diagrams:

dense-search-fixed-labels

sparse-search-restyled

hybrid-fusion-teal-sparse

rrf-fusion-ranking-restyled

kanungle (changes requested):
Looks great. Just needs updated video. Re-request review from me when that's ready and I'll watch / approve

Dylancouzon:
Tightened the module:
- Removed the local-mode aside and its Knowledge Check. It described an undocumented client-side quirk, duplicated the per-prefetch rule already covered above, and distracted beginners from more relevant concepts. The replacement question now covers strict mode and missing payload indexes.
- Rewrote “What’s Next” to remove promises about tenant isolation and query-planner behavior that Module 4 doesn’t cover.
- Added alt text to all four diagrams.

kanungle:
LGTM, just needs updated video then i'll approve

kanungle (approved):

PR #2497, opened 2026-07-10
Beginners course module 4

manas95826:
[PREVIEW](https://deploy-preview-2497--condescending-goldwasser-91acf0.netlify.app//course/beginners/module-4/)


qdrant-landing/content/course/beginners/module-4/_index.md
```diff
@@ -0,0 +1,292 @@
...
+5. Deployment Options
+6. Knowledge Check
+7. References & Further Reading
+
+By the end, you'll be able to reason about any vector search architecture layer by layer, and you'll have designed one system end to end.
+
+## 1. The Layers of the Stack
+
+Every vector search system, from a notebook prototype to a deployment serving millions of queries, is built from the same five layers. When something is slow, wrong, or expensive, the first diagnostic question is always the same: which layer is the problem in?
+
```
kanungle, line 35: Needs a diagram

qdrant-landing/content/course/beginners/module-4/_index.md
```diff
@@ -0,0 +1,292 @@
...
+- **Indexing layer**: The structures that make search fast: the HNSW graph for vectors (Module 2) and payload indexes for filter fields. Mistakes here don't make results wrong. They make everything slow.
+- **Storage layer**: Where points actually live: vectors, payloads, and IDs on disk and in memory. Decisions about what goes in each point's payload, and how big your vectors are, land here.
+- **Knowledge layer**: The data itself and how it's prepared: chunking, embedding model choice, payload schema design. Problems in this layer can't be fixed by tuning any other layer. Garbage in, garbage retrieved.
+- **Distribution layer**: How the system grows beyond one machine: sharding, replication, multi-node clusters. You don't need this on day one. Growing systems eventually do.
+
+### Key insight
+
+Every design decision belongs to a layer. "Add a payload index" is an indexing decision. "Switch to a multilingual embedding model" is a knowledge decision. "Move to three nodes" is a distribution decision. Once you sort decisions into layers, intimidating architecture diagrams become checklists.
```
kanungle, line 46: This is nice

qdrant-landing/content/course/beginners/module-4/_index.md
```diff
@@ -0,0 +1,292 @@
...
+  source        string    (indexed)
+  published_at  datetime  (indexed)
+  summary       string    (not indexed: returned, never filtered)
+```
+
+*(Knowledge and indexing layers.)*
+
+### Question 4: What does the retrieval pipeline look like?
+
+*Dense-only? Hybrid? Reranking?*
```
kanungle, line 138: Reranking is never explained in this course. It should be in Module 2 or 3.

qdrant-landing/content/course/beginners/module-4/_index.md
```diff
@@ -0,0 +1,292 @@
...
+    field_name="published_at",
+    field_schema=models.PayloadSchemaType.DATETIME,
+)
+```
+
+Schema types: `KEYWORD` for exact strings, `INTEGER` and `FLOAT` for numbers, `DATETIME` for timestamps, `GEO` for coordinates, `TEXT` for full-text conditions.
+
+### A Special Case Worth Knowing: Scoping by User or Tenant
+
+One filtering pattern deserves a call-out because you'll meet it in almost any multi-user product: **scoping every query to one user's (or one customer's) data.** The instinct is to create a collection per user. With many users, that becomes thousands or millions of collections, which is operationally unmanageable. The standard pattern instead:
```
kanungle, line 238: point to further reading on multitenancy in docs

qdrant-landing/content/course/beginners/module-4/_index.md
```diff
@@ -0,0 +1,292 @@
...
+2. Create a payload index on it.
+3. Filter on it at every query. Never omit it.
+
+One collection, complete per-tenant isolation at query time, and thanks to filtered traversal plus the index, it stays fast at any tenant count. This is the same mechanism as the country filter in the news system, just applied to identity instead of geography.
+
+### Key insight
+
+Design the payload schema before you ingest, driven by one question: what will I need to filter on? Time, geography, identity, permissions, and status flags are the usual suspects. Adding a payload field later is easy. Discovering at query time that you never stored `language` is not.
+
+## 4. The Production RAG Pipeline
```
kanungle, line 261: The concept of RAG is never overviewed in this course

kanungle (dismissed):
Looks pretty good. I left a few comments, some of which may take effort to address via other modules

kanungle (dismissed):
Content here is strong: the five-questions worked example is exactly the judgment-teaching we want, and the filtering section is the best treatment of filtered traversal we have in any course. Fixes before it ships:

1. **The Question 4 code sample crashes.** `models.Range(gte="2026-07-01T00:00:00Z")` throws `ValidationError`; `Range` only takes floats. For datetime fields it's `models.DatetimeRange(gte="2026-07-01T00:00:00Z")`. Please run every snippet in the module before we ship, this is the flagship example.
2. **"retrieving relevant passages from a vector database"** (Section 4) → "vector search engine". Hard rule. Same fix in the course `_index.md` ("No prior experience with Qdrant or vector databases required").
3. **Diagrams are off-palette.** layers.png and deployment.png use colors from outside the Qdrant color palette, and the type doesn't match our diagram system. Rebuild both from the Docs/Diagrams Figma component library using the Qdrant palette.
4. **This PR carries `_index.md` and module-0 files owned by #2492.** Please remove them so this PR only adds `module-4/` plus its two images; same request going out on every module PR. The syllabus also needs Module 4 and 5 entries in #2492.
5. **Multitenancy is linked to two different URLs** (Section 3 vs References). Pick one canonical page, and make all internal links relative paths rather than absolute `https://qdrant.tech/...`.
6. **Ending:** if the theme's next-module button renders on the preview, drop the "End of Module 4. Continue to Module 5" sentence; if it doesn't render, that's the bug to fix.
7. Title Case: "Key Insight" (three occurrences), "Rule of Thumb".

One structure question: Essentials splits each day into per-lesson pages with a short `_index.md` overview; this module is one 296-line page. Was single-page intentional? If not, splitting sections 1–5 into lesson files like `essentials/day-4/` gets us per-lesson videos and cleaner sidebar nav.

ESzyszka:
Key improvements: 

- Moved the Question 4 filter into each Prefetch; it was at the top level, which post-filters
- FusionQuery(Fusion.RRF) → RrfQuery(rrf=Rrf())
- Added a "Common Mistake: Filters in the Wrong Place" subsection
- Rewrote the filtered-traversal section around the query planner's three per-segment strategies, plus ACORN
- Moved payload index creation into Question 2, in correct build order, with the reason it must precede ingestion
- Replaced Question 3's "nothing to design around" with the bulk-backfill and optimizer-lag decisions
- Explained cardinality estimation as the real cost of a missing index; added strict mode
- Added BOOL, UUID, is_principal, is_tenant=True, geo polygon, and min_should
- Added the fusion-method decision, the fusion-across-shards constraint, and a Local Mode caveat
- Converted the knowledge check to collapsible details cards, rewrote the post-filtering answer, added two questions
- Rewrote the opener, layer bullets, brief passage, Question 1, the Decision line, and the Section 2 Key Insight
- Tagged the schema block yaml and made it valid so it highlights
- Reformatted references, fixed two broken links, added four entries
- Fixed a comma splice, two passives, and two sentence-case table headers
- video after the title 
- Q&A section as drop-down

Dylancouzon (dismissed):
I wasn't part of the decisions behind this module, so this is a read on what the page does to a reader rather than a verdict on the plan behind it. It's opinionated because the issues look structural.

My read is that this is for someone who wants a deep dive into filters internals, while a one day beginners course reader wants to choose a shape for their own system.

**The five design questions never get to do any work.** They're extremely verbose, and we answer two up front. The reader just watches five decisions being made. 

**The depth is set by what Qdrant can do rather than by what a beginner has to choose.** Section 3 covers ACORN, per-segment planner strategies, strict mode, `is_principal`, seven payload index types, the optimizer's indexing threshold, weighted Reciprocal Rank Fusion and Distribution-Based Score Fusion, all before the reader has run a filtered query.

**That depth is also where accuracy gets harder to hold.** Each extra mechanism is one more detail to land. 
[From Claude, not fully verified since rewrite is probably needed]
Section 3 says a top level `query_filter` never reaches the prefetches, and on 1.18.2 and 1.12.0 a prefetch with `limit: 1` still returns the filtered point over HTTP, so it does reach them. Local mode ignores it, which is what the Notebook's BROKEN cell shows, and Module 5 repeats the same reading twice. This needs fixing whatever else changes. 

**The reader never runs anything.** Modules 1 to 3 each end on a Try It. We have eight quiz questions and four code blocks that can't run: no embedding model is named, `size=1024` has no source, the query vectors are undefined, and there's no install line. The Notebook has all the output the page is missing, and its Japanese, Thai and Chinese results are ones most readers can't judge.

I'm not sure what the fix is, since I don't have the scope for this module, but the direction I'd consider is fewer mechanisms and one real decision the reader makes and sees, with deeper filtering moved to an article.

One idea, take it or leave it: instead of one worked example, show a few systems side by side, each answering the five questions differently, each linking the tutorial that builds it. Code search leans on sparse, collaborative filtering has no text at all, multimodal needs named vectors, PDF retrieval is mostly chunking. That way people pick whichever is closest to their own problem and go experiment, rather than following ours to the end.

ESzyszka:
Neil-Review skill output 

**Must Fix (Blocks Shipping)**

- [x] **Every filter demo in the notebook is a silent no-op, and the wrong output is committed.**

  **Why:** cell 20 filters `country=VN AND topic=shipping` and returns SG, JP, CN, TH, and DE, with zero VN results. Cell 22 filters `language in [ja, zh, th, vi]` and returns SG/en first and DE/de fifth. The worst one is cell 24: the tenant demo prints identical results for `europe-desk` and `asia-desk`, so the isolation lesson ships with output proving there is no isolation. The cause is that `search()` passes `query_filter` at the top level and `:memory:` ignores it, which is the exact failure the page's own Local Mode caveat describes.
  **Fix:** move the filter into both `Prefetch` objects inside `search()`, re-run every cell, and recommit the outputs. Check that the two tenant views actually differ before shipping.

- [x] **The page states as fact something the `course-standards` skill records as verified false.**
  **Why:** "A filter passed at the top level, as `query_filter`, therefore never reaches the prefetches." The standards skill says a top-level `query_filter` does reach the prefetches on a real server, verified on 1.12.0 and 1.18.2, and that only local mode ignores it. The claim is load-bearing: it is a named subsection, a Knowledge Check answer, and it is repeated twice in Module 5.
  **Fix:** rewrite "Common Mistake: Filters in the Wrong Place" as a local-mode gotcha rather than an engine behavior. Per-prefetch filters stay the habit worth teaching, but the top-level form is applied by the server, and only a notebook will show it having no effect. Then grep Module 5 for the same sentence.

- [x] **Page and notebook disagree on almost every technical decision in the module.**
  **Why:** the diff that `course-standards` asks for fails on six lines.
  - Page: `size=1024`, no model named. Notebook: `paraphrase-multilingual-MiniLM-L12-v2`, 384 dimensions.
  - Page: `modifier=models.Modifier.IDF`, with a comment saying BM25 requires it. Notebook cell 7: bare `SparseVectorParams()`, so every sparse score in the notebook runs `Qdrant/bm25` without the IDF half.
  - Page: `models.RrfQuery(rrf=models.Rrf())`. Notebook: `models.FusionQuery(fusion=models.Fusion.RRF)`.
  - Page: `is_tenant=True`, described as what keeps the tenant filter fast as tenant count grows. Notebook: a plain keyword index, while the markdown still claims "fast at any tenant count".
  **Fix:** make the notebook the source of truth, run it, then copy its values onto the page. Name the model on the page and drop the unsourced 1024. Add the IDF modifier to cell 7.

- [x] **The follow-along link is a 404, and the notebook is not in the repo.**
  **Why:** `github.com/qdrant/examples/blob/master/Beginner-course/Module4.ipynb` returns 404. The repo has `course/` with `day_1` through `day_5` and no `Beginner-course/` directory. The notebook currently exists only as a Drive file, so it is neither version-controlled nor reviewable, and the standard is that notebook PRs merge before the pages that link them.
  **Fix:** open the notebook PR against `qdrant/examples` first, matching the Essentials path convention, then point the page at the merged path using the same badge and label as the other modules.

- [x] **The video embed will not render.**
  **Why:** the iframe `src` is `https://youtu.be/0qQ3B9uirz0`, and `youtu.be` refuses to be framed. The video itself is fine and already published: "Beginner course | Module 4 | Designing a Vector Search System" on the Qdrant Vector Search channel, so the launch note about a missing video is resolved.
  **Fix:** `https://www.youtube.com/embed/0qQ3B9uirz0?rel=0`. Confirm the outro bumper is present and the intro bumper is not.

- [x] **`description` frontmatter is 168 characters against a 160 limit.**
  **Why:** frontmatter counts are a review item every round, and this one is countable.
  **Fix:** `description: "Design a vector search system: the layers of the stack, five design questions, filtering, a production RAG pipeline, and deployment options."` That is 140 characters. While you are in there, drop the spaced hyphen from `short_description` and from the "What's Next" heading; both read as stand-in em dashes.

## Should Fix

- [x] **Length: 3,680 words of prose, excluding all six code blocks.**
  **Why:** the standard is roughly 900 to 1,500 words for a module. This is the countable version of the structural read in the review above, and it is why the five questions stop doing work. Questions 3 and 4 each carry a second essay inside them.
  **Fix:** cut Section 3 to the two decisions a beginner actually makes, which are index what you filter, and filters are safe to use on every query. Move ACORN, the per-segment planner strategies, strict mode, `is_principal`, `min_should`, and the fusion-plus-formula multi-shard constraint into a filtering article and link it. That gets the module close to range without losing rigor.

- [x] **The reader never runs anything on the page.**
  **Why:** Modules 1 through 3 each end on a Try It. This module has eight quiz questions and four code blocks that cannot execute: `client = QdrantClient(url=..., api_key=...)` passes `Ellipsis`, `dense_vec` and `sparse_vec` are never defined, and there is no install line.
  **Fix:** add one Try It, ideally the thing the notebook now proves. Run the tenant query with the filter in the wrong place, then in the right place, and show both result sets.

- [x] **The notebook runs on `:memory:` while the page tells readers to verify against a real server.**
  **Why:** payload indexes have no effect in local mode, so cell 9's comment about a filtered query being O(n) without an index demonstrates nothing. The fusion scores printed in the notebook (1.0, 0.5, 0.333) are local-mode artifacts and will not reproduce against a cluster.
  **Fix:** switch the notebook to a free Cloud cluster with Colab `userdata` placeholders, which matches the page's own instruction to create a cluster before Module 5. That fixes the filter demos and the payload-index lesson in one move.

- [x] **Two claims are stronger than the docs support.**
  **Why:** "Create a payload index after ingesting, and you have to rebuild the HNSW index to get any benefit from it", and the Section 3 Key Insight, "Adding a payload index later means rebuilding the HNSW graph". The docs and the standards skill both say an index created later still works for filtering and for the planner's cardinality estimate. Only the filter-aware edges need the rebuild.
  **Fix:** "to get the filter-aware edges from it". Separately, ACORN is opt-in through `search_params`; the current wording reads as though Qdrant applies it automatically.

- [x] **Two sentence-level defects, both in the flagship section.**
  **Why:** they are the first prose a reader meets.
  **Fix:** opening paragraph, "learning about reasoning that turns" becomes "the reasoning that turns". Question 1, "You saw this failure in Module 3, while in retrieving exact IDs like 'SKU-48291'" is ungrammatical; cut the clause back to "You saw this failure in Module 3, retrieving exact IDs like `SKU-48291`."

- [x] **Terms a beginner will not know are used without introduction.**
  **Why:** the course promises no prior vector search knowledge, and the standard is that every term is introduced where it first appears.
  **Fix:** introduce or cut segment, cardinality, optimizer, indexing threshold, upsert, backfill, and cross-encoder. Cut ACORN entirely from the page body, since paper acronyms are out of scope for a beginners course, and link the docs section instead.

- [x] **Notebook headings are sentence case where the page uses Title Case.**
  **Why:** the two artifacts sit side by side, and the notebook is also published content.
  **Fix:** "Filtering, Properly", "The Filter Toolbox", "Key Insight", "Rule of Thumb", and the five question headings. Also spell out Reciprocal Rank Fusion on first use in the notebook, the way the page does.

## Also Flagged, Not Blocking

Link text "Multitenancy - Qdrant" is a pasted page title, so use "the multitenancy guide". Notebook cell 28 reads `os.environ["ANTHROPIC_API_KEY"]` where the standard is Colab `userdata`, and it pins a model handle that will date the notebook. Confirm `is_principal` is available on a `DATETIME` index, since the docs example uses `integer`. Once the page previews, open both diagrams, check that no text inside either one says "vector database", and confirm the fills are palette tokens rather than Mermaid or Tailwind defaults.

ESzyszka:
Summary table of all changes that were made to Module 4
Screenshot 2026-08-11 at 12 31 27

Dylancouzon:
I’m keeping the feedback fairly broad for this pass since there’s still some rewriting needed. Once that’s addressed, I can go deeper on the next review.
Tag me once this is applied and I’ll get to it as quickly as I can.

- Follow the same structure as the other modules: TL;DR, Today’s Path, and maybe Where We Left Off if it fits here.
- Give it the same messaging, jargon, narration, and verbosity pass as Module 2, using Module 1 as the anchor. Keep it controlled, but overall this is still too verbose.
- Keep vocabulary consistent, e.g. don’t switch between “chunk” and “passage.”
- The filtering section repeats a lot from Module 3. Trim it to what’s new here.
- Move tenant scoping to the Bonus Module, where it’s already planned.
- Cut the deployment table to Local, Docker, Managed Cloud, and Edge. One sentence + a link is enough for the rest.
- All code needs to be runnable.
- The five-layer framework is introduced as a diagnostic tool, but we never actually use it that way. Add one concrete slow/wrong/expensive scenario and have the learner identify which layer is responsible.
- The knowledge check mostly tests recall. Since the goal is for learners to design a system themselves, give them a new scenario and have them apply the framework rather than explain what we already built.
- The second Try It query tells essentially the same story as the first. Either find an example where dense and sparse behave differently enough to teach something new, or drop the “watch which retriever keeps up” framing.

ESzyszka:
@Dylancouzon's comments addressed

qdrant-landing/content/course/beginners/module-4/_index.md
```diff
@@ -0,0 +1,360 @@
...
+#### TL;DR
+```
+Module 3 gave you hybrid retrieval. In this module, you'll learn how to
+turn the building blocks into a system. You'll explore the five layers of
+a vector search stack and the five questions that turn a brief into a
+design, then see how Qdrant plans a filtered query instead of discarding
+results afterward. You'll also learn what a production RAG pipeline looks
+like and how to pick a deployment mode. By the end, you'll have designed
+a news search system end to end.
+```
```
Dylancouzon:
```suggestion
Module 3 gave you hybrid retrieval. In this module, you'll learn how to
turn the building blocks into a system. You'll explore the five layers of
a vector search stack and the five questions that turn a brief into a
design, then see how Qdrant plans a filtered query instead of discarding
results afterward. You'll also learn what a production RAG pipeline looks
like and how to pick a deployment mode. By the end, you'll have designed
a news search system end to end.
```
Remove code marks

qdrant-landing/content/course/beginners/module-4/_index.md
```diff
@@ -0,0 +1,360 @@
...
+
+## 1. The Layers of the Stack
+
+Every vector search system is built from the same five layers. When something is slow, wrong, or expensive, the first diagnostic question is always the same: **which layer is the problem in?**
+
+- **Query layer**: embedding the query, choosing dense, sparse, or hybrid, fusing results, setting limits. Module 3's territory.
+- **Indexing layer**: the HNSW graph for vectors and payload indexes for filter fields. Mistakes here make things slow rather than wrong.
+- **Storage layer**: vectors, payloads, and IDs on disk and in memory.
+- **Knowledge layer**: the data and how it's prepared, meaning chunking, embedding model choice, and payload schema. Garbage in, garbage retrieved.
+- **Distribution layer**: sharding, replication, and multi-node clusters. You won't need
```
Dylancouzon: Won't need what?

qdrant-landing/content/course/beginners/module-4/_index.md
```diff
@@ -0,0 +1,360 @@
...
+
+Every design decision belongs to one of them: "add a payload index" is indexing, "chunk ang them whole" is knowledge, "move to three nodes" is distribution.
```
Dylancouzon: Not even sure what that means?

qdrant-landing/content/course/beginners/module-4/_index.md
```diff
@@ -0,0 +1,360 @@
...
+
+![The five layers of a vector search stack, from the query layer at the top down through wledge to the distribution layer at the base.](/courses/beginners/module-4/layers.png)
+
+### Diagnose It
+
+Three reports from a running system. Name the layer responsible for each before opening t
+
+1. Searches scoped to one country take four seconds. The same search without the country ds.
```
Dylancouzon: ds?

Dylancouzon (dismissed):
@ESzyszka Stopping here. The sentences barely make sense, it seems parts are missing.

manas95826:
Some changes I think are needed before we ping anyone:

- TL;DR shouldn't be in a box, it should be simple paragraph to maintain consistency throughout the course 
- "The Layers of the Stack" section seems confusing to me, i think it lacks proper explanation or it is just tmi to read, would love your thoughts on this @Dylancouzon 
- in this section's diagram, there are some terms that we haven't defined i think like sharding, replication, multi-node clusters
- I like the "Diagnose It" Section: However, this is not consistent throughout the course, so I'm doubtful if we should include it or not, as this seems kind of off if you see the entire course. 
- (It's my opinion, please feel free to disagree) I think the narration of this course is very different from our other modules, and it is kind of clearly visible that multiple people actually wrote it. And this seems to be a very big step/complex if compared, maybe from the language?

Dylancouzon:
@manas95826 mostly with you. I’d keep "Diagnose it", since the five layers are introduced as a diagnostic framework, but the exercise needs work.

Right now, each scenario gives away its own answer. “Filter” maps directly to “payload indexes for filter fields,” “dense vectors only” to “dense, sparse, or hybrid,” and “memory” to “in memory.” A reader can get 3/3 by matching keywords without understanding the layers. It has the same problem as the previous Knowledge Check: it tests recall, not diagnosis. **Knowledge** and **Distribution** are also missing entirely. 

The layers diagram just repeats the five bullets, I'd keep the diagram and add a better explanations below it instead of the bullets. And maybe that's where we can explain the diagnosis. So that way we could merge two sections in one. 

On complexity, I don’t think length is the issue. It’s 2,558 words vs. 2,187 in Module 2 and 3,052 in Module 3. The issue is voice and readability.
Now that Modules 1–3 are in a good place, I’d use them as the reference and have Claude review this one for narration, verbosity, clarity, and vocabulary without losing technical depth. I’d also have it read as a beginner taking the course and flag what feels confusing, dense, or awkward. Then go through those points manually rather than letting it rewrite the whole thing blindly.

A few other things:
- Question 4 says to embed the headline + lead, but embeds only the headline.
- I’d also remove ACORN. It’s off by default and too deep for a beginner module.
- The page code itself runs cleanly end to end.

Edit: Fixed the paragraph on the diagram

ESzyszka:
@Dylancouzon @manas95826 I would keep the TL;DR section with the > tag as it is now, it is consistent across all of the modules

Dylancouzon:
I like the > tag, on the TL;DR, much better than the blockquote

kanungle:
one updated diagram below. second diagram is okay

stack-layers-restyled

ESzyszka:
layers
Added new graphic from @kanungle , changed the colors to the correct Qdrant branding

kanungle:
I just looked at the beginning and it seemed off. The Overview/intro section is different from the other courses and the diagram wasn't updated to what I sent

ESzyszka:
I changed it to your diagram, the only difference is colors @kanungle

kanungle:
Video has audio cut and leveling issues around 0:50 to 1:10

kanungle:
> I changed it to your diagram, the only difference is colors @kanungle

use the diagram i sent please. The colors of yours with the red make this look like they mean different things than the purple

kanungle:
If you want to use purple, use shades of only purple across all layers. Having some purple and some red/grey is confusing

kanungle:
The section on RAG should be shown as 'optional' since its outside a retrieval system, and it should explain more clearly that the retrieved context goes to an LLM. Diagram useful here as well.

Dylancouzon:
Neil and I went through this together and had some thoughts. I made the edits directly on the branch rather than another round of comments. The worked example was strong. Here is the reasoning behind the changes.

**It needed to teach decisions.** The two exercises tested understanding without teaching the decisions, so both are gone. The five layers are now ordered by the cost of changing each one, which shows why the data layer (Renamed from Knowledge) matters more than the query layer.

**It was covering ground other modules own.** Filtering internals belong in the docs, and Module 2 already teaches filter syntax and payload indexes. Dense and sparse search are Module 3's subject. Removing that overlap took the page from 3,100 to 2,257 words, matching Modules 1 and 2.

**It was not a prerequisite for Module 5.** The capstone teaches payload-indexes-before-ingest from scratch, so this module does not need to introduce it. The module now focuses on the decisions that are expensive to change later: what you embed, which model, chunk size, and payload schema. The running example and the Try It both sit on one of those decisions. Embedding the article body collapses the score gap between the right result and the next one from 0.2655 to 0.1901.

**Reranking moved to the bonus module.** Presented as step three of a RAG pipeline, it looked like a generation technique. Reranking improves any ranked list, so it belongs with the other relevance features.

**The diagrams** now use a single purple ramp and palette tokens, with the text cut back. Every code block runs against a real cluster, and the printed output comes from an actual run.

Three follow-up fixes belong in neighbouring PRs: Module 3's What's Next promises a query-planner section that no longer exists, Module 5 describes this module as covering filtering in depth, and Module 6 says this module introduced formula queries. I can take those.

kanungle:
Thanks @Dylancouzon it looks good to me. @ESzyszka I think this just needs an updated video

kanungle (approved):

PR #2498, opened 2026-07-10
Beginners course module 5

manas95826:
[PREVIEW](https://deploy-preview-2498--condescending-goldwasser-91acf0.netlify.app//course/beginners/module-5/)


qdrant-landing/content/course/beginners/module-5/_index.md
```diff
@@ -0,0 +1,467 @@
...
+
+## 1. Project Overview
+
+Supply chain risk is invisible until it isn't. A factory fire in Vietnam, a labor dispute in Bangladesh, a regulatory change in China: these events appear as signals across dozens of data sources, in different languages, in different modalities, before they become formal incidents. The goal of this project is to make those signals retrievable before the incident.
+
+This is the news search system you designed in Module 4, extended along two dimensions: more modalities (audio and imagery alongside text) and a new capability (clustering signals into risk themes). The five design questions from Module 4 already gave you the blueprint. Now you build it.
+
+- 📰 **Daily Signals**: News, earnings calls, satellite images, factory footage, ingested every 24 hours.
+- 🔍 **Clustering**: Group signals by supplier, topic, and risk theme using vector proximity.
+- 🌏 **Cross-Language Queries**: Ask in English, retrieve what Japanese and Chinese sources are saying.
```
kanungle: not sure about emojis

qdrant-landing/content/course/beginners/module-5/_index.md
```diff
@@ -0,0 +1,467 @@
...
+| Signal source | Modality | Embedding model | Payload fields |
+|---------------|----------|-----------------|----------------|
+| News articles | text | multilingual-e5-large | supplier_id, country, language, date, risk_score |
+| Earnings calls | audio -> text | Whisper + MiniLM | supplier_id, quarter, sentiment, date |
+| Factory footage | video -> frames | CLIP per keyframe | supplier_id, facility_id, alert_type, timestamp |
+| Satellite imagery | image | CLIP / ResNet | supplier_id, facility_id, region, capture_date |
+| Financial filings | text | fin-e5 or MiniLM | supplier_id, filing_type, fiscal_year, currency |
+| Social / forums | text | multilingual-e5-large + BM25 | supplier_id, platform, language, date, severity |
+
+### Multilingual Text: Why multilingual-e5-large
```
kanungle, line 99: Multilingual can be hard, and would be difficult for users to verify

kanungle (changes requested):
Multimodal is good, but multilingual introduces challenges for evaluating search results. We can't expect learners to know multiple spoken languages.

kanungle (changes requested):
Thanks for pushing the series forward. My earlier point stands though: multilingual is still the spine of this capstone (Sections 3 and 6, plus half the knowledge check). The gloss note helps for one hardcoded example, but learners still can't evaluate results in languages they don't read. Keep the multimodal story (text + image + audio, that part is strong) and reduce cross-language to a short "going further" section.

Code bugs to fix either way:
1. **`cross_language_risk_query` ignores its `supplier_id` parameter** and hardcodes `MatchValue(value="SUP-7291")`. Should be `match=models.MatchValue(value=supplier_id)`.
2. **`published_at` is indexed as DATETIME but filtered with `models.Range`**, which only accepts numbers, so the filter fails as written. Use `models.DatetimeRange(gte=yesterday)` here and in `query_supplier_risk`. Also swap `datetime.utcnow()` for `datetime.now(timezone.utc)`; `utcnow()` is deprecated.
3. **`bm25_encode`, `chunk_text`, and `score_risk` are used but never defined.** A beginner copying this hits NameError immediately. Define them (BM25 is available via fastembed's `Qdrant/bm25`) or state clearly they're stubs and link the Colab notebook with the runnable version.

Course mechanics:
4. **Knowledge check has the answers printed under each question**, so there's nothing to do. Turn two or three into hands-on tasks (extend `query_supplier_risk` to filter by `source_type`) and collapse the answers.
5. **When images land, `four-stage.png` should be built from the Docs/Diagrams Figma component library** using the Qdrant color palette so it matches the rest of the site.
6. **This PR carries `beginners/_index.md` and module-0 files owned by #2492.** Please remove them so this PR only adds `module-5/`; same request going out on every module PR.
7. Section 1 opener reads as AI prose: "A factory fire in Vietnam, a labor dispute in Bangladesh, a regulatory change in China: these events appear as signals…" Something plainer lands better: "A factory fire in Vietnam shows up in local news hours before any English wire picks it up."

ESzyszka:
Changes done to Module 5:

Screenshot 2026-08-11 at 13 32 52

Dylancouzon:
Holding this review until module 4 rewrite lands

ESzyszka:
Must fix (blocks shipping)

  - [x] {{< details >}} appears six times and is not a shortcode in your theme. The page will not build.
  Why: Hugo fails hard on an undefined shortcode. The theme ships accordion, banner, card, cards-list, code-snippet, congratulations-banner, course-card, date, grid-row, include, list — no details. Module 4 has the identical problem.
  Suggested fix: native HTML, already used elsewhere on qdrant.tech:
  html <details> <summary>Why does the collection use named vectors instead of one collection per modality?</summary> ... </details>
  Markdown does not render inside raw HTML in Goldmark, so `dense_matrix` becomes <code>dense_matrix</code> and *text* becomes <em>text</em>.
  - [x] §3's table contradicts §4's code on satellite imagery. The table says Satellite imagery produces image only. But ingest_satellite_capture passes capture["caption"] as text, so it also produces text_dense and text_sparse — and the module depends on that: the §4 comment, the §5 clustering, and Knowledge Check Q2 all turn on captioned images having a text_dense vector.
  Why: the table teaches the opposite of the lesson the rest of the module is built on.
  Suggested fix: change that row's "Vectors it produces" to `image`, `text_dense`, `text_sparse` (caption) and add "image + caption" to its Modality cell.
- [x] The Course Summary contradicts your actual Module 0. The table says "Module 0 | Setting Up | Dependencies, a Qdrant Cloud cluster, and a first working connection." The Module 0 file you sent me is Qdrant Beginners Certification at weight: 100, which renders last.
Why: the final page of the course misdescribes the course. Either a Setting Up module exists and I haven't seen it, or this row is describing something that was cut.
Suggested fix: confirm which is real. If the certification page is the only Module 0, drop that row and add a Module 6 row for the bonus module.
- [x] Four British spellings, one of them in alt text: labelled (§1 prose, and inside the four-stage.png alt text), centre twice in §5.
Suggested fix: labeled, center. Re-grep the alt text specifically — it's the surface reviewers skip.
- [x] The notebook link 404s. course/beginners does not exist in qdrant/examples.
Why: same dead path as Modules 3 and 4. Note Module 1 uses a different convention (Beginner-course), which also 404s, so the whole course needs one decision here.
- [x] No video, and two TODO comments ship in the file. The Courses lane requires a video at the top of every module page, and this is the capstone.
Suggested fix: record it, or get the exemption in writing. Either way the TODO comments belong in the PR description, not the published page.

Should fix

- [x] Vocabulary drift on "passage." "score a clearly unrelated passage alongside your real one" uses "passage" where the course now standardizes on "chunk." The literal query: / passage: prefixes must stay — those are e5's API, not prose.
Suggested fix: "score a clearly unrelated chunk alongside your real one."
- [x] None of the ingestion code runs as written. chunk_text, bm25_encode, and score_risk are declared as exercises for the reader, which is honest, but it means the capstone's central pipeline is unexecutable.
Why: course-standards requires tested, runnable code with output shown, and Module 4's review asked for exactly this. A capstone is the worst place to hand over three unimplemented functions.
Suggested fix: at minimum give bm25_encode a real body via FastEmbed (models.Document(text=..., model="Qdrant/bm25"), as Modules 3 and 4 now use), since it's a one-liner and the other two are genuinely pedagogical.
- [x] Today's Path has no closing line. Modules 1, 3, and 4 all end theirs with "By the end, you'll…".
Suggested fix: "By the end, you'll have built the whole system: ingestion, clustering, and analyst queries over one collection."
- [x] One image for a 2,000-word capstone. Module 3 has five. The four-stage diagram is good, but nothing visualizes the one-signal-one-point model, which is the module's central idea.
Suggested fix: one diagram showing a single point with four named vectors hanging off it and a shared payload. Build from the Docs/Diagrams library, N98 #F0F3FA fills with N30 #28324D text, P50 #DC244C accent, Geist Mono for vector names.
- [x] Install line style differs from Modules 3 and 4, which now use a separate ```bash block. Here it's a comment inside the Python block.
- [x] Course Summary titles don't match the real module titles: "Sparse, Dense & Hybrid" vs "Sparse vs Dense vs Hybrid Search"; "Understand Search" vs "Let's Understand Search".
- [x] "Try It" is nested inside §7 Knowledge Check. In Module 3 it sits in a content section. Minor, but it makes the exercise look like a seventh quiz question.

Nice to have

- [x] Centroids aren't unit length. k-means returns the mean of unit vectors, which has magnitude < 1. Qdrant normalizes query vectors for cosine collections so signals_like_cluster is correct — but the module just made a careful point about normalization, so a reader may wonder.
Suggested fix: half a sentence — "the centroid itself isn't unit length, but Qdrant normalizes query vectors for cosine, so it can be passed straight in."
- [x] summary is typed keyword for a 300-character excerpt. It's indexed: false so nothing breaks, but keyword is the exact-match type.

qdrant-landing/content/course/beginners/module-5/_index.md
```diff
@@ -0,0 +1,600 @@
...
+
+#### TL;DR
+
+> Module 4 turned the building blocks into a design. In this module, you'll
+build that design as a working system. You'll explore named vectors that
+hold text, image, and audio evidence on a single point, then see how a
+daily job clusters those signals into the events they describe.
+You'll also learn how to search images with text and how to extend the
+system across languages. By the end, you'll have
+ingested, clustered, and queried multimodal signals from one collection.
```
Dylancouzon:
```suggestion
> Module 4 turned the building blocks into a design. Now you’ll build that design into a working system. You’ll use named vectors to connect text, image, and audio evidence to a single point, then cluster those signals into the events they describe. You’ll also search images with text and extend the system across languages. By the end, you’ll have ingested, clustered, and queried multimodal signals from one collection.
```
Punchier, more clear

qdrant-landing/content/course/beginners/module-5/_index.md
```diff
@@ -0,0 +1,600 @@
...
+6. Analyst Queries
+7. Knowledge Check
+8. Course Summary
+9. References & Further Reading
+
+By the end, you'll have built the whole system: ingestion, clustering, and analyst queries over one collection.
+
+## 1. Project Overview
+
+A factory fire at a supplier's plant shows up in a local news report hours before a wire service picks it up. It also shows up in a satellite image, in an earnings call where an executive gets asked about it, and in a supplier's own filing weeks later. Each of those is a signal, and none of them arrives labeled as an incident.
```
Dylancouzon:
```suggestion
A factory fire at a supplier's plant reaches you four ways. A local news report, a satellite image, an earnings call where an executive gets asked about it, and the supplier's own filing weeks later. You are building the system an analyst uses to see all four, and none of them arrives labeled as an incident.
```
More clear

qdrant-landing/content/course/beginners/module-5/_index.md
```diff
@@ -0,0 +1,600 @@
...
+News articles and filings each get two text vectors, a dense one for meaning and a sparse one for exact tokens: the hybrid pairing from Module 3. `bge-small-en-v1.5` is the dense side, the same model Module 4's design used, at 384 dimensions. It is English-only, and Section 6 covers what to change if your sources are not.
+
+```bash
+pip install "qdrant-client[fastembed]" sentence-transformers openai-whisper transformers torch pillow scikit-learn numpy
+```
+
+The dense text model loads once and is reused at ingestion and at query time:
+
+```python
+from sentence_transformers import SentenceTransformer
```
Dylancouzon: We need to use FastEmbed, as the rest of the modules

Dylancouzon: openai-whisper needs the ffmpeg CLI on the system, so this line alone leaves a beginner with a RuntimeError on the first transcribe. I'd ship the transcripts as fixture strings instead and keep Whisper as a one-line aside, which drops both ffmpeg and torch from the install.

qdrant-landing/content/course/beginners/module-5/_index.md
```diff
@@ -0,0 +1,600 @@
...
+    ingest_signal({
+        "text":         capture["caption"],
+        "image_path":   capture["image_path"],
+        "source_type":  "satellite",
+        **{k: capture[k] for k in ("supplier_id", "facility_id", "country", "published_at")},
+    })
+
+def ingest_earnings_call(call: dict):
+    for chunk in transcribe(call["audio_path"]):
+        ingest_signal({**call, "text": chunk, "transcript": chunk, "source_type": "audio"})
```
Dylancouzon: The same string goes in as both text and transcript, so ingest_signal embeds it twice: bge into text_dense at line 271 and MiniLM into audio_text at line 276. Two 384-dim vectors of identical text in two spaces, on every transcript chunk.

Dylancouzon (dismissed):
Started an inline review, but bigger changes are needed. 

The concepts are right for a beginner’s course. The stack around them isn’t: four embedding models, an ASR step, and manual torch, even though Modules 1 through 4 never went beyond models.Document.

I’d cut the code to text and image only. CLIP through FastEmbed removes torch. Shipping transcripts as fixture strings removes Whisper and its ffmpeg dependency. And `audio_text` currently duplicates `text_dense` anyway. Audio and video can become one “the same pattern extends to” paragraph. Video is only a promise right now: keyframes appear in the Section 3 table and diagram, but there’s no code behind them.

Keep the scroll loop and KMeans. That’s the one thing Modules 1 through 4 don’t teach, and it’s what makes this a capstone instead of a longer Module 4.

Separately, I’d give them `chunk_text` and `score_risk` instead of leaving both as exercises. `score_risk` feeds a payload field that readers filter on two sections later, so leaving it as a stub stalls Section 6.

qdrant-landing/content/course/beginners/module-5/_index.md
ESzyszka: Fixed ✅

ESzyszka: Fixed ✅

kanungle (changes requested):
Video is good. Content is good except two caveats:

1- The intro/overview section isn't consistent with Modules 1-3
2- I feel people with get confused with Qdrant Cloud "Clusters" and the point clusters discussed in this capstone. Make an `aside` note that states these two shouldn't be confused between each other

After these changes, I will approve

Dylancouzon:
Pushed fixes for both review points and the four items from the cross-module audit.

- Reworked the Overview to match Modules 1–3 and removed the duplicate “By the end” line after Today’s Path.
- Clarified the two meanings of “cluster” in Section 5: the Cloud cluster is the Module 2 deployment; clustering here runs in your own code.
- Removed “Common Mistake: Filters in the Wrong Place.” In 1.18.2, a top-level `query_filter` is applied to every prefetch, so both placements return the same scoped results. Local-mode behavior is an undocumented client quirk. Per-prefetch filtering is now taught simply as the placement, consistent with Module 4.
- Removed “Capstone:” from the H1 to match Modules 1–4; renamed TL;DR to Overview.
- Rewrote the Course Summary entries for Modules 4 and 6 to reflect what those modules actually deliver.

kanungle (approved):

PR #2500, opened 2026-07-12
Document per-shard fusion in distributed collections

Dylancouzon:
**Preview:** https://deploy-preview-2500--condescending-goldwasser-91acf0.netlify.app/documentation/search/hybrid-queries/#fusion-in-distributed-collections

Adds a note to the Hybrid Queries page: in a multi-shard collection, a fusion merges globally only when it's the main query. Nested inside a prefetch, it runs per shard on that shard's local results.

```json
// Global: fusion is the main query, so it merges over global positions
{ "prefetch": [ dense, bm25 ],
  "query": { "rrf": {} } }

// Per shard: fusion nested under another main query, so it runs inside each shard
{ "prefetch": {
    "prefetch": [ dense, bm25 ],
    "query": { "rrf": {} }
  },
  "query": { "formula": { /* rescore */ } } }
```

A formula is the common reason to nest a fusion, but it isn't the cause: any nested fusion runs per shard, including under an outer fusion or a vector rescore.

**Why:** documents current behavior (only the top-level fusion is merged across shards) that the docs don't cover today, and that can change ranking on the fusion-under-formula pattern the page itself shows.

New `### Fusion in Distributed Collections` subsection, placed right after the formula example that nests a fusion in a prefetch.

Repro + diagram: https://claude.ai/code/artifact/14e4a5e9-0925-45d7-b928-28a1840e776a


abdonpijpelink (approved):
LGTM. Thanks!

kanungle (approved):

PR #2537, opened 2026-07-26
Update Qdrant Setup video and fix course landing order

manas95826:
- Swap the Qdrant Setup video in the Essentials course, and add the same video to Multi-Vector Search's Qdrant Setup lesson (previously had none)
- Add explicit weight to Essentials and Multi-Vector Search course sections so Essentials always sorts first in the course sidebar, instead of relying on date as a tiebreaker


kanungle (approved):
Thanks @manas95826

PR #2579, opened 2026-08-01
Update neural search tutorial to use FastEmbed and Qdrant Cloud + Notebook Update

manas95826:
[PREVIEW](https://deploy-preview-2579--condescending-goldwasser-91acf0.netlify.app/articles/neural-search-tutorial/)

Need the preview, social_preview image for this.


kanungle (commented):
Thanks @manas95826 for your efforts on this. I think they are good updates, however, I'm thinking we don't update this one. Why: it's one of the first articles by @generall and Qdrant and good to show our legacy in neural search. It becomes a proof point that we've long been a 'search engine' and it has good historical reference for how we've progressed.

@generall - do you agree/disagree?

PR #2504, opened 2026-07-14
Document `max_disk_usage_percent` for Strict Mode

szabosteve:
### [Preivew](https://deploy-preview-2504--condescending-goldwasser-91acf0.netlify.app/documentation/ops-configuration/administration/#maximum-disk-usage)

This PR documents the new `max_disk_usage_percent` strict mode guardrail for `1.19` that rejects disk-consuming write operations once filesystem usage exceeds the configured percentage. Adds a "Maximum Disk Usage" subsection to the Strict Mode admin docs and a runnable code-snippet set across the clients.

PR #2582, opened 2026-08-03
Revert "Document `max_disk_usage_percent` for Strict Mode"

abdonpijpelink:
Reverts qdrant/landing_page#2504. `max_disk_usage_percent` is removed in favor of the new Global quota API in 1.19


szabosteve (approved):
LGTM!

PR #2587, opened 2026-08-03
Link Cloud page customer cards to their case studies

meinsta:
The three customer cards at the top of `/cloud` (Tripadvisor, Deutsche Telekom, FAZ) were static. Each now links to its case study post.

| Card | Links to |
|---|---|
| Tripadvisor | `/blog/case-study-tripadvisor/` |
| Deutsche Telekom | `/blog/case-study-deutsche-telekom/` |
| FAZ | `/blog/case-study-faz/` |

Reuses the existing `link link_dark link_md` style from the resources section further down the same page, the one on "Talk to Engineering". No new link styles.

### Implementation notes

- **The whole card is the anchor**, and the "Read Case Study" label is a `<span>`, not a nested `<a>`. An anchor inside an anchor is invalid HTML and browsers handle it inconsistently; this way the card is a single tab stop and a single link for SEO, while the clickable area covers the whole card.
- Needed `text-decoration: none` on the card now that it is an `<a>`, otherwise the anchor's underline is inherited by all the card text.
- Link text lives in the front matter, so it is editable without touching the template.
- The arrow animates on hover of the label rather than hover of the whole card. `.link`'s animation is keyed to `.link:hover`, and propagating it from the card would mean re-declaring the `::before`/`::after` gradient rules and outweighing the existing `.link:not(:hover)` rule. The card already has its own box-shadow hover. Happy to add it if you want the arrow moving on card hover.

### Verified

Local Hugo build: all three cards render as anchors with the correct hrefs, all three target pages exist in the output, and both new CSS rules compile.

### DIFF
--- a/qdrant-landing/content/qdrant-cloud/qdrant-cloud-customers.md
+++ b/qdrant-landing/content/qdrant-cloud/qdrant-cloud-customers.md
@@ -10,6 +10,9 @@ cards:
     description: Travelers using the generative AI experience show 2-3x more revenue than those using traditional search.
     category: CONSUMER-FACING AI APP
     label: 11M BUSINESSES · 21 COUNTRIES
+    link:
+      text: Read Case Study
+      url: /blog/case-study-tripadvisor/
   - id: 1
     icon:
       src: /img/customer-logo/telekom.svg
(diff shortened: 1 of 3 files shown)


daniel-azoulai (commented):
Approved.

daniel-azoulai (approved):

PR #2588, opened 2026-08-03
Track clicks and scroll depth on the /cloud landing

meinsta:
## Why

`/cloud` had 5 tracked elements, all inherited from the shared header, against ~68 links and buttons of its own. Nothing in the page body reported to Segment/GA4 — no CTA clicks, no card clicks, no tab interactions, no scroll depth.

While tagging it I found a bug in the shared click handler that affected every page using `data-metric-loc`.

## Commits

1. **`fix(analytics)`: read click metrics from `currentTarget`, not `target`** — `handleClickInteraction` read `data-metric-loc`/`data-metric-label` off `event.target`, which is the innermost node clicked rather than the tagged element. Any anchor wrapping child elements reported `location: ''` and a label scraped from whichever ``/`<p>` the user hit, and the dedupe `removeEventListener` ran against the wrong node so it never deduped. 37 of 46 anchors on `/cloud` are that shape. Existing text-only nav links are unaffected (`currentTarget === target`).

2. **`feat(analytics)`: track every link and button on `/cloud`** — hand-tagging ~68 elements would have meant ~70 attributes across 13 files and would still miss anything added later, so this tracks by delegation. `trackAllClicksIn(root)` resolves the element with `closest()`, derives `location` from the nearest `section`/`header`/`footer` class (`qdrant-cloud-hero` → `cloud_hero`), and skips anything carrying `data-metric-loc` so existing listeners stay authoritative and nothing double-counts. Adds `href` + `outbound` to the payload and widens the label fallback to `aria-label`/`img[alt]`/`title`/`href` so icon-only links stop reporting empty labels.

3. **`feat(analytics)`: scroll depth at 25/50/75/90** — GA4 Enhanced Measurement only reports a single 90% milestone. Each milestone fires once per page load, applied in ascending order so one large jump still reports every threshold it crossed. Coalesced to one measurement per animation frame.

4. **`feat(analytics)`: section views, max scroll depth and time-to-milestone** — three additions that all reuse the existing `interaction` schema, so they need no Segment mapping beyond what the milestones already require:
   - **`trackSectionViews()`** reports which sections a visitor actually reached, which a page-level percentage only implies — on a 9-section landing page "50% scrolled" is a proxy for "saw the capabilities section", and an `IntersectionObserver` measures that directly. Fires `action: 'viewed'` once per section at a 25% threshold, reusing the `location` vocabulary already derived for clicks, with the section heading as label. 10 sections report on `/cloud`; OneTrust consent markup and unclassed wrappers are skipped so they don't create buckets named after their tag.
   - **`action: 'scroll_max'`** on `pagehide`/`visibilitychange` reports the deepest point reached, which is the only depth we get from visitors who leave without crossing a milestone. The scroll listener therefore stays attached after the last milestone, since deepest still needs updating.
   - **`ms_to_reach`** on each milestone, via `performance.now()`, separating fast skimming from actual reading.

## Event vocabulary

One event name, `interaction`, with `action` distinguishing the kind:

| `action` | fires when | carries |
|---|---|---|
| `clicked` | any link or button | `location`, `label`, `href`, `outbound` |
| `scrolled` | 25/50/75/90% crossed | `location: page`, `percent`, `ms_to_reach` |
| `scroll_max` | page hidden or unloaded | `location: page`, `percent` (deepest), `ms_to_reach` |
| `viewed` | a section becomes 25% visible | `location` (section), `label` (heading) |

## Scope

Bound from `qdrant-cloud.js`, which Hugo only loads for section `qdrant-cloud`. **Segment event volume on the rest of the site is unchanged.** Promoting this sitewide is moving three lines into `index.js` — worth doing deliberately, since ~2/3 of the clickables on any page are header/footer chrome.

## Verification

Tested in the browser against the built bundle, and end-to-end on the deploy preview: **`POST https://events.eu1.segmentapis.com/v1/t` → HTTP 200**, payloads carrying correct `location`, `label`, `href`, `outbound`.

| case | result |
|---|---|
| click the `<p>` inside a customer card (previously broken) | `cloud_customers` / "2-3x Revenue Lift" |
| click the `<h3>` inside a capability feature | `cloud_capabilities` / "Hybrid Search" |
| click the `` inside a dev-experience card | `cloud_dev_experience` / "Language SDKs" |
| header + footer `contact-us` links | both fire, `site_header` / `footer` |
| repeat clicks on tabs | fires every time |
| the 5 explicitly tagged elements | no double-counting |
| gradual scroll 10→95% | `25,50,75,90` once each, in order |
| scroll to top then bottom again | 0 duplicates |
| single jump to bottom | all four, ascending |
| page shorter than viewport | silent |
| every milestone carries `ms_to_reach` | ✅ |
| `scroll_max` once-guard | two `visibilitychange` events → exactly one event |
| section `location` + `label` derivation | all 10 clean and readable |
| `IntersectionObserver` fires and reports | verified live in a focused tab, 0 duplicates |
| OneTrust / unclassed sections | 4 skipped, no junk buckets |

**`IntersectionObserver` firing is verified** on the deploy preview in a focused tab: six sections reported live with correct locations and readable headings, no duplicates — `cloud_customers`, `cloud_dev_experience`, `pricing_banner`, `cloud_resources`, `cloud_faq`, `cta_banner` — alongside all four scroll milestones with plausible `ms_to_reach` values from real scrolling. A direct probe confirmed delivery for `cloud_hero` at `intersectionRatio` 0.797. The remaining sections fire at page load, before instrumentation can wrap `analytics.track`, and are then `unobserve`d — correct behaviour, simply not observable from a console attached after load.

Each commit builds standalone (`hugo` exit 0). Rebased on latest `master`.

## Config status for GA4 visibility

The website property is **"Qdrant landing"** (`p271085741`, stream `qdrant.tech`, `G-NZYW2651NE` — matching `config.toml:136`).

**Done — custom dimensions registered** (4 Aug 2026), all Event-scoped:

| Dimension name | Parameter |
|---|---|
| Interaction location | `location` |
| Interaction label | `label` |
| Interaction action | `action` |
| Scroll depth percent | `percent` |
| Interaction outbound | `outbound` |

**Done — custom metric registered**, Event-scoped, unit Milliseconds:

| Metric name | Parameter |
|---|---|
| Scroll time to milestone | `ms_to_reach` |

Registered ahead of the mapping on purpose: custom definitions do not backfill, so they must exist before events start arriving. `href` is deliberately not registered — URLs are high-cardinality and would collapse into an `(other)` bucket.

**Blocked — Segment mapping.** On the production source (write key `eQYi1nZE…`), the GA4 destination needs a mapping triggered on `event = "interaction"`, branching on `action` for nicer GA4 event names:

| `action` | GA4 event name |
|---|---|
| `clicked` | `cta_click` |
| `scrolled` | `scroll_depth` |
| `scroll_max` | `scroll_max` |
| `viewed` | `section_view` |

Segment Actions destinations forward nothing without an explicit trigger, which is why GA4 currently reports exactly 0 `interaction` events over 28 days rather than a partial count. This needs someone with access to the production Segment workspace.

Splitting event names in the mapping rather than in code is deliberate: renaming in code would not remove the mapping requirement, and would drop `/cloud` out of Segment-side `interaction` analysis, which is currently the only surface receiving this data.

**The parameter names above must pass through the mapping unrenamed.** If the mapping renames `location`, the registered dimension stays permanently empty and the mismatch is invisible until someone builds a report.

**Deferred to merge day — Enhanced Measurement "Scrolls".** It is currently **on** for this property (152k `scroll` events in 28 days). Leave it on until the mapping is live and this PR is merged, then switch it off. Flipping it early would leave no scroll data at all in the gap; flipping it late means GA4's `scroll` (90% only) double-counts against this PR's `scrolled` (25/50/75/90).

Also worth knowing: Enhanced Measurement **Outbound clicks** is already on, so GA4 already captures outbound navigation via its own `click` event. The `outbound` flag here is not net-new at the GA4 layer — it adds outbound *within the Segment `interaction` schema, carrying section context* (`location: cloud_hero`), which GA4's `click` does not have.

Form-submit tracking is out of scope: `/cloud` has no form, "Talk to Engineering" hands off to `/contact-us/`.

## Note on CI

`linkChecker` was failing for a pre-existing reason unrelated to this PR (a link to a drafted article, see the comment below). That was fixed on `master` in #2590 and this branch has been rebased on top of it.

### DIFF
--- a/qdrant-landing/themes/qdrant-2024/assets/js/qdrant-cloud.js
+++ b/qdrant-landing/themes/qdrant-2024/assets/js/qdrant-cloud.js
@@ -1,8 +1,9 @@
-import { trackAllClicksIn } from './segment-helpers';
+import { trackAllClicksIn, trackScrollDepth } from './segment-helpers';
 
 // ponytail: /cloud only, so Segment volume elsewhere is untouched.
 // Move to index.js to take it sitewide.
 trackAllClicksIn(document);
+trackScrollDepth();
 
 // Capabilities
 (function initCapabilities() {
--- a/qdrant-landing/themes/qdrant-2024/assets/js/qdrant-cloud.js
+++ b/qdrant-landing/themes/qdrant-2024/assets/js/qdrant-cloud.js
@@ -1,3 +1,9 @@
+import { trackAllClicksIn } from './segment-helpers';
+
+// ponytail: /cloud only, so Segment volume elsewhere is untouched.
+// Move to index.js to take it sitewide.
+trackAllClicksIn(document);
+
 // Capabilities
 (function initCapabilities() {
   const stickyWrap = document.querySelector('.qdrant-cloud-capabilities__tabs-sticky');
--- a/qdrant-landing/themes/qdrant-2024/assets/js/qdrant-cloud.js
+++ b/qdrant-landing/themes/qdrant-2024/assets/js/qdrant-cloud.js
@@ -1,9 +1,10 @@
-import { trackAllClicksIn, trackScrollDepth } from './segment-helpers';
+import { trackAllClicksIn, trackScrollDepth, trackSectionViews } from './segment-helpers';
 
 // ponytail: /cloud only, so Segment volume elsewhere is untouched.
 // Move to index.js to take it sitewide.
 trackAllClicksIn(document);
 trackScrollDepth();
+trackSectionViews();
 
 // Capabilities
 (function initCapabilities() {
--- a/qdrant-landing/themes/qdrant-2024/assets/js/segment-helpers.js
+++ b/qdrant-landing/themes/qdrant-2024/assets/js/segment-helpers.js
@@ -28,32 +28,81 @@ const nameMapper = (url) => { // Mapping names based on pathname for Segment
 /***************/
 /* DOM helpers */
 /***************/
-const handleClickInteraction = (event) => {
-  // currentTarget, not target: the tagged element is where the listener was attached,
-  // while target is the innermost node clicked (an <img>/<p> inside a card anchor).
-  const el = event.currentTarget;
-  const rawLabel = el.getAttribute('data-metric-label') ?? el.innerText;
-  const cleanedLabel = rawLabel ? rawLabel.replace(/\s+/g, ' ').trim() : '';
+const LABEL_MAX = 120;
 
-  const payload = {
-    ...PAYLOAD_BOILERPLATE,
-    location: el.getAttribute('data-metric-loc') ?? '',
-    label: cleanedLabel,
-    action: 'clicked'
-  };
+// Fall back to the nearest landmark when an element carries no data-metric-loc,
+// so untagged clicks still land in a readable bucket ('cloud_hero', 'footer', ...).
+const deriveLocation = (el) => {
+  // No 'nav' here on purpose: a <nav> nested in the footer would otherwise split
+  // footer clicks into its own bucket.
+  const region = el.closest('section, header, footer');
+  if (!region) return 'page';
 
+  const cls = [...region.classList].find((c) => c && !/^(col|row|g|container)([-_]|$)/.test(c));
+  return (cls || region.tagName.toLowerCase())
+    .replace(/^qdrant-cloud-/, 'cloud-')
+    .replace(/[^a-z0-9]+/gi, '_')
+    .toLowerCase();
+};
+
+// || not ??: an icon-only link has innerText '' and needs to fall through to the image alt.
+const deriveLabel = (el) =>
+  el.getAttribute('data-metric-label')
+  || el.getAttribute('aria-label')
+  || el.innerText
+  || el.querySelector('img[alt]')?.getAttribute('alt')
+  || el.getAttribute('title')
+  || el.getAttribute('href')
+  || '';
+
+const emitInteraction = (payload) => {
   // If consented to tracking the track
-  if(getCookie('cookie-consent')) {
+  if (getCookie('cookie-consent')) {
     trackInteractionEvent(payload);
-
-    // If element can be clicked more than once (ie user remains on same page)
-    if (!el.hasAttribute('data-metric-keep')) {
-      el.removeEventListener('click', handleClickInteraction);
-    }
   } else { // If no consent yet the store in sessionStorage in case of later consent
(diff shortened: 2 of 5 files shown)


meinsta:
### The failing `linkChecker` check is pre-existing, not from this PR

`Links` has been red on `master` since well before this branch existed, with the **same single error**:

```
🚫 Errors  1
### Errors in qdrant-landing/public/course/essentials/day-7/quotient/index.html
* [404] http://localhost:1314/articles/rapid-rag-optimization-with-qdrant-and-quotient/
```

Timeline:

| when | what |
|---|---|
| Jul 31 | `Links` green on `master` |
| Aug 3, 14:35 UTC | `02aae557` sets `draft: true` on `content/articles/rapid-rag-optimization-with-qdrant-and-quotient.md` |
| Aug 3, 14:52 UTC | `706a241f` removes one referrer, `content/blog/qdrant-relari.md` |
| Aug 3, 15:03 UTC | first red `Links` run on `master` |
| Aug 3, 19:42 UTC | this branch pushed |

This PR changes 2 JS files and 3 `qdrant-cloud` partials. It adds no links and touches no article or course content, so it just inherits the red check.

@chadha-sridi — the referrer cleanup in `706a241f` caught `qdrant-relari.md` but there's a second one still pointing at the now-draft article:

**`qdrant-landing/content/course/essentials/day-7/quotient.md:121`**
```
- [Optimizing RAG Through an Evaluation-Based Methodology](/articles/rapid-rag-optimization-with-qdrant-and-quotient/):
```

Since `draft: true` stops Hugo rendering the article, that link 404s. Dropping or re-pointing it should get `Links` green on `master` again — and unblock the check on every open PR, not just this one. Happy to push that one-liner separately if you'd rather not context-switch.

chadha-sridi:
Thanks @meinsta for flagging this. I removed the broken link in `content/course/essentials/day-7/quotient.md` (PR #2590). Once the branch is updated with the latest master, the link checker should pass.

generall:
Have you measured the performance impact of those chagnes?

meinsta:
> Have you measured the performance impact of those changes?

@generall YKW I only checked empirically, but went ahead and measured, and the cost is analytics volume, with no measurable effect on paint, layout stability, or frame rate.

daniel-azoulai (approved):

PR #2462, opened 2026-06-29
Create agent-friendlier landing page for Learn

szabosteve:
### Previews 

* [Learn index](https://deploy-preview-2462--condescending-goldwasser-91acf0.netlify.app/learn/index.md)
* [Qdrant articles](https://deploy-preview-2462--condescending-goldwasser-91acf0.netlify.app/articles/index.md)

This PR populates the MD version of the Learn landing page with links to the respective sub-pages (Articles, Courses, Tutorials). It also creates a Hugo layout for dynamically list articles at `/articles/index.md`.

PR #2594, opened 2026-08-04
Agent-readable Markdown for the marketing pages

meinsta:
Marketing pages already have `.md` URLs. `curl https://qdrant.tech/cloud/index.md` returns 200 `text/markdown` today, but the body is only the page title. The docs Markdown output renders `.RenderShortcodes`, and marketing pages keep their copy in the front-matter params of headless section bundles, so there is nothing for it to render. Every marketing landing page comes out at 250 to 330 bytes.

Two gaps, both fixed here:

**Empty bodies.** `partials/md-params.txt` walks a bundle's params recursively and renders the copy: keys that read like copy are kept, image/icon/bookkeeping keys are dropped, nested lists become headings and bullets. `partials/md-marketing-page.txt` concatenates a section's bundles, hero first, plus any page bodies. Each page gets a one-line `list.markdown.md` that calls it. `/cloud` goes from 282 bytes to 18.8KB. 33 pages are covered.

**Discovery.** `llms.txt` only walked `RegularPages`, so no section landing page was listed. `/cloud`, `/pricing`, `/edge` and `/customers` were absent entirely. A `## Pages` group now lists all 40, including the four industry pages, which needed `GetPage` because their parent section sets `build.list: never`.

Two pages are hand-written, because the generic renderer flattens what they are actually for:

- **Pricing.** Its value lives in per-tier cells (tier by feature), which only survive as tables. Writing that template surfaced a bug in it, caught before merge: `pricing/list.html` renders `qdrant-pricing-doors-b`, and the Markdown template was reading `doors-a`, a retired A/B variant. The HTML page was always correct; the Markdown published tier cards the site no longer shows, dropped the live ones, and flattened the Support and Reliability matrix to bare row names. Tables now come from a shared `md-tier-table.txt` partial driven by tiers by sections, so both matrices render.
- **Customers.** `customers/list.html` renders two bundles, `customers-hero` and `clients`. The generic rollup walked `.Pages`, which still holds 18 bundles the page stopped rendering (`case-study-*`, `case-studies-1/2`, `logo-cards-*`, `testimonial-*`, `vector-space-wall`), and republished them, duplicating nine case studies. The client directory now renders as a table so the facets the page filters on stay queryable.

`/rag` had the same shape of problem across pages rather than within one: it rolled up the `rag-evaluation-guide-*` bundles, which build `/rag/rag-evaluation-guide/`. A bundle that overrides the section cascade with `render: always` is a page in its own right, so it and the blocks named after it are excluded. That page rendered from `_default` at 27 words and now rolls up its own sibling blocks.

`automation/check-markdown-output.sh` runs in the Links workflow, which already builds the site. It fails if a page stops rolling up (a broken template falls back to the title, which the word count catches), if HTML escaping or `<span>`/`<br>` leak into the copy, if internal links are not rewritten, if pricing loses its tier tables, if a bundle `pricing/list.html` renders does not reach the Markdown, if the client table stops matching the source, or if `llms.txt` stops listing the pages.

### Relation to #2428

Same goal, different approach. #2428 hand-maps each page's params, about 57 lines per page. That PR also shows the maintenance cost: it reads `qdrant-cloud-bento-cards` and `qdrant-cloud-features-link`, both deleted in #2542 when the Cloud page was restructured, so two of its six blocks render nothing and nothing fails. `/cloud` builds to 182 words through that template today, against 2465 through this one, with a `[<no value>](<no value>)` where `contactUs` used to be.

The generic renderer follows content restructuring on its own, at the cost of output that reads less deliberately than hand-written mapping. #2428's `qdrant-vector-database` template is better than the generated version of that page: it stitches `content` + `contentLink` + `contentSecondPart` into one sentence where this leaves a stray `available.`, and it orders blocks the way the page does. Happy to go the other way on any page where that trade looks wrong. A section can override with its own `list.markdown.md`, which is what pricing and customers do, so both approaches coexist.

### Notes for review

- Generated Markdown is only as good as the params. `/private-cloud` renders `# private-cloud` with the body `private-cloud`, because `content/private-cloud/_index.md` has that as its literal title and description. `/customers`, `/partners` and `/events` use their title as their description. These are the H1 and the `llms.txt` summary now, so they are the first thing an agent reads. Left for whoever owns that copy.
- Three content fixes are included. `content/pricing/_index.md` read `Solutions  Qdrant`, a double space where the site's `|` separator belongs, which was wrong in the HTML `<title>` too. One client in `content/customers/clients/_index.md` had `locationse` instead of `location`, which also left it out of the location filter on the live page. `content/get_anonymous_id/_index.md` gets `sitemapExclude`, since it is the iframe endpoint for the Cloud console rather than a page, and it was appearing in both the page list and the sitemap. Its HTML still builds unchanged.
- `/contact-us` and `/subscribe` stay title-only, being forms with nothing to render. `/learn` has its own agent page from #2462.
- Known rough edges, documented in the templates: block order follows filename rather than the order the page renders in; the one stray `available.` on `/qdrant-vector-database`; four duplicate lines across all 33 pages, all of them content that genuinely repeats, such as the same testimonial author twice on `/e-commerce`.
- Built and verified locally on Hugo 0.164. CI builds the pinned 0.160.1.

### DIFF
--- a/automation/check-markdown-output.sh
+++ b/automation/check-markdown-output.sh
@@ -9,18 +9,65 @@ set -uo pipefail
 PUBLIC=${1:-qdrant-landing/public}
 fail=0
 
-# Still title-only, deliberately: /contact-us and /subscribe are forms with
-# nothing to render. /learn has its own agent page (PR #2462).
-PAGES=(
-  cloud qdrant-vector-database pricing hybrid-cloud private-cloud edge
-  cloud-inference ai-agents advanced-search recommendations
-  data-analysis-anomaly-detection rag use-cases healthcare
-  enterprise-solutions qdrant-for-startups customers partners benchmarks
-  about-us lp/lucene e-commerce hr-tech legal-tech hospitality-and-travel
-  rag/rag-evaluation-guide
-  community stars events vector-space-day-sf-26 vector-space-day-sf-26-recap
-  brand-resources demo
+# The pages to check are whichever ones have a Markdown template, so adding or
+# removing a page needs no edit here. A list.markdown.md means the section
+# landing page; a single.markdown.md means the pages in that section that build
+# themselves (build.render: always), which is how the industry pages and the RAG
+# evaluation guide work. The URL comes from the page's own url: override, or the
+# section name when there isn't one.
+#
+# Still title-only, deliberately, and so absent from this list: /contact-us and
+# /subscribe are forms with nothing to render. /learn has its own agent page
+# (PR #2462).
+PAGES=()
+while IFS= read -r page; do PAGES+=("$page"); done < <(python3 - <<'PY'
+import glob, os, re
+LAYOUTS = "qdrant-landing/themes/qdrant-2024/layouts"
+CONTENT = "qdrant-landing/content"
+
+def front_matter(path):
+    try:
+        text = open(path).read()
+    except OSError:
+        return ""
+    m = re.match(r"^---\n(.*?)\n---", text, re.S)
+    return m.group(1) if m else ""
+
+def own_render(fm):
+    """render: from the page's own build block, ignoring any cascade block."""
+    m = re.search(r"^build:\n((?:[ \t]+.*\n)+)", fm + "\n", re.M)
+    if not m:
+        return ""
+    r = re.search(r"render:\s*(\w+)", m.group(1))
+    return r.group(1) if r else ""
+
+def url(fm, fallback):
+    m = re.search(r'^url:\s*"?([^"\n]+)"?', fm, re.M)
+    return (m.group(1) if m else fallback).strip().strip("/")
+
+pages = []
+for template in sorted(glob.glob(f"{LAYOUTS}/*/list.markdown.md")):
+    section = os.path.basename(os.path.dirname(template))
+    fm = front_matter(f"{CONTENT}/{section}/_index.md")
+    if own_render(fm) == "never":
+        continue  # a container for other pages, not a page itself
+    pages.append(url(fm, section))
+for template in sorted(glob.glob(f"{LAYOUTS}/*/single.markdown.md")):
+    section = os.path.basename(os.path.dirname(template))
+    for page in sorted(glob.glob(f"{CONTENT}/{section}/*.md")):
+        if page.endswith("_index.md"):
+            continue
+        fm = front_matter(page)
+        if own_render(fm) == "always":
+            pages.append(url(fm, os.path.basename(page)[:-3]))
+print("\n".join(p for p in dict.fromkeys(pages) if p))
+PY
 )
+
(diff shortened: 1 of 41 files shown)


meinsta:
Update: the pricing template was reading `qdrant-pricing-doors-a`, a retired A/B variant, while `pricing/list.html` renders `doors-b`. The Markdown was reporting tier cards the site no longer shows, dropping the live ones, and flattening the entire Support & Reliability matrix to bare row names. Fixed in 87d6bd3 — tier tables now come from a shared `md-tier-table.txt` partial driven by tiers × sections, so both the feature matrix and the support matrix render.

The CI check now derives which bundles `list.html` actually uses and asserts their headline copy (`title`, `pricing`, `price`) reaches the Markdown. Verified it fails on exactly this bug by pointing the template back at doors-a:

```
pricing/index.md is missing copy from /pricing/qdrant-pricing-doors-b: "Free forever"
pricing/index.md is missing copy from /pricing/qdrant-pricing-doors-b: "Usage-based pricing"
pricing/index.md is missing copy from /pricing/qdrant-pricing-doors-b: "Minimum spend required"
```

generall:
why agents would care about marketing materials?

automation/check-markdown-output.sh
```diff
@@ -0,0 +1,83 @@
...
+PUBLIC=${1:-qdrant-landing/public}
+fail=0
+
+PAGES=(
+  cloud qdrant-vector-database pricing hybrid-cloud private-cloud edge
+  cloud-inference ai-agents advanced-search recommendations
+  data-analysis-anomaly-detection rag use-cases healthcare
+  enterprise-solutions qdrant-for-startups customers partners benchmarks
+  about-us lp/lucene e-commerce hr-tech legal-tech hospitality-and-travel
+)
```
generall: who's going to maintain this list?

generall (changes requested):
> Marketing pages already have .md URLs — curl https://qdrant.tech/cloud/index.md returns 200 text/markdown today — but the body is only the page title. 

I would prefer to not generate index.md for those pages instead.

meinsta:
> why agents would care about marketing materials?

project collab with @abdonpijpelink

automation/check-markdown-output.sh
meinsta: Fixed so that this is not hardcoded

meinsta:
> why agents would care about marketing materials?

We already know buyers arrive to our page via LLM (chatbot) use. If a buyer asks an agent "is Qdrant HIPAA-eligible and what does a production cluster cost", our current /prcing/index.md only has a title - it invites hallucination on the part of the agent.  In contrast, if an agent can read a competitors pricing, for example, or their compliance page, and it has all this information, we lose evaluation without the buyer ever signing up or contacting us directly.

PR #2602, opened 2026-08-06
Clarify automatic shard rebalancing

abdonpijpelink:
@ihorponom reported that, in the docs, three things could be more clear about automatic shard rebalancing:

- Rebalancing works constantly in the background; not just when scaling a cluster.
- Rebalancing is not per collection, but globally.
- When manually moving shards, the rebalancer may move your shard back. Disable the rebalancer if you want to manually assign shards.

This PR addresses that by making the following changes:

- `cloud/configure-cluster.md` — Rebalancing is continuous, not scaling-only; balances per-node across all collections; added warning that manual moves can be reverted.
- `scaling/distributed_deployment.md` — Same "continuous, not just on node add/remove" clarification; added a note that manual shard moves can be reverted.
- `cloud/cluster-monitoring.md` — Two FAQ entries updated to explain rebalancing is continuous and per-node-across-collections.
- `hybrid-cloud/configure-scale-upgrade.md` — Reworded to say rebalancing is continuous, not just tied to scaling.
- `private-cloud/qdrant-cluster-management.md` — Same continuous-rebalancing clarification plus the manual-move-reverted warning.


ihorponom (changes requested):
Looks good overall; I would just change this part 
`deactivate automatic shard rebalancing by not selecting a value for the Shard Rebalance Strategy option.`
Because to deactivate, we need to click the 'cross' and then save.  (we can rephrase but it's the same sense)

#upd
I see in development we have changed it to Disable option in dropdown
image

generall:
> Because to deactivate, we need to click the 'cross' and then save.

This doesn't work for me, it just goes back to default.

ihorponom:
> > Because to deactivate, we need to click the 'cross' and then save.
> 
> This doesn't work for me, it just goes back to default.

Yes, you are right, just checked on Prod. 

But 'Disabled' option seems to work on Development

abdonpijpelink:
I've pushed a change that changes the wording into _"deactivate automatic shard rebalancing by selecting Disabled for the Shard Rebalance Strategy option"_, to match the new option in Dev. We can wait with merging this PR until that's been released to Prod.

ihorponom (approved):

bashofmann (approved):
lgtm

PR #2613, opened 2026-08-10
Add blog post on cleaning up a vector collection

Dylancouzon:
[Preview](https://deploy-preview-2613--condescending-goldwasser-91acf0.netlify.app/blog/clean-vector-database-collection/)

Follow-up to the webinar I ran with Future AGI. The blog owns the problem statement and the video is embedded at the end.

The title: "vector database" is in there for the SEO grab, which means check_content.py flags it as a blocker. Deliberate, but happy to change it

My goal is to follow this up with a full-fledged article on collection hygiene, since we have almost nothing on it. Does that sound like a good idea?


kanungle (commented):
"Collection" is a Qdrant-specific term so "Vector Database Collection" wouldn't have much SEO effect. Might as well just say "Qdrant Collection".

Should be okay to make this `featured: true`

PR preview link is broken

Idea for article sounds good to me

Thanks

Dylancouzon:
My b, all fixed

kanungle (approved):
Nice work

PR #2621, opened 2026-08-12
Publish merged redirect table for consumers outside the Netlify CDN

szabosteve:
## Overview

Related to https://github.com/qdrant/skills/pull/139

qdrant.tech's redirects live in three separate tables: _redirects, netlify.toml, and Hugo aliases: front matter. None of which is fetchable over HTTP, so the `/md/` mirror on `search.qdrant.tech` has no way to learn that a path moved and 404s old paths the site redirects correctly. This adds a build step that merges all three into public/redirects.txt. 
The `/md/` 404s are not fixed until the counterpart PR in `skills` lands the changes that consume this file.


abdonpijpelink (approved):
Thanks!

PR #2622, opened 2026-08-12
fix: generated md syntax check to avoid broken code

AstraBert:
This is an attempt to enhance snippets CI by adding a syntax checking step for those language that allow checking without compiling (go, python, bash, rust).

All existing code snippets pass syntax checking and still compile/build.


kanungle (commented):
Deferring decision to @abdonpijpelink and @szabosteve on this

abdonpijpelink (approved):
Nice one! Thanks!

PR #2625, opened 2026-08-13
Move Supernova post from blog to research section

nleroy917:
Relocates the 10B-vector Supernova post and its figures out of the blog section and into the new research section, and rewrites the asset paths from /blog/supernova-10b/ to /research/supernova-10b/.

Post was never published, so no aliases are needed for the URL change.


nleroy917:
Closing in favor of putting it in the typical blog section

PR #2628, opened 2026-08-14
Add Bayer to customers page

daniel-azoulai:
Add a Bayer entry to the customers page client list, placed in alphabetical order between Aracor and Bazaarvoice, linking to the merged case study at /blog/case-study-bayer.

Also fix two typos in the adjacent ConvoSearch entry: company_size "1-50 by" -> "1-50" and the misspelled key "locationse" -> "location", which had been keeping ConvoSearch out of the Location filter.


daniel-azoulai:
@nastyapash I added the Bayer case study to the new customers page. Can you just check that I did it correctly? Once I figure it out, I can do it independently. 

Thanks

PR #2629, opened 2026-08-14
Add Edge API page to Edge docs

szabosteve:
### [Edge preview](https://deploy-preview-2629--condescending-goldwasser-91acf0.netlify.app/documentation/edge/)

This PR adds a multi-page Edge API reference documenting every EdgeShard method in Python and Rust, with parameters, return values, and language differences.


qdrant-landing/content/documentation/edge/edge-api/reading-data.md
```diff
@@ -0,0 +1,212 @@
...
+In Rust, each request type has a fluent builder, so you only set the parameters you need:
+
+```rust
+let request = QueryRequest::builder()
+    .limit(10)
+    .with_payload(true)
+    .build();
+```
+
+Builders are available for `QueryRequest`, `SearchRequest`, `ScrollRequest`, `RetrieveRequest`, `CountRequest`, `FacetRequest`, `GroupRequest`, `SearchMatrixRequest`, and `Prefetch`. The Python bindings construct requests through their class constructors instead, where every optional parameter defaults to `None`.
```
xzfc: There are no `builder()` method for many of these.

E.g. there is no `QueryRequest::builder()`, but `QueryRequestBuilder::new()` exists.

qdrant-landing/content/documentation/edge/edge-api/reading-data.md
```diff
@@ -0,0 +1,212 @@
...
+
+Runs a single scoring query. `search` is a convenience path; with no prefetches or fusion to express, the request is simpler to build than the `query` equivalent.
+
+```python
+def search(self, search: SearchRequest) -> List[ScoredPoint]
+```
+
+```rust
+fn search(&self, request: SearchRequest) -> OperationResult<Vec<ScoredPoint>>
+```
```
xzfc: We are deprecating `search()` in favor of `query()`.

qdrant-landing/content/documentation/edge/edge-api/shard-lifecycle.md
```diff
@@ -0,0 +1,143 @@
...
+
+In Rust, `info` comes from the `EdgeShardRead` trait rather than from `EdgeShard` itself, so bring it into scope with `use qdrant_edge::EdgeShardRead;` before calling it.
+
+`ShardInfo` carries the following fields:
+
+| Field | Type | Description |
+|---|---|---|
+| `segments_count` | `int` | Number of segments in the shard. |
+| `points_count` | `int` | Number of points stored. |
+| `indexed_vectors_count` | `int` | Number of vectors that have been added to a vector index. |
```
xzfc: Add the word "approximate".

(we had many users' reports that `points_count` value is incorrect, so we want to say the word "approximate" very explicitly)

qdrant-landing/content/documentation/edge/edge-api/reading-data.md
```diff
@@ -0,0 +1,212 @@
...
+```rust
+fn count(&self, request: CountRequest) -> OperationResult<usize>
+```
+
+**Returns** the number of matching points.
+
+| Parameter | Type | Description |
+|---|---|---|
+| `filter` | `Filter` | Conditions the counted points must satisfy. Omit to count every point. |
+| `exact` | `bool` | Count exactly rather than estimating. Defaults to `True` in Python. |
```
xzfc:
```suggestion
| `exact` | `bool` | Count exactly rather than estimating. |
```

qdrant-landing/content/documentation/edge/edge-api/shard-lifecycle.md
```diff
@@ -0,0 +1,143 @@
...
+def create(path: str, config: EdgeConfig) -> EdgeShard
+```
+
+```rust
+pub fn new(path: &Path, config: EdgeConfig) -> OperationResult<EdgeShard>
+```
+
+| Parameter | Type | Description |
+|---|---|---|
+| `path` | `str` (Python) / `&Path` (Rust) | Path to the shard directory. Must not already contain segment data. |
```
xzfc: (opinion)

I think the "Type" column is redundant (here and in other such tables): we already have types in the snippet just above. Especially when it is verbose, e.g., "`str` (Python) / `&Path` (Rust)".

xzfc (changes requested):

szabosteve:
The Edge quickstart snippet sets `max_search_threads` and `search_pool_core` on `EdgeConfig`. Both parameters were added in qdrant-edge-py 0.8.0, but the snippet harness pinned 0.7.2, so mypy rejected them as unexpected keyword arguments and typecheck-snippets failed.

I bumped the pin and regenerated the lockfile via [97d36f2](https://github.com/qdrant/landing_page/pull/2629/commits/97d36f22b70c300e5125ea620548dacca1f9d832). qdrant-edge-py is the only package that moved; the other Edge snippets typecheck unchanged against 0.8.0.

xzfc (approved):

qdrant-landing/content/documentation/edge/edge-api/configuration.md
```diff
@@ -0,0 +1,206 @@
...
+
+```rust
+pub fn builder(size: usize, distance: Distance) -> EdgeVectorParamsBuilder
+```
+
+| Parameter | Description |
+|---|---|
+| `size` | Vector dimension. Required. |
+| `distance` | Distance metric. Required. |
+| `on_disk` | Store vectors on disk with mmap rather than in RAM. |
```
abdonpijpelink:
```suggestion
| `on_disk` | Whether to cache vectors in RAM for faster access, or serve them from disk. |
```
Vectors are always stored on disk. We should avoid "store" describing these "on disk" parameters.

qdrant-landing/content/documentation/edge/edge-api/configuration.md
```diff
@@ -0,0 +1,206 @@
...
+    .vector("text", EdgeVectorParams::builder(384, Distance::Cosine).build())
+    .on_disk_payload(true)
+    .build();
+```
+
+| Parameter | Description |
+|---|---|
+| `vectors` | Dense vector configuration. In Python, a single `EdgeVectorParams` configures the default unnamed vector. Optional if `sparse_vectors` is given. |
+| `sparse_vectors` | Sparse vector configuration. |
+| `on_disk_payload` | Store payload on disk with mmap rather than in RAM. Defaults to on-disk. |
```
abdonpijpelink:
```suggestion
| `on_disk_payload` | Whether to cache payloads in RAM for faster access, or serve them from disk. |
```

qdrant-landing/content/documentation/edge/edge-api/configuration.md
```diff
@@ -0,0 +1,206 @@
...
+```
+
+```rust
+pub fn builder() -> EdgeSparseVectorParamsBuilder
+```
+
+| Parameter | Description |
+|---|---|
+| `full_scan_threshold` | Threshold below which a full scan is used instead of the sparse index. |
+| `on_disk` | Store the sparse index on disk rather than in RAM. |
```
abdonpijpelink:
```suggestion
| `on_disk` | Whether to cache sparse vector indexes in RAM for faster access, or serve them from disk.  |
```

qdrant-landing/content/documentation/edge/edge-api/configuration.md
```diff
@@ -0,0 +1,206 @@
...
+```
+
+| Parameter | Description |
+|---|---|
+| `deleted_threshold` | Minimum fraction of deleted vectors in a segment required to run vacuum. Default: `0.2`. |
+| `vacuum_min_vector_number` | Minimum number of vectors in a segment required to run vacuum. Default: `1000`. |
+| `default_segment_number` | Target number of segments. `0` chooses automatically from the CPU count. |
+| `max_segment_size` | Maximum segment size in KB. Derived from the CPU count when unset. |
+| `indexing_threshold` | Size in KB above which a segment gets an HNSW index. |
+| `prevent_unoptimized` | Store points written to unoptimized segments larger than the indexing threshold as deferred points: they are persisted but excluded from reads and searches until the segments are optimized. |
```
abdonpijpelink:
```suggestion
| `prevent_unoptimized` | [Prevents slow reads from large unoptimized segments](/documentation/ops-optimization/optimizer/#prevent-reads-from-large-unindexed-segments) by deferring the visibility of points until they've been indexed. |
```
Just a suggestion. I think a link here is helpful.

qdrant-landing/content/documentation/edge/edge-api/updating-data.md
```diff
@@ -0,0 +1,81 @@
...
+| `set_payload` | `point_ids`, `payload`, `key` | Merge payload fields into points. |
+| `set_payload_by_filter` | `filter`, `payload`, `key` | Merge payload fields into matching points. |
+| `overwrite_payload` | `point_ids`, `payload`, `key` | Replace the entire payload on points. |
+| `overwrite_payload_by_filter` | `filter`, `payload`, `key` | Replace the entire payload on matching points. |
+| `delete_payload` | `point_ids`, `keys` | Delete named payload fields from points. |
+| `delete_payload_by_filter` | `filter`, `keys` | Delete named payload fields from matching points. |
+| `clear_payload` | `point_ids` | Delete all payload from points. |
+| `clear_payload_by_filter` | `filter` | Delete all payload from matching points. |
+| `create_field_index` | `field_name`, `schema` | Index a payload field. |
+| `delete_field_index` | `field_name` | Remove a payload field index. |
```
abdonpijpelink, line 58:
```suggestion
### Payload Indexes

The `update` operation also enables you to create and delete payload indexes.

| Operation | Parameters | Description |
|---|---|---|
| `create_field_index` | `field_name`, `schema` | Index a payload field. |
| `delete_field_index` | `field_name` | Remove a payload field index. |

### Modify the Vector Schema

Add or remove named vectors to an existing Edge Shard’s schema. This is useful when migrating to a new embedding model or adding hybrid search to an Edge Shard that already contains data.

| Operation | Parameters | Description |
|---|---|---|
```

qdrant-landing/content/documentation/edge/edge-api/updating-data.md
```diff
@@ -0,0 +1,81 @@
...
+
+**Returns** nothing in Python. In Rust, returns `Ok(())` on success.
+
+Creating a named vector that already exists with different parameters fails.
+
+## Update Operations
+
+In Python, `UpdateOperation` is a class with static constructors, one per operation. Each returns an `UpdateOperation` to pass to `update`.
+
+| Operation | Parameters | Description |
```
abdonpijpelink, line 34: I've left a suggestion to break this table into three:

- one for operations on data
- one for payload index changes
- one for schema-level changes

The latter two should perhaps be in the "Configuration" section, but for reference purposes it's also fine to keep them here.

qdrant-landing/content/documentation/edge/edge-api/updating-data.md
```diff
@@ -0,0 +1,81 @@
...
+| `delete_payload` | `point_ids`, `keys` | Delete named payload fields from points. |
+| `delete_payload_by_filter` | `filter`, `keys` | Delete named payload fields from matching points. |
+| `clear_payload` | `point_ids` | Delete all payload from points. |
+| `clear_payload_by_filter` | `filter` | Delete all payload from matching points. |
+| `create_field_index` | `field_name`, `schema` | Index a payload field. |
+| `delete_field_index` | `field_name` | Remove a payload field index. |
+| `create_dense_vector` | `vector_name`, `size`, `distance`, `multivector_config`, `datatype` | Add a dense named vector to the schema. |
+| `create_sparse_vector` | `vector_name`, `modifier`, `datatype` | Add a sparse named vector to the schema. |
+| `delete_vector_name` | `vector_name` | Remove a named vector from the schema. |
+
```
abdonpijpelink, line 69:
```suggestion

### Parameters

```
If breaking up the table, we need this header I think

abdonpijpelink (commented):
This is looking great. Left a few suggestions.

abdonpijpelink (approved):

PR #2631, opened 2026-08-14
Add Google Tag Manager with OneTrust-driven Consent Mode v2

kanungle:
## TL;DR (human-generated)
Google Tag Manager appears to not have been configured correctly when the qdrant website 2024 theme was launched. There was no GTM container ID present. This means web tracking for Google Analytics is off (though we do track through Segment as well, starting a few months ago).

This PR addresses the GTM issue, and also includes in OneTrust Consent.

We do not have anyone fully owning this at Qdrant so I've tagged a few suggested reviewers. If unsure, let me know.

Note: OneTrust Consent should be validated separately by whoever owns that.

Thanks

# Claude Code Details:

## Why

GTM has never been live on qdrant.tech. Verified against production HTML (`/`, `/documentation/`, `/blog/`, `/articles/`): zero hits for `gtm.js`, `ns.html`, or any `GTM-` string.

The repo *looks* like it has GTM already — `themes/qdrant/layouts/partials/head.html` carries a full snippet gated on `.Site.Params.googleTagManager`. That file is dead code: both `config.toml:4` and `config-theme.toml:1` set `theme = "qdrant-2024"`, so the snippet never renders no matter what params are set. This PR adds GTM to the **active** theme instead.

## What

Container `GTM-M35C8G6`, wired into `themes/qdrant-2024/layouts/partials/js.html` immediately after the OneTrust stub so the CMP always loads first.

**Consent Mode v2, denied by default.** Every signal (`ad_storage`, `ad_user_data`, `ad_personalization`, `analytics_storage`, `functionality_storage`, `personalization_storage`) is set to `denied` in a synchronous script *above* the `gtm.js` loader, with `wait_for_update: 500`. No tag in the container can fire before a consent choice exists. The site runs a OneTrust CMP and serves EU traffic (EU HubSpot, EU Segment), so a bare GTM snippet would have been a compliance regression.

**Consent is granted from OneTrust, reusing the existing plumbing.** `gtm-helpers.js` maps `window.OnetrustActiveGroups` onto Consent Mode signals. It is called from the `onetrust_loaded` and `OnConsentChanged` handlers already present in `index.js` — no second listener, no duplicate CMP wiring.

| OneTrust group | Consent Mode signal |
|---|---|
| `C0002` performance | `analytics_storage` |
| `C0003` functional | `functionality_storage`, `personalization_storage` |
| `C0004` targeting | `ad_storage`, `ad_user_data`, `ad_personalization` |

**Production only.** The ID is set in `[context.production.environment]` alongside the existing `segmentWriteKey` / `onetrustScriptId`, so deploy previews and local `hugo serve` render no GTM at all and won't pollute the container.

## ⚠️ One assumption to confirm before merge

The group IDs above are **OneTrust defaults**. Nothing in this repo names a specific group — `onetrust-helpers.js` only passes `OnetrustActiveGroups` through as an opaque string — so I could not verify them from code.

**If the OneTrust console for domain `01960152-5e40-782d-af01-f7f5768a214e` uses custom group IDs, this mapping will look correct in review and silently never grant consent in production.** The `OT_GROUP` constants at the top of `gtm-helpers.js` are the only thing to change. Could whoever owns the OneTrust console confirm?

## Verification

Built locally with the param set — correct render and ordering (HubSpot → OneTrust → consent defaults → `gtm.js` → Segment):

```
onetrust otSDKStub       778192
consent default          778xxx   (all denied)
gtm.js loader            778985   (GTM-M35C8G6)
segment analytics.load   779xxx
```

Built again *without* the param: `0` occurrences of `gtm.js`, the container ID, and the consent block — confirming previews stay clean.

Consent mapping has a dependency-free self-check (no test runner in this theme):

```bash
node qdrant-landing/themes/qdrant-2024/assets/js/gtm-helpers.check.mjs
# gtm-helpers: all consent mapping checks passed
```

It covers no-consent, strictly-necessary-only, each group in isolation, full consent, `undefined`, and that `C00020` does not satisfy `C0002`.

**Manual check on the preview deploy is not possible** (GTM is production-gated by design). After merge, confirm in GTM Preview mode that consent transitions from all-denied to granted when accepting cookies.

## Deliberately skipped

- **`<noscript>` iframe fallback** — needs a `baseof.html` edit, only serves JS-disabled users, and cannot respect consent. Add if someone wants it.
- **Migrating any Segment tracking into GTM** — this PR is purely additive. Segment, HubSpot, and the `ga_measurement_id` cookie are untouched.
- **Deleting the dead `themes/qdrant/` theme** — out of scope, but it's ~a whole theme of unused code and worth its own PR.

## Unrelated bugs found while tracing this (not fixed here)

1. `segment-helpers.js:32,37` — `handleClickInteraction` reads `event.target` instead of `event.currentTarget`. Any `data-metric-loc` element containing child markup will report the child and send `location: ''`. Today's 19 tagged elements are all plain-text anchors so it works, but it breaks silently the moment someone wraps an icon or `<span>` in one.
2. `partials/home/hero.html:11-14` — duplicate `data-metric-loc="hero"` on both anchors.

### DIFF
--- a/qdrant-landing/themes/qdrant-2024/assets/js/gtm-helpers.js
+++ b/qdrant-landing/themes/qdrant-2024/assets/js/gtm-helpers.js
@@ -1,0 +1,43 @@
+/**
+ * Bridges OneTrust consent choices onto Google Consent Mode v2 signals for GTM.
+ *
+ * The GTM container is loaded in partials/js.html with every consent signal
+ * defaulted to 'denied'. Nothing in the container fires until updateGtmConsent()
+ * pushes an update, which happens once OneTrust reports its stored choice and
+ * again on every preference change (see index.js).
+ *
+ * Group IDs below are the OneTrust defaults. If the OneTrust console for
+ * qdrant.tech uses custom group IDs, these constants are the only thing to change.
+ */
+const OT_GROUP = {
+  PERFORMANCE: 'C0002',
+  FUNCTIONAL: 'C0003',
+  TARGETING: 'C0004',
+};
+
+/**
+ * OneTrust exposes consent as a delimited string, e.g. ',C0001,C0002,'.
+ * Split rather than substring-match so 'C0002' never matches 'C00020'.
+ */
+export function mapOneTrustConsent(activeGroups) {
+  const groups = (activeGroups ?? '').split(',');
+  const performance = groups.includes(OT_GROUP.PERFORMANCE);
+  const functional = groups.includes(OT_GROUP.FUNCTIONAL);
+  const targeting = groups.includes(OT_GROUP.TARGETING);
+
+  return {
+    ad_storage: targeting ? 'granted' : 'denied',
+    ad_user_data: targeting ? 'granted' : 'denied',
+    ad_personalization: targeting ? 'granted' : 'denied',
+    analytics_storage: performance ? 'granted' : 'denied',
+    functionality_storage: functional ? 'granted' : 'denied',
+    personalization_storage: functional ? 'granted' : 'denied',
+  };
+}
+
+export function updateGtmConsent() {
+  // gtag is only defined when the container is configured (production w/ a container ID)
+  if (typeof window.gtag !== 'function') return;
+
+  window.gtag('consent', 'update', mapOneTrustConsent(window.OnetrustActiveGroups));
+}
--- a/netlify.toml
+++ b/netlify.toml
@@ -15,6 +15,7 @@ HUGO_ENABLEGITINFO = "true"
 DEPLOY_PRIME_URL = "https://qdrant.tech"
 HUGO_PARAMS_segmentWriteKey = "eQYi1nZE2zQSmnHuxjMRlDd2uLl65oHe"
 HUGO_PARAMS_onetrustScriptId = "01960152-5e40-782d-af01-f7f5768a214e"
+HUGO_PARAMS_googleTagManager = "GTM-M35C8G6"
 
 [context.deploy-preview.environment]
 HUGO_ENV = "staging"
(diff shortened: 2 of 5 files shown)


kanungle:
@ohitsmekatie can you validate this would not double-count page views? Is Segment upstream, downstream, or parallel to this tag?

qdrant-landing/themes/qdrant-2024/assets/js/gtm-helpers.js
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * The GTM container is loaded in partials/js.html with every consent signal
+ * defaulted to 'denied'. Nothing in the container fires until updateGtmConsent()
+ * pushes an update, which happens once OneTrust reports its stored choice and
+ * again on every preference change (see index.js).
+ *
+ * Group IDs below are the OneTrust defaults. If the OneTrust console for
+ * qdrant.tech uses custom group IDs, these constants are the only thing to change.
+ */
+const OT_GROUP = {
```
ohitsmekatie, line 12: I'm not sure who the best version to verify the consent tracking versions would be but maybe @Alyonkka would know. These are the versions that I see recently but I have no context on what they mean or if they all need to be added. 


Image

ohitsmekatie (commented):
I think this will double count pageviews into Google Analytics. Segment would be parallel to this tag. The tracking is setup to track every page view and then from Segment we disperse that data to a few places (GTM + Snowflake). 

1. I can stop sending the page views from Segment --> GA4 if folks would rather use GTM directly on the site. Just let me know when this PR is tested + merged and I will disable that integration. It might be worth mentioning this change in #ops so @hannahgeiser-qdrant  + @maddie-qdrant  can set up any new page view triggers that they need using this. 
2. Someone who is more familiar with the consent process should review the versions included and/or test the changes to make sure it's functioning the way it needs to be 


Just leaving a comment because I didn't review the logic/code.

trean:
Just fyi, there was a moment when the marketing team requested that the GTM script be removed from the website.
https://github.com/qdrant/landing_page/pull/1412

Just in case, to be sure it won't connect twice now.

kanungle:
> Just fyi, there was a moment when the marketing team requested that the GTM script be removed from the website.
> https://github.com/qdrant/landing_page/pull/1412
> 
> Just in case, to be sure it won't connect twice now.

Thanks, this is useful context. I don't believe it should have been disconnected as it doesn't hurt to have both fire at the same time in parallel but let's leave this PR open for now while I double-check

kanungle:
ongoing discussions in Ops and Marketing about this. Closing the PR for now

PR #2633, opened 2026-08-14
Align Go, C#, and Java snippet booleans with their descriptions

Dylancouzon:
Two snippets set a boolean that contradicts their own `_description.md`, so the language tabs on the same page disagree with each other.

**`query-points/with-ignored-quantization`** — the description says "By setting the `ignore` parameter to true", and python, typescript, rust, java, csharp, and http all pass true. Go passed `Ignore: qdrant.PtrOf(false)`, which does the opposite of what the snippet is demonstrating. Renders on `manage-data/quantization.md:341`.

**`create-payload-index/lowercase-full-text`** — the description says the index is for case-sensitive matching and that lowercasing is disabled here. Python, typescript, rust, and http pass false. Go, C#, and Java passed true. The sibling dirs `simple-full-text` and `phrase-full-text` set true in all seven languages, so this one looks copied from a sibling with only four of the seven files flipped. Renders on `manage-data/indexing.md:222`.

Checked the field types against `go-client` v1.19.0, which `automation/snippets/templates/go/go.mod` pins: `QuantizationSearchParams.Ignore` and the text-index `Lowercase` are both `*bool`, so `qdrant.PtrOf(false)` is the right shape and only the literal changes.

Regenerated the `generated/*.md` files with `automation/snippets/generate-md.py` per the README, rather than hand-editing them. The generator touched only the four corresponding files.

Not included: `query-points-explore/relevance-feedback-naive` sets `c = 0.16` in Go, Rust, Java, and C# but `c = 0.03` in Python, TypeScript, and HTTP. There is no description to arbitrate which is intended, so someone who knows the intent should pick.

### DIFF
--- a/qdrant-landing/content/documentation/headless/snippets/create-payload-index/lowercase-full-text/csharp.cs
+++ b/qdrant-landing/content/documentation/headless/snippets/create-payload-index/lowercase-full-text/csharp.cs
@@ -16,7 +16,7 @@ public class Snippet
 		        TextIndexParams = new TextIndexParams
 		        {
 		            Tokenizer = TokenizerType.Word,
-		            Lowercase = true,
+		            Lowercase = false,
 		        }
 		    }
 		);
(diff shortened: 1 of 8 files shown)


Anush008 (approved):

Anush008:
LGTM! Thanks.

PR #2634, opened 2026-08-14
Hybrid fusion Articles 

Dylancouzon:
Promotion blog: https://deploy-preview-2634--condescending-goldwasser-91acf0.netlify.app/blog/tuning-retrieval-which-knob-first/
## What This Is

Five articles on tuning Qdrant retrieval: what to check, what to measure, and what to change next. Each stands on its own, but they work best as a series and cross-reference heavily.

They are for experienced builders(Not in search) who have already set up a basic retrieval pipeline and now need to improve it.

- **[What to Check Before Tuning a Qdrant Collection](https://deploy-preview-2634--condescending-goldwasser-91acf0.netlify.app/articles/before-tuning-a-qdrant-collection/):** Eight settings that fail silently, and a symptom table that routes each failure to a fix.
- **[Candidate Depth: How Much Retrieval Is Enough?](https://deploy-preview-2634--condescending-goldwasser-91acf0.netlify.app/articles/candidate-depth/)** What raising prefetch `limit` buys, measured against the best score your candidates could reach.
- **[How to Tune Hybrid Search in Qdrant:](https://deploy-preview-2634--condescending-goldwasser-91acf0.netlify.app/articles/how-to-tune-hybrid-search/)** RRF against DBSF, choosing `k`, and why weights are pairs.
- **[When Is a Reranker Worth It?](https://deploy-preview-2634--condescending-goldwasser-91acf0.netlify.app/articles/when-a-reranker-is-worth-it/)** A cross-encoder beat a tuned fusion on one dataset of five. How to test yours.
- **[When Your Collection Outgrows RAM:](https://deploy-preview-2634--condescending-goldwasser-91acf0.netlify.app/articles/when-your-collection-outgrows-ram/)** Memory placement and what rescoring costs once the original vectors leave RAM.

Every number is measured and source-checked. I optimized for what the reader needs to take away, not exhaustive coverage. That means less code than usual: the reader can write it, or hand it to an agent. Our job here is to give them a useful order of operations.

## Results

The gains are modest because the collections were clean, single-shard, unquantized, built in one batch, and ingested at once, while the hybrid baseline was already well tuned.

Neither changes the conclusions. More realistic collections would likely improve the numbers without changing what they teach. The dataset selected were diverse, but lack the messiness of production.  The harness is still up if a rerun is worth it. I’ll leave that call to @kanugle.


Dylancouzon:
@kanungle I used you prompt template in /neil-review to create the hero but they do not look on brand. Feel free to replace

kanungle:
@mrscoopers @Dylancouzon let's together review the first article only for an initial pass. If writing and direction meet our aligned goals, then we can review the others

kanungle:
The first article looks really good to me. The only issue I found is the prose in the intro section feels very AI generated. The rest of the article was well structured and easy to follow. Moving on to the second article...

mrscoopers:
Skimmed through "What to Check Before Tuning a Qdrant Collection"
Btw, I don't get the difference between the "Mastering Search" and "Search Quality" categories in articles now, but judging by the description, the series belongs more in "Mastering Search"(?)

1. Title might be confusing; I'd expect some major guide about everything, but throughout I get a feeling it's more about quality of search ranking(?) (but not latency/resources) (f.e. "You cannot separate a gain from noise" -> this article)
2. "A few settings are correctness checks, not tuning knobs. Get one wrong, and every benchmark or parameter sweep after it measures a broken setup." I'd drop or give example/ explicit reference to these settings ("Silent Settings Can Break Quality"?)
3. "Dense-only search runs the first stage alone." in caption - what does it mean? I assume sparse one, too?
4. Each ran unquantized on one shard in a laptop Docker container, using all-MiniLM-L6-v2 and Qdrant’s core BM25,.. "except the depth article’s quantization comparison.
[The memory article](https://deploy-preview-2634--condescending-goldwasser-91acf0.netlify.app/articles/when-your-collection-outgrows-ram/) is the exception: its 4.6 million vectors are large enough to cross a real RAM boundary" - I'd drop everything after "except"
5. I'd drop "The query returns results, scores look plausible, and the retrieval setup is still wrong."
6. "Other retrieval settings depend on a latency, memory, or rebuild budget you supply. These checks do not." - honestly didn't get why
7. Idk if I'd mix indexed vectors count and threshold in one group; count is not smth that is definable by user(?)
8. "With one segment per two CPUs by default, clamped to between two and eight, multiply the relevant per-segment figure by your segment count to see when the whole collection is covered." I think this is very dense; I'd extend or give smth more like a formula? I have a feeling assistants have a very bad manner of putting everything in text, even something that is better to structure visually
9. "full_scan_threshold tells Qdrant when to use an exact full scan instead of HNSW." reads like for sparse vectors it's applicable, too
10. "Enable it on the sparse vector." for IDF - might be risky if smbd without thinking uses splade + IDF
11. "Measure the indexed field’s post-stemming average token count and supply that value to BM25." I think this is also a little dense. Maybe we could leave a link to BM25 docs?
12. "It is healthy when fusion is the root query" reads like fusion not on root level is unhealthy in any case
13. "Strict filters cost recall even once the indexes are in place, and none of the measurements in this series were taken on a filtered collection. [What ACORN fixes, and what fixes ACORN](https://deploy-preview-2634--condescending-goldwasser-91acf0.netlify.app/articles/filtered-vector-search-acorn/) measures that on one million points." Maybe drop this, because of "none of the measurements in this series were taken on a filtered collection"?
14. "and 0.6 to 1.5 ms of query time on one shard" this one seems experiments-based? Maybe we should remove?
15. "quantization" seems like one thing that doesn't fit the table; everything else seems to be about search ranking quality, not resource optimization. 
16. "A Matryoshka, SPLADE, etc" part seems like it could be dropped; too many new concepts & not a part of collection knobs anyway
17. I'd structure and cut a little bit of the metrics section. It's a good reminder about recall, but it's slightly hard to read; I'd group insights under the corresponding metric 
18. "The number of labeled queries determines the width of that interval. Across our datasets, the median 95% interval half-width for paired nDCG@10 gains was:" 
19. I feel like "Check the Winner on Fresh Queries" doesn't quite read like a conclusion; there are a lot of numbers and several recommendations, and it feels a little overwhelming. Same feeling about the ending of "Make Sure Your Labels Can Detect a Gain"

So it to sum up: I'd free it up a bit and add some structure for readability; in general feels like a very nice read!

mrscoopers (dismissed):
comment above

kanungle (changes requested):
Cannot comment inline since there are so many files. Here's my feedback so far:

**Article 1-**
- Rewrite intro, too much Claude
- This paragraph too data-heavy:
> In our measurements, [fusion settings](https://deploy-preview-2634--condescending-goldwasser-91acf0.netlify.app/articles/how-to-tune-hybrid-search/) moved nDCG@10 by 0.012 to 0.038. These are gains from tuning an already-working collection, not from rebuilding the retrieval pipeline.
With 50 labeled queries, only the 0.038 gain had a 95% interval that excluded zero in most draws: 93%, versus 7% to 38% for gains under 0.02. Detecting a 0.015 gain required 200 to 1,000 queries, depending on the dataset.

**Article 2-**
- Many linebreaks between paragraphs missing in the rendered preview
- Question: how generalizable is this advice with increasing scale?
- Make it `limit=500` in 3rd section
- Show a conceptual diagram like a "saturated graph" for raised hnsw_ef (optional)
- Recall vs Latency chart might be useful (optional)

I will provide hero images and review next articles after these changes addressed.

mrscoopers:
> Candidate Depth: How Much Retrieval Is Enough?
1. "In hybrid search, each prefetch has its own limit" -> multistage? Or do we want to focus only on a 1-level-deep hybrid? Maybe we should reflect it in the title
2. "Unless noted otherwise, these measurements use " - maybe we need to be more explicit that articles have to be read & applied sequentially 
3. In the short version (and full one, too), section 2nd point looks unrelated to the beginning of the article, which says that here we're tuning: "Candidate depth is the number of candidates a retrieval stage passes to a later ranking stage.", hnsw_ef won't affect that, afaik(?) Or I understand it differently; Maybe it should be in some other article? Say in ANN recall one
4. "More Candidates Can Raise the Best Possible Score" -> "Use your [labeled query set](https://deploy-preview-2634--condescending-goldwasser-91acf0.netlify.app/articles/before-tuning-a-qdrant-collection/#make-sure-your-labels-can-detect-a-gain) to score the candidate set as if it were ordered perfectly. ... For a single-prefetch pipeline, score the candidates passed to the downstream stage." - true but hard to read, maybe better on a small example (1 query + NSCG@10? As, f.e., best possible hit@1 won't change with limit increase, or I don't get what's "best possible score" & "current score")
5 I feel like "Depth Is Per Shard" first paragraph should be before this section or called smth like "But mind latency"
6 "That is what a saturated graph looks like." Calls for a graph visual?
7 "Reducing candidate depth does not make the collection smaller. "Does it refer to hnsw_ef experiments? Otherwise I can't imagine smbd thinking it reduces collection
8 In "When RAM Limits Candidate Depth," I'd expect something proving why quantization > reducing the limit (resources, latency, quality)
9 "Settings for Specific Cases" feels like they'd be better digested when distributed in relevant sections

Overall, this article for me didn't read like one story or a step-by-step instruction, ig mainly because of the intro/title, which set for me different expectations(?) I am missing context, especially if it's supposed to be read independently of the first one. If it's only for agents, not for ppl, lmk; I am still reviewing more from a human reader perspective.

Dylancouzon:
Article 1

Applied all of @kanungle's feedback.

@mrscoopers 

**Title and category**

I checked the other search-quality articles. `predicting-weak-retrieval` is the closest match, practical and measured. `relevance-feedback`, `search-feedback-loop`, and the Quotient tutorial are different content types, so the category was already mixed. `mastering-search` is where the feature how-tos live, like hybrid search and filtering, so search-quality still looks like the better home for this one.

On the title: the article also covers latency and memory, including p95, `hnsw_ef`, and prefetch limits, so I think it holds. Happy to revisit if you feel strongly.

**The rest**

2, 3, 4, 5: applied.

6: removed. The rewritten intro already makes that point.

7, 8, 9: collapsed. The count comparison and `indexing_threshold` are now one check, and `full_scan_threshold` says outright that dense and sparse are separate settings in different units, 10,000 KB and 5,000 vectors, with the sparse one applying only to filtered search.

10, 11, 12: applied. IDF now names which models expect it and why SPLADE does not, and fusion placement describes what each placement does instead of calling one of them healthy.

13: dropped the claim, kept the link, and rewrote the paragraph.

14: kept the number, it earns its place in a table about cost. The setup line now reads "The measurements in this article" rather than "The relevance measurements", so a latency figure sits inside the stated basis.

15: kept quantization in the table, but gave it its own row labeled for memory-limited collections. The symptom table routes RAM pressure here, so I would rather not send it elsewhere.

16: mostly agree. Cut SPLADE and ColBERT, kept Matryoshka `mrl`, since it is the other lever when memory is the constraint.

17: agreed, this needed it. Each insight now sits under its metric.

18: read this as a density problem and rewrote the lead-in. Tell me if you meant something else.

19: applied. The closing section is four paragraphs and ends on one next step.

I also did a full narration and clarity pass over the article after applying these.

Dylancouzon:
## Article 2: Candidate Depth

### @kanungle

Fixed the line breaks. All 18 were `<br>` tags.

Added the scale answer in two places. The best possible score change rises with corpus size across the five datasets, while the current score change stays flat. And at `hnsw_ef=128` the full 4,635,922-document DBPedia collection returned 0.957 of the exact top 10, so about 4% of the true nearest neighbors never came back. Depth headroom grows with scale, graph-search recall shrinks with it.

Holding on `limit=500` until I know what it should improve. That table was measured at `limit=100`, so it needs a rerun, and a higher limit flattens the recall column rather than sharpening it. Depth 500 also costs 37% to 43% more median latency than depth 10 for no quality gain.

Added the saturated graph, the recall vs latency one did not add a lot of value on top of the table so I skipped it. 

### @mrscoopers

It is written for both people and agents. It assumes a reader who already has retrieval working, who knows vector search basics but not Qdrant internals, and who likely has an AI assistant open alongside. Each article stands alone for the decision it covers, but article 1 is the prerequisite for all five, so the intro states that dependency instead of repeating its checks.

1. Added `prefetch` and nested multi-stage limits. Kept the title, since the article covers depth wherever it appears.
2. Added the baseline dependency. Do you want the full series order spelled out there too?
3. Added the relationship up front, including that they are measured differently: `limit` against your labels, `hnsw_ef` against exact search.
4. Added a worked example: one query, three relevant documents at ranks 4, 30, and 180.
5. Moved the latency paragraph up, and the per-shard point is now an aside.
6. Added the diagram, see above.
7. Cut it. The section now says `limit` is a query-time budget that leaves the footprint where it was.
8. Added the comparison: depth 500 to 10 removes 27% to 30% of median latency and frees no memory, while int8 quantization stores the vectors at one-quarter the size for at most 0.0001 of fused `nDCG@10`.
9. Distributed `m`, `ef_construct`, and ACORN under `hnsw_ef`. Cut `is_tenant` and `quantile`, since neither is a depth decision.

A manual pass also tightened the article

kanungle:
Article 2 LGTM

Consider pre-pending this paragraph with a boldface "Note:"

>The measurements in this article come from five public datasets between 5,183 and 100,000 documents. Each ran unquantized on one shard in a laptop Docker container, using all-MiniLM-L6-v2 and Qdrant’s core BM25.

kanungle:
@mrscoopers feel free to continue reviewing the other 4 articles as time permits you

kanungle:
**Article 3:**

In general, I really enjoyed this article and am glad we're finally addressing the k-value defaults in RRF. These are my minor comments:

- Qdrant defaults to k=2 for RRF while everyone else defaults to k=61. This article answers why later in the piece, but should give a nod early on to "continue reading to learn more about k values"
- This doesn't seem to be illustrated well by the data:
> The second prefetch costs a second index, a second vector per point, and 0.6 to 1.5 ms of query time in these single-shard measurements. Keep it when it improves relevance on your own labels.
- Your sentence could benefit from specific "increments of 5" or something, and why:
> Start with RRF at k=2 and equal weights, then run DBSF. If RRF wins, test k values from 2 to 61.
- The bold values in the RRF k-values table is hard to see. Can we style any better?
- Could more be done to explain the weighting guidance?
- Put a "Takeaways" section at the end

kanungle:
I think Article 3 could also use a refresher on RRF vs DBSF. I know it's covered elsewhere, but it would be a valuable addition to complete this piece

Dylancouzon:
> Consider pre-pending this paragraph with a boldface "Note:"

Applied to all 5 articles

kanungle:
Article 4 needs a manual writing pass. 

Also we should include a blog in this PR, summarizing each article's key findings/advice, then linking to them

Dylancouzon:
# Article 3 

@kanugle's feedback

**k nod early.** Added right after the first table, no anchor link so nobody skips ahead. One correction: the common default is 60, not 61. `k=61` is what 60 becomes under Qdrant's zero-based positions, which is what the porting note covers.

**Second prefetch cost.** The per-dataset latency was measured but never reached the page. The first table now carries a `Second Prefetch Cost` column, +0.60 to +1.47 ms per row.

**k increments.** Now "sweep `k` over 1, 2, 5, 20, and 61", with the reason: rank 1 outweighs rank 10 by 5.50x at `k=2`, 2.80x at `k=5`, 1.45x at `k=20`, and 1.15x at `k=61`. Even steps of 5 spend most of the sweep where the curve is already flat.

**Bold in the table.** It's a theme bug, `_qdrant-post.scss` sets `table { font-weight: $font-weight-base }`, which flattens every `<strong>` inside a table site-wide. That's a theme change, so I kept it out of this PR and starred the winning cell instead.

**Weighting guidance.** Three paragraphs from a weight sweep that existed on disk but was unpublished: equal weights win outright on two of five datasets, the two that gain clear their intervals at 0.0096 and 0.0060, and the winning direction does not follow from which prefetch is stronger.

**Takeaways.** Added as "What to Tune Next", matching candidate-depth's closer: the tuning order in one paragraph rather than a recap.

**RRF vs DBSF refresher.** Expanded from two sentences to two paragraphs, covering what each one reads and why RRF is indifferent to score scale while DBSF is not.

## Also changed

1,915 words, up from 1,644. Everything added came from this feedback, so I paid for it elsewhere: the standalone int8 quantization section is now one sentence inside a new scope aside, the held-out section is tighter, and six repeated statements went (the retrieval stack was described twice, "unquantized" three times, the score formula printed twice). The same pass cut density, so DBSF's normalization reads "using each list's average score and how spread out its scores are" rather than "against that prefetch's mean and standard deviation for the query", and the weights section leads with the lesson instead of our numbers.

kanungle:
Article 3 LGTM

The last heading "What to Tune Next" might need a revision per the new closing comments

kanungle:
**Article 5:**

- This is not true with `turbo4` datatype (which you mention later, but maybe worth clarifying early):
>In hybrid search, quantization can keep a compressed copy of the dense vectors in RAM while the original dense vectors stay on disk.
- This would read easier as bullets
> cold data loads lazily from disk, so the first request that needs a page waits for it. cached data enters the page cache when the collection loads, but the kernel may evict it later. pinned data stays in RAM, so the structure has to fit.
- Feels like AI (if you want to revise):
> The read column explains the gap. At 4 GiB, rescoring read far more data than the selected vectors themselves require because storage reads pages, not individual vectors. That amplification is why the latency increase is much larger than the rescore candidate set suggests.
- Feels like AI (if you want to revise):
>cached asks the operating system to warm data at startup; it cannot keep data resident past the memory cap. Pin the quantized vectors, leave the original vectors cold, and verify by tracking recurring original-vector reads and block reads under your production memory limit.
- This "Turbo4" should be `turbo4`:
>This article does not measure Turbo4, so validate it separately on your own queries and labels.
- Rewrite this:
>This is an evaluation contract, not a complete script.
- In the closing, point back to the 1st article again to close the loop for a reader coming in fresh

kanungle:
Majority of my feedback is stylistic, with just some areas for added clarification. Overall good work

kanungle:
Majority of my feedback is stylistic, with just some areas for added clarification. Overall good work

kanungle:
## **Article 1:**

- `pytrec_eval` is an external dependency that can change. Typically we try not to rely on these. I feel it's okay to leave in but probably worth omitting the code snippet

Otherwise LGTM

Status:
Article 1 - Approved with above change
Article 2 - Approved
Article 3 - Approved
Article 4 - Needs manual prose pass
Article 5 - Requires above changes

Dylancouzon:
> The last heading "What to Tune Next" might need a revision per the new closing comments

Done

> Article 4 needs a manual writing pass.

True, it wasn't at the same level as the others, I improved it and did a prose pass. I'm rerunning experiments overnight to see if different models yield better results. Ultimately, we can't show strong numbers because the baseline is too clean, a point I noted in my original PR message.

> Blog 

Will do

kanungle:
# **Article 4:**

Reads pretty well and I like it. Only stylistic feedback really:

- remove "cost" from:
>  Every candidate costs a forward pass at query time, so this cost rules it out as a first stage.

- remove the first code snippet? I don't feel it adds much
- There's a lot of talk about "cost" and "budgets" which feel like AI talking about tradeoffs
- hard to read the second table. Chart maybe?

mrscoopers:
> @mrscoopers feel free to continue reviewing the other 4 articles as time permits you

Article 2 done yesterday, afaik Dylan already applied fixes, doing 3 today:) cc @kanungle

mrscoopers:
"How to Tune Hybrid Search in Qdrant"
1. "RRF and DBSF..." section could use a visual (afaik we have some already)
2. "Use Labels to Choose a k Range" section: is it possible to keep the table without DBSF and introduce it before this section in a separate subsection? Otherwise, DBSF comments jump out at you without too much context. "Compare RRF and DBSF on Your Labels" suggests "Start with RRF at k=2 and equal weights, then run DBSF. If DBSF wins, take it forward to held-out validation. If RRF wins, tune k next.", so we could do a DBSF subsection first with k=2 from the table, and then transition to "Use Labels to Choose a k Range" & tune weights. The fact that RRF loses is alright; we can explain that here we demo all possible techniques
3 "A weight sweep often returns to equal weights. " - I still wouldn't generalize it to "often in production" from our experiments. Also, this section misses the weights tuning motivation part a bit; sometimes (often) users have an unstoppable will to weight something based on some intuition or business rules. Here we need to state that we have no prior knowledge of what we prefer and sweep weights only to optimize results on our labeled set. Also, I heard a very interesting talk on MICES where better hybrid search metrics (compared to pure lexical) in prod led to customers' dissatisfaction https://youtu.be/qiVbbReNQUo
4 "The single-prefetch scores do not tell you which way to lean, so sweep both. " - Can we explain the intuition behind this, besides numbers?
5 Also, I believe this (x, y) style of weights reporting might be slightly confusing; I'd reiterate at the beginning of the section  what it means & make the whole section more instruction-like (advice building up on each other)
6 In "Tune in This Order" I'd rather see a list

Overall, a very nice one!

Dylancouzon:
Article 4 has been updated and is much stronger now. Feedback has also been applied. 
Ready for review. @kanungle @mrscoopers

kanungle:
## **Article 4:**

You removed lines like:

>If reranking beats default RRF, [tune fusion](/articles/how-to-tune-hybrid-search/) next. Add reranking only when it improves the strongest first-stage ranking on held-out labeled queries.

and

>If you find a mismatch, swap in a model whose window and training data fit your documents and rerun the 10-candidate test. If the model fits and still loses, keep the tuned first stage and spend the tuning budget elsewhere.

and you replaced them with experiment results. This causes the article to read more like a benchmark report and less like guidance, which I don't like.

Could you rework the narrative of the article to be more guidance-oriented?

Also, can you adjust the table header text length in the below table to make the table read easier?
image

kanungle:
Also I strongly agree with @mrscoopers advice to add more visual and context to DBSF in Article 3. That's actually the direction I was going when mentioning to provide refreshers on the topic 🙂

Dylancouzon:
Article 1 pytrec_eval removed for manual ndcg calculation, broke down the snippets, and made the more clear. Let me know if that works for you @kanungle

Dylancouzon:
@mrscoopers Article 3

  1. Added a figure of the two methods
  2. Done. DBSF has its own result table right after the code
  3. Dropped "often". It now says equal weights won outright on two of five, and the section says to let the labels pick the direction. Left the MICES talk out since the article doesn't measure production satisfaction, happy
  to link it as a pointer if you want it.
  4. Done
  5. The weights section now opens by defining the pair, then `k`, then the sweep.
  6. Now a numbered list.

Dylancouzon:
@kanungle Article 4: 
Reworked the narrative so each section leads with the decision and the numbers sit under it, the two guidance lines you quoted are back, and Table headers shortened on both tables.

I also put the reranking code back, as three short blocks instead of the one long snippet you flagged earlier. Better technical depth and parity with the other articles.

Dylancouzon:
@kanungle Article 5 fixes are in

Dylancouzon:
Did a consistency check on the five articles and removed experiments/additional files.

As we approach merge, inline comments for feedback are appreciated to save time on re‑reading.

Dylancouzon:
@kanungle here's the [promotion blog](https://deploy-preview-2634--condescending-goldwasser-91acf0.netlify.app/blog/tuning-retrieval-which-knob-first/)
Let me know if this is what you had in mind

mrscoopers:
# **Article 4:**

Line refs against `when-a-reranker-is-worth-it.md`.

1. **L27** ("A cross-encoder reranker reads the query and candidate together") — Somehow reranking reads in this article as "reranker = only cross-encoder" until "Use Other Stages for Different Problems". (The SEO description on L4 leans the same way.)

2. **L92-94** ("If reranking beats default RRF, tune fusion next.") — "A reranker measured against the default can look like a win" → "If reranking beats default RRF, tune fusion next." didn't we advise to compare reranking directly to tuned fusion (default fusion → tuned → reranking)?

3. **L118** ("the prefetches scored the query and page separately and fusion sees only their ranks") — good example but hard to read as a block of text & not truly clear as an explanation. If both arms matched the doc perfectly, it'd be top-1, I assume lexical should, idk about the dense one. Explanation reads as "fusion is bad here bcs it's not a cross-encoder, and cross-encoder is good here bcs it's a cross-encoder".

4. **L122** ("check model fit before you change anything else") — imo I'd advise this before running a reranker, at least the "read model card" part & at least think about the context window part, maybe not in detail, but a sensibility check. Feel free to disagree though. (Would sit in step 2, L36.)

5. **L25 / results table L104-110** ("At 200 candidates, that nDCG@10 gap ran from 0.247 to 0.487") — I feel like we go back and forth from the 10 candidates we advise to the 200 we experiment on, so it might be confusing? Do we advise to rerank 10 or measure top-10 reranked?

6. **L136** ("A higher count refines a win, and it rescued no loss here") — "Set Candidate Count After a Win" would be nice to explain the logic why increasing the amount of candidates won't always bring you better ranking in the top even if your reranker seems very good (aka pouring money at the top-relevance problem doesn't work).

7. **L176** ("Test ColBERT when a cross-encoder is too slow") — Instead of ColBERT I'd just say "late interaction models" & we have some tutorials to link here. (Also L165 in the symptom table.)

Overall I like the article.

mrscoopers (commented):
Overall I like the article!

qdrant-landing/content/articles/when-a-reranker-is-worth-it.md
```diff
@@ -0,0 +1,182 @@
...
+
+Your candidate list can already contain documents your ranking never shows. Score those candidates as if they were perfectly ordered, then compare that with the score your pipeline returns today. The gap between the two is everything a better ranking stage could recover, so measure it before you reach for a model. Use `nDCG@10`, which grades the top 10 results and gives more credit to relevant documents near the top.
+
+A wide gap means a better order is worth chasing. At 200 candidates, that `nDCG@10` gap ran from 0.247 to 0.487 across the five datasets in this article, and [candidate depth](/articles/candidate-depth/) shows how to measure it on your own collection.
+
+A cross-encoder reranker reads the query and candidate together as a single sequence, with a classification head that returns one relevance score for the pair. Joint reading lets it model token interactions that separately encoded query and document vectors miss. Every candidate takes a forward pass at query time, which rules the model out as a first stage and keeps it in the reranking slot.
```
mrscoopers: Somehow, reranking reads in this article as "reranker = only cross-encoder" until "Use Other Stages for Different Problems". (The SEO description in the frontmatter leans the same way.)

qdrant-landing/content/articles/when-a-reranker-is-worth-it.md
```diff
@@ -0,0 +1,182 @@
...
+after = ndcg_at_k([point.payload["doc_id"] for point in reranked], relevance)
+```
+
+Run that over your labeled queries, average the per-query difference, then check the interval around it before you trust the direction.
+
+## Compare with the Best First Stage
+
+Compare the reranker against the strongest first stage you can build. Qdrant's default reciprocal rank fusion (RRF) is already a solid baseline, and fusion tuned on your own labels is stronger. A reranker measured against the default can look like a win that tuning would have delivered for far less work at query time.
+
+If reranking beats default RRF, [tune fusion](/articles/how-to-tune-hybrid-search/) next. Add reranking only when it improves the strongest first-stage ranking on held-out labeled queries.
```
mrscoopers: "A reranker measured against the default can look like a win" -> "If reranking beats default RRF, tune fusion next." didn't we advise to compare reranking directly to tuned fusion (default fusion -> tuned -> reranking)?

qdrant-landing/content/articles/when-a-reranker-is-worth-it.md
```diff
@@ -0,0 +1,182 @@
...
+
+Keep fusion when the reranker loses to the tuned first stage. WANDS gained +0.039 over default RRF and still lost to fusion tuned on the same candidates.
+
+Both confirmed wins came from the model whose window and training data fit the corpus, scoring the same candidates the other three saw. Even those wins closed part of the gap: 0.135 of a 0.293 gap on CodeSearchNet, and 0.115 of a 0.487 gap on DBPedia-entity.
+
+On DBPedia-entity, the query "John Lennon Yoko Ono album Starting Over" is answered by the page for "(Just Like) Starting Over", and every query term sits in that page's first two sentences. Fusion still left it at rank 49 of 200, because the prefetches scored the query and page separately and fusion sees only their ranks. The cross-encoder read the pair as one sequence and put it first.
```
mrscoopers: good example but hard to read as a block of text & not truly clear as an explanation. If both arms matched the doc perfectly, it'd be top-1, I assume lexical should, idk about the dense one. Explanation reads as "fusion is bad here bcs it's not a cross-encoder, and cross-encoder is good here bcs it's a cross-encoder".

qdrant-landing/content/articles/when-a-reranker-is-worth-it.md
```diff
@@ -0,0 +1,182 @@
...
+
+## Diagnose a Loss Before You Stop
+
+If the reranker loses at 10 candidates, check model fit before you change anything else. The first check is the context window. Tokenize a sample of your query-document pairs with the model's tokenizer, and compare the 95th percentile length against the model's window: a longer pair gets truncated, and the model scores a document it only partly read. The second is the training data, so read the model card for its languages and domains.
```
mrscoopers: imo I'd advise this before running a reranker, at least the "read model card" part & at least think about the context window part, maybe not in detail, but a sensibility check. Feel free to disagree though. (Would sit in step 2 above.)

qdrant-landing/content/articles/when-a-reranker-is-worth-it.md
```diff
@@ -0,0 +1,182 @@
...
+  - search relevance
+  - FastEmbed
+category: search-quality
+---
+
+Before you tune a reranker, use the [pre-tuning checks](/articles/before-tuning-a-qdrant-collection/) to verify index state and set a labeled baseline.
+
+Your candidate list can already contain documents your ranking never shows. Score those candidates as if they were perfectly ordered, then compare that with the score your pipeline returns today. The gap between the two is everything a better ranking stage could recover, so measure it before you reach for a model. Use `nDCG@10`, which grades the top 10 results and gives more credit to relevant documents near the top.
+
+A wide gap means a better order is worth chasing. At 200 candidates, that `nDCG@10` gap ran from 0.247 to 0.487 across the five datasets in this article, and [candidate depth](/articles/candidate-depth/) shows how to measure it on your own collection.
```
mrscoopers: I feel like we go back and forth from the 10 candidates we advise to the 200 we experiment on (here and in the results table), so it might be confusing? Do we advise to rerank 10 or measure top-10 reranked?

qdrant-landing/content/articles/when-a-reranker-is-worth-it.md
```diff
@@ -0,0 +1,182 @@
...
+
+{{< figure src="/articles_data/when-a-reranker-is-worth-it/reranker-gain-by-candidate-count.png" alt="Five small line charts, one per dataset, showing the best nDCG@10 change over tuned fusion at candidate counts 10, 25, 50, 100, and 200. SciFact, CodeSearchNet, and DBPedia-entity stay above the zero line, WANDS stays below it at every count, and ArguAna peaks at 25 then falls to zero by 200." caption="The best nDCG@10 change over tuned fusion among the four models, by candidate count. A line above zero is a reranker win; WANDS never crosses it." width="100%" >}}
+
+Each dataset stopped climbing at a different point, which is why the count is measured rather than assumed. DBPedia-entity reached 0.111 of its eventual 0.115 by 50 candidates, so going to 200 quadrupled the reranking work for 0.003 more. CodeSearchNet kept climbing through 200. ArguAna peaked at 25 and became a loss by 200, so more candidates can also hurt.
+
+A higher count refines a win, and it rescued no loss here: every configuration behind tuned fusion at 10 was still behind it at 200.
```
mrscoopers: "Set Candidate Count After a Win" would be nice to explain the logic why increasing the amount of candidates won't always bring you better ranking in the top even if your reranker seems very good (aka pouring money at the top-relevance problem doesn't work).

qdrant-landing/content/articles/when-a-reranker-is-worth-it.md
```diff
@@ -0,0 +1,182 @@
...
+
+Grouping fits collections that store each chunk of a document as its own point. `query_points_groups` with `group_by` on the document ID field returns the best chunk per document, so one long document cannot fill the first page. The grouped field needs a [payload index](/documentation/manage-data/indexing/#payload-index); without one on `document_id`, Qdrant Cloud returns a 400.
+
+Formula Query rescores the same candidates with an expression over payload fields, such as recency or popularity, and needs a payload index on each field the formula references.
+
+Test [ColBERT](https://arxiv.org/abs/2004.12832) when a cross-encoder is too slow for your queries. It moves the reranking work from query time to ingest: document vectors are built once, so only the query goes through the model per request, and storage grows to a vector per token. Reranking needs those vectors but never traverses a graph over them, so set [`m=0`](/documentation/search/hybrid-queries/#multi-stage-queries) on the multivector and Qdrant stores it unindexed. ColBERT rescoring also runs inside Qdrant, in the same multi-stage query, while a cross-encoder runs as its own service after Qdrant returns candidates.
```
mrscoopers: Instead of ColBERT I'd just say "late interaction models" & we have some tutorials to link here. (Also the symptom table row above.)

qdrant-landing/content/articles/when-your-collection-outgrows-ram.md
```diff
@@ -0,0 +1,185 @@
...
+keywords:
+  - memory tiers
+  - quantization
+  - rescoring
+  - oversampling
+  - TurboQuant
+category: search-quality
+---
+
+Once a collection no longer fits in RAM, decide which structures stay resident and whether rescoring is worth a disk read. In hybrid search, quantization can keep a compressed copy of the dense vectors in RAM while the original dense vectors stay on disk. If rescoring is enabled, Qdrant reads those original vectors after the dense prefetch to repair compression errors. The same mechanism applies to dense-only search. The [`turbo4` datatype](/documentation/manage-data/vectors/#turbo4) stores one 4-bit copy and has no original vectors to reread. A later section covers it separately.
```
mrscoopers: Two things about the opening. "whether rescoring is worth a disk read" might not be clear for people not super well-versed with Qdrant, and it's the first line, maybe a smoother intro into the problem? And "The turbo4 datatype stores one 4-bit copy and has no original vectors to reread" adds to it: we give a lot of details about details up here and I think one can lose what this is all about. I'd structure the intro as: what it means for a collection to outgrow RAM -> one option is quantization on the dense arm -> a quick reminder of what that is and what it is in Qdrant (rescoring is used with no explanation, same with `cold`/`pinned` before we introduce them) -> then the rest.

qdrant-landing/content/articles/when-your-collection-outgrows-ram.md
```diff
@@ -0,0 +1,185 @@
...
+
+The measurements isolate the dense path. Rerun your full hybrid query before you use them to set a production latency budget.
+
+Use `nDCG@k` on a [labeled baseline](/articles/before-tuning-a-qdrant-collection/) to compare quantization settings in your hybrid query, where `k` matches the result count your product evaluates. Select a configuration on one part of that set, then confirm it on held-out queries that did not take part in selection. Use `Recall@k` against exact search to measure what quantization changes in the dense prefetch. The latency and Recall@k checks also apply to dense-only search when you do not have labels.
```
mrscoopers: This feels too dense for me, I'd provide the advice together with pseudo code or examples.

mrscoopers (commented):
Overall this one reads to me like a benchmark and a little bit all over the place, I don't see a build-up or a 1->2->3 instruction flow the way Article 4 has it. A few things:

- The title paints broader than the content. We really only cover dense quantization + rescoring; we don't touch the sparse arm, and another option to shrink an arm is e.g. not using full precision, which we don't mention. We don't have to give all the options, but then the title reads broader than "Quantization when your collection outgrows RAM".
- I miss at least one intro line where we recommend TurboQuant as the default option, and what it is / its parameters.
- Cloud vs local advice I'd try to visually separate more.

qdrant-landing/content/articles/when-your-collection-outgrows-ram.md
```diff
@@ -0,0 +1,185 @@
...
+
+## The Short Version
```
mrscoopers: The Short Version feels off: why 3 after 2? Why is TurboQuant treated as common knowledge? What's the logic behind the highlights? It feels like it's not understandable before reading the subsequent sections anyway, so not so much value in it.

qdrant-landing/content/articles/when-your-collection-outgrows-ram.md
```diff
@@ -0,0 +1,185 @@
...
+
+1. Measure rescoring at the memory cap you deploy with.
+2. Pin the quantized vectors and leave the original vectors `cold`. Under a limit that cannot hold them, the operating system evicts original vectors set to `cached` anyway.
+3. Choose a quantization method first. If you use TurboQuant, choose its `bits` value and `rescore` setting together. Lower `bits` values use less memory and introduce more approximation error. `rescore` rereads the original vectors to correct the top candidates, trading disk reads for `Recall@k` against exact search.
+4. Compare `nDCG@k` on held-out labeled hybrid queries. Use `Recall@k` against exact search to diagnose what changed in the dense prefetch.
+
+## Start with a RAM Estimate
```
mrscoopers: And what about the sparse part, how do we consider it here?

qdrant-landing/content/articles/when-your-collection-outgrows-ram.md
```diff
@@ -0,0 +1,185 @@
...
+
+- `cold` loads lazily from disk, so the first request that needs a page waits for it.
+- `cached` enters the page cache when the collection loads, and the kernel may evict it later.
+- `pinned` stays in RAM, so the structure has to fit.
+
+This measurement concerns the dense vectors and their quantized copy. In a hybrid query, the dense prefetch scores the quantized vectors during graph traversal, then rereads the original vectors during rescoring. The same placements apply to a dense-only query. Set both placements explicitly: the default placement for quantized vectors depends on the placement of the original vectors.
+
+Qdrant rejects `pinned` for dense vectors, leaving `cold` and `cached` as the available placements for the originals. The [memory tiers documentation](/documentation/ops-configuration/memory-tiers/) recommends this pairing. The rest of this article measures its latency and disk-read trade-off.
```
mrscoopers: This reads oddly, since the quantized copy is also a dense vector and we pin it two lines later. "original dense vectors"?

qdrant-landing/content/articles/when-your-collection-outgrows-ram.md
```diff
@@ -0,0 +1,185 @@
...
+
+## When Rescoring Improves Quality
+
+<aside role="status">
+Quality scope: the quality table uses the 200-query held-out set at Qdrant's default `memory` configuration. It does not report latency because sequential query passes warmed the operating system's page cache. The latency table reports the `memory` configurations shown there.
+</aside>
+
+An exact search scans every vector and gives the reference result. The graph search is approximate: it can miss exact neighbors in exchange for lower latency. `Recall@10` is the share of the exact top 10 that a configuration returned. `nDCG@10` grades the top 10 against DBPedia-entity's labels, giving more credit to relevant documents near the top.
+
+We selected a candidate on a separate labeled set: the lowest `oversampling` and most aggressive TurboQuant `bits` value within 0.01 `nDCG@10` and 0.02 `Recall@10` of float32. The table reports the held-out result.
```
mrscoopers: `oversampling` drives the quality table, the figure, and the per-shard math, but I don't think we ever define it.

qdrant-landing/content/articles/when-your-collection-outgrows-ram.md
```diff
@@ -0,0 +1,185 @@
...
+
+| Quantization | `rescore` | `nDCG@10` | `Recall@10` Against Exact |
+|---|---|---|---|
+| float32 | not applicable | 0.3103 | 0.957 |
+| TurboQuant `bits4` | off | 0.3218 | 0.918 |
```
mrscoopers, line 114: Quantized row score `nDCG@10` at or above float32(??)

qdrant-landing/content/articles/when-your-collection-outgrows-ram.md
```diff
@@ -0,0 +1,185 @@
...
+
+This article measures TurboQuant. For a comparison of TurboQuant bit depths across ten datasets, see [TurboQuant in Qdrant](/articles/turboquant-quantization/). If TurboQuant is not the right fit, Qdrant also supports [Scalar, Binary, and Product Quantization](/documentation/manage-data/quantization/). Choose the method that fits the compression, recall, and latency trade-off you need, then validate it with the same dense-prefetch and held-out hybrid checks.
+
+- [Scalar Quantization](/documentation/manage-data/quantization/#scalar-quantization): converts vector components to `int8`. Start here when moderate compression is enough.
+
+- [Binary Quantization](/documentation/manage-data/quantization/#binary-quantization): a compact, fast option that works best with high-dimensional embeddings whose components have a centered distribution. Measure whether rescoring recovers enough quality for your workload.
+
+- [Product Quantization](/documentation/manage-data/quantization/#product-quantization): prioritizes a smaller memory footprint, with a larger accuracy and search-speed trade-off to validate.
+
+### `turbo4` Removes the Rescoring Option
```
mrscoopers: This reads more like an alert (don't mix the two) than a section.

qdrant-landing/content/articles/when-your-collection-outgrows-ram.md
```diff
@@ -0,0 +1,185 @@
...
+
+The [`turbo4` datatype](/documentation/manage-data/vectors/#turbo4) is not TurboQuant. It stores a 4-bit dense-vector representation as the only copy, so there are no original vectors to rescore against.
+
+Use `turbo4` when disk capacity is the constraint and its measured quality meets your target. Use a full-precision vector with quantization when you need `rescore` to recover dense-prefetch quality. This article does not measure `turbo4`, so validate it separately on your own queries and labels.
+
+## Verify It on Your Own Collection
```
mrscoopers, line 155: This looks like it should have been placed earlier? Reading the titles only I get: "Start with a RAM estimate" -> <details about how settings affect results> -> "Consider Other Quantization Methods" -> <try TurboQuant, here in "Verify It on Your Own Collection">, which feels off.

qdrant-landing/content/articles/when-your-collection-outgrows-ram.md
```diff
@@ -0,0 +1,185 @@
...
+)
+```
+
+First, compute the exact dense top `k` once for a representative sample of your queries, where `k` matches the result count your product evaluates. This article reports `k=10`. An exact search reads every original vector, so keep the sample small enough for the cost you can accept. Then run your existing dense prefetch with each `rescore` and `oversampling` variant, changing no other search parameters. `Recall@k` against the exact result shows what quantization changes in the dense prefetch.
+
+For hybrid search, keep the dense prefetch, sparse prefetch, [fusion settings](/articles/how-to-tune-hybrid-search/), and filters that your service already uses. Compare the final result's `nDCG@k` on held-out labeled queries. That is the metric that decides whether the configuration serves your product.
+
+The steps above list what to measure. Use your existing request, or ask a coding agent to build a small harness with your dense vector name, current prefetches, fusion method, filters, query sample, and labels. It should sweep only `rescore` and `oversampling`, then report dense-prefetch `Recall@k`, final `nDCG@k`, and latency.
```
mrscoopers: Maybe we can give a good prompt here?

qdrant-landing/content/articles/when-your-collection-outgrows-ram.md
```diff
@@ -0,0 +1,185 @@
...
+
+On a self-hosted deployment, run the dense-prefetch check under the memory cap you deploy with, from a cold page cache followed by a measured pass. Run `rescore=False` even if you would never ship it, because it shows the cost of the rest of the dense prefetch. On Qdrant Cloud, measure the full request under its normal operating conditions instead.
+
+Keep the first configuration that meets your held-out `nDCG@k` and latency requirements. Use dense-prefetch `Recall@k` to explain a quality loss. If none qualifies, test another quantization method or a higher `oversampling` value.
+
+## Scope and Next Checks
```
mrscoopers: Tbh the whole "Scope and Next Checks" section doesn't read like scope & next checks, or like a conclusion/wrap-up. The Short Version would be more fitting here as a recap.

qdrant-landing/content/articles/when-your-collection-outgrows-ram.md
```diff
@@ -0,0 +1,185 @@
...
+
+<aside role="status">
+Scope: every figure comes from a dense-only request against one shard on a laptop, with Qdrant running in Docker's Linux VM behind macOS. The cgroup limit evicted mapped original vectors, and block-read counters confirmed recurring disk reads. Do not transfer the latency or disk-read ratios without measuring your own deployment.
```
mrscoopers: The scope alert looks weird here, what's its purpose? To say that when it comes to prod-level multishard we have no experiments to give specific advice, just some general considerations?

qdrant-landing/content/articles/when-your-collection-outgrows-ram.md
```diff
@@ -0,0 +1,185 @@
...
+</aside>
+
+On a multi-shard hybrid collection, measure the full request on your deployed shard layout. Each shard runs the dense prefetch and rescoring against its own data. With a `limit` of 200 and `oversampling=1`, rescoring can read up to 200 original vectors per shard: up to 2,400 across 12 shards. That total shapes disk reads and tail latency, and [candidate depth](/articles/candidate-depth/) covers how to set that limit.
+
+Qdrant's `cold` `memory` tier leaves original vectors on disk until a query accesses them. If you use it, set [`storage.performance.io_uring` to `auto`](/documentation/ops-configuration/memory-tiers/#async-io) in Qdrant v1.19 to issue reads asynchronously when the Linux kernel supports it. In hybrid search, the sparse prefetch shares the same page cache. Rerun the full request after you set the dense-vector `memory` configuration.
```
mrscoopers: This part feels like it could go in the section where we check the different memory placement configs. Here it feels out of context.

qdrant-landing/content/articles/when-a-reranker-is-worth-it.md
Dylancouzon: Done

qdrant-landing/content/articles/when-a-reranker-is-worth-it.md
Dylancouzon: Done

qdrant-landing/content/articles/when-a-reranker-is-worth-it.md
Dylancouzon: Done

qdrant-landing/content/articles/when-a-reranker-is-worth-it.md
Dylancouzon: Done

qdrant-landing/content/articles/when-a-reranker-is-worth-it.md
Dylancouzon: Clarified

qdrant-landing/content/articles/when-a-reranker-is-worth-it.md
Dylancouzon: Done

qdrant-landing/content/articles/when-your-collection-outgrows-ram.md
Dylancouzon: I'd say no. A prompt can date fast, and the paragraph already lists every input the harness needs

qdrant-landing/content/articles/when-a-reranker-is-worth-it.md
Dylancouzon: Done

qdrant-landing/content/articles/when-your-collection-outgrows-ram.md
Dylancouzon: Done

qdrant-landing/content/articles/when-your-collection-outgrows-ram.md
Dylancouzon: Done

qdrant-landing/content/articles/when-your-collection-outgrows-ram.md
Dylancouzon: Cut

qdrant-landing/content/articles/when-your-collection-outgrows-ram.md
Dylancouzon: Done

qdrant-landing/content/articles/when-your-collection-outgrows-ram.md
Dylancouzon: Done

qdrant-landing/content/articles/when-your-collection-outgrows-ram.md
Dylancouzon: Defined

qdrant-landing/content/articles/when-your-collection-outgrows-ram.md
Dylancouzon: Explained

qdrant-landing/content/articles/when-your-collection-outgrows-ram.md
Dylancouzon: Done, now a subsection

qdrant-landing/content/articles/when-your-collection-outgrows-ram.md
Dylancouzon: Reordered

qdrant-landing/content/articles/when-your-collection-outgrows-ram.md
Dylancouzon: Renamed "What to Tune Next" and opens on the decision rule

qdrant-landing/content/articles/when-your-collection-outgrows-ram.md
Dylancouzon: Merged into the setup note at the top

qdrant-landing/content/articles/when-your-collection-outgrows-ram.md
Dylancouzon: Moved into the placement section

Dylancouzon:
Article 5 got a substantial rewrite, I kept the title and extended the Scope. Thanks for the feedback

kanungle:
Article 4:

Apologies for not commenting inline, I'm on my phone, but these are the two areas I think could use improvement. In general they are too number-dense to really digest. Try talking about the multiplier gain or the percentage gain instead of long decimal numbers, which are hard to compute in one's head. Also focus not on the dataset but the dataset qualities that can be generalized (in these sections)

kanungle:
Same feedback on Article 5..

This is too results dense whereas I'm interested in generalized guidance :

kanungle:
And on Article 5, I'd remove lines like

> Carry over that ratio rather than the absolute milliseconds.

kanungle:
Those were the only issues I saw in articles 4 and 5

kanungle:
## **Article 2**

There's a lot of talk about datasets on here and their results but what if you don't know the datasets? It's kind of meaningless to say "WANDS does this" and "a scifact does that"

If youre going to talk about hm guidance in relation to tested data sets I think that makes sense but there should be something at the beginning that at least describes the nature of these datasets. 

Lastly and this is minor, but the below snippet isn't needed

Dylancouzon:
@kanungle Article 2 I added information about the datasets, kept it short because this isn't the point of the article. 

Rewrote the "Raise hnsw_ef Only When Recall Is Still Climbing" section, I still kept a short code snippet, it gives people instructions on how to do it, but made this section a lot more clean.

Dylancouzon:
Article 4 feedback applied

Dylancouzon:
Article 5 updated, did a full line by line manual review :)

Dylancouzon:
Did a full blog rewrite and updated the dates for a Monday publication.

qdrant-landing/content/articles/how-to-tune-hybrid-search.md
```diff
@@ -0,0 +1,178 @@
...
+
+{{< figure src="/articles_data/how-to-tune-hybrid-search/fusion-signals.png" alt="Two panels of dot plots, RRF on the left and DBSF on the right. Each panel has a dense line, a sparse line, and a fused line holding documents A, B, C, and D. The RRF lines space every document evenly and label the slots 4, 3, 2, 1. The DBSF lines keep the raw score spacing on one shared axis, dense running 0.55 to 0.91 with A far out to the right and B, C, and D clustered, sparse running 12.9 to 14.8. The fused lines put B first under RRF and A first under DBSF." caption="RRF reads each document's slot, so A's dense lead flattens to one step and B, ranked near the top by both prefetches, wins. DBSF keeps the spacing on a shared axis, so A's lead survives the sum and A wins." width="100%" >}}
```
kanungle: Use markdown image insertion:

`image-text`

qdrant-landing/content/articles/how-to-tune-hybrid-search.md
```diff
@@ -0,0 +1,178 @@
...
+Use your [labeled query set](/articles/before-tuning-a-qdrant-collection/#make-sure-your-labels-can-detect-a-gain) to compare RRF and DBSF over the same prefetches. Run RRF at `k=2` and equal weights, then run DBSF.
+
+```python
+from qdrant_client import QdrantClient, models
+
+client = QdrantClient(
+    url="https://YOUR-CLUSTER.cloud.qdrant.io",
+    api_key="<your-api-key>",
+)
+```
```
kanungle: This snippet doesn't seem necessary to me

qdrant-landing/content/articles/how-to-tune-hybrid-search.md
```diff
@@ -0,0 +1,178 @@
...
+| DBPedia-entity | 0.4822 | +0.0184 |
+| WANDS | 0.7637 | +0.0383 |
+
+DBSF takes no parameters: `k` and the weight pair are RRF settings, and the public API accepts them only on an `RrfQuery`. So if DBSF wins on your labels, skip the next two sections and go to the held-out check.
+
+## Use Labels to Choose a `k` Range
+
+Qdrant scores a document at position `pos` in one prefetch as `1 / ((pos + 1) / weight + k - 1)`, then sums across prefetches. With equal weights that reduces to `1 / (pos + k)`, and `k` alone decides how steeply the head of a list outranks its tail.
+
+{{< figure src="/articles_data/how-to-tune-hybrid-search/rrf-k-rank-weight.png" alt="Grouped bar chart comparing the share of a retrieval prefetch's top-10 score mass at each rank, for k equal to 2 and k equal to 61. At k=2 rank 1 takes 24.8 percent and rank 10 takes 4.5 percent. At k=61 the shares are nearly flat, 10.7 percent at rank 1 and 9.3 percent at rank 10." caption="At Qdrant's default of k=2, rank 1 carries 5.50 times the score weight of rank 10. At k=61, it carries 1.15 times the weight, so a candidate's presence in a prefetch matters almost as much as its position." width="100%" >}}
```
kanungle: Use markdown image insertion

qdrant-landing/content/articles/how-to-tune-hybrid-search.md
```diff
@@ -0,0 +1,178 @@
...
+
+| Dataset | Queries | Relevant per Query | k=1 | k=2 | k=5 | k=20 | k=61 |
+|---|---|---|---|---|---|---|---|
+| ArguAna | 1,401 | 1.0 | 0.5171 | 0.5216 | 0.5304* | 0.5269 | 0.5207 |
+| CodeSearchNet | 1,000 | 1.0 | 0.6501 | 0.6555 | 0.6580* | 0.6511 | 0.6258 |
+| SciFact | 300 | 1.1 | 0.7117 | 0.7175* | 0.7154 | 0.7122 | 0.7067 |
+| DBPedia-entity | 400 | 38.2 | 0.4625 | 0.4638 | 0.4641 | 0.4682* | 0.4606 |
+| WANDS | 480 | 358.9 | 0.7232 | 0.7254 | 0.7336 | 0.7571 | 0.7614* |
+
+On WANDS, `k=2` and `k=61` chose a different top result for 202 of 480 queries, while `nDCG@10` rose by 0.0360. A small aggregate gain can still change what a user sees first.
```
kanungle: replace:

> 202 of 480 queries

with:

> 42% of queries

qdrant-landing/content/articles/when-a-reranker-is-worth-it.md
```diff
@@ -0,0 +1,188 @@
...
+
+## Set Candidate Count After a Win
+
+Start with 10 candidates, and confirm on your labeled queries that the reranker beats tuned fusion before you change the count. Every configuration that trailed tuned fusion at 10 candidates still trailed it at 200, so a deeper list does not rescue a reranker that loses at 10. `nDCG@10` grades the same top 10 results at every count, so the count changes only what the reranker gets to choose from.
+
+{{< figure src="/articles_data/when-a-reranker-is-worth-it/reranker-gain-by-candidate-count.png" alt="Five small line charts, one per dataset, showing the best nDCG@10 change over tuned fusion at candidate counts 10, 25, 50, 100, and 200. SciFact, CodeSearchNet, and DBPedia-entity stay above the zero line, WANDS stays below it at every count, and ArguAna peaks at 25 then falls to zero by 200." caption="The best nDCG@10 change over tuned fusion among the four models, by candidate count. A line above zero is a reranker win; WANDS never crosses it." width="100%" >}}
```
kanungle: Use markdown image insertion

qdrant-landing/content/articles/when-your-collection-outgrows-ram.md
```diff
@@ -0,0 +1,201 @@
...
+
+What rescoring recovers depends on how much precision the bit depth discarded. At `bits4` it lifted dense-prefetch `Recall@10` from 0.918 to 0.993, while 200 held-out queries did not establish an `nDCG@10` difference. Recovered neighbors can improve dense-prefetch recall without improving the labeled top 10.
+
+At a deep bit depth, rescoring is what makes the quantization usable. One pass raised `bits1` from 0.605 to 0.951 `Recall@10`. Qdrant [enables `rescore` by default](/documentation/manage-data/quantization/#searching-with-quantization) for `bits1`, `bits1_5`, `bits2`, and binary quantization for this reason.
+
+{{< figure src="/articles_data/when-your-collection-outgrows-ram/bits1-rescore-recovery.png" alt="Line chart of the share of the exact top 10 that bits1 returns, across rescore off and rescore on at oversampling 1, 2, and 4. The share jumps from 0.605 with rescore off to 0.951 at oversampling 1, crossing the dashed float32 reference at 0.957, then flattens at 0.977 and 0.988." caption="One rescoring pass does most of the recovery at bits1. Raising oversampling past 1 buys little, which is why the disk reads it adds are the cost to watch." width="100%" >}}
```
kanungle: Use markdown image insertion

kanungle (approved):
Looks excellent. Thanks for the hard work and bearing through these changes. I think these are good additions to our technical articles.

I left a couple nits worth considering, and a few of your article images were embedded the wrong way.

Other than that, looks good. Approved.

PR #2643, opened 2026-08-19
docs(tutorial): OpenTelemetry End-to-End with Qdrant Cloud

anush008:
## [PREVIEW](https://deploy-preview-2643--condescending-goldwasser-91acf0.netlify.app/documentation/tutorials-build-essentials/opentelemetry-end-to-end/)

cc @meinsta, does this cut it?


meinsta:
Reviewed this end to end against a 1.17.1 Cloud cluster. The Collector bridge is the piece that was missing from our docs and it's right: two scrape jobs, `per_collection=true` on the node job, and the correlation walkthrough all hold up.

I found four things that would have shipped broken, and rather than leave them as a round trip I fixed them in #2687, which is already merged into this branch. Writing them up here so the reasoning is on the record.

**1. `restore.py` failed as written.** `recover_snapshot` defaults to `wait=True`, and restoring 18k points outruns the client's default timeout, so the reader gets `ResponseHandlingException` and a stack trace. Worse, the restore finishes server-side, so re-running hits the `collection_exists` guard and silently no-ops — it looks like the script never worked. Fixed with `timeout=600` on the client.

**2. Step 7's payoff didn't land.** This was the important one. `app.py` picked the slowest span by client wall-clock, but measured over 20 requests per shape the two query shapes sit ~1% apart at the client while in-database time moves 26%:

| Query shape | Client median | Mean time inside Qdrant |
|---|---|---|
| `docs`, `limit=10` | 312.6 ms | 2.71 ms |
| `blog`, `limit=1000` | 316.2 ms | 3.42 ms |

Under 1% of what the client waits for is spent in the database. So "slowest span" was usually a network outlier on the *cheap* query, and the reader opened a trace in Tempo with no cluster-side story behind it — teaching people to blame the database for a transport problem, which is the exact confusion this page should be clearing up. Step 7 now has a "Read the ratio, not the total" section with those numbers and tells readers to rank by `rest_responses_duration_seconds` when they want work the database actually did.

**3. The layer boundary was never stated.** Qdrant emits database metrics only; token counts, model latency and retrieval relevance come from a tracing library in the app. "Do you integrate with Datadog / New Relic / Arize?" is one of the most common questions we get on calls and the answer wasn't written down anywhere. Now a "Which layer reports what" section above the fold.

**4. The Grafana dashboard was missing,** while the page told readers to look at `traefik_service_request_duration_seconds`. Now linked, flagged as community-maintained rather than a supported surface, with a warning that several API-latency panels query `envoy_cluster_upstream_rq_*` and render empty on Traefik clusters.

Also added: the vendor rows are marked endpoint-and-header only and untested, there's a "Verified against" section that names the three paths which were *not* verified, and an aside noting `per_collection=true` is documented as v1.18+ but works on 1.17.1 with histograms included — without which a reader on an older cluster gets an empty PromQL result and nothing to explain why.

The Kubernetes section landed too, since every account that raised this on a call was already running Prometheus in Kubernetes. The specific trap is that a kube-prometheus-stack `ScrapeConfig` against `/sys_metrics` gives you node and edge metrics but not `collection_running_optimizations` — the one metric you most need is the one that silently goes missing.

Two corrections to things I raised earlier and got wrong: `otlp_grpc` is a valid registered alias in collector-contrib, and the blog section holds 3,330 points, not under 1,000.

Content-wise this is done from my side. CI is green and there are no conflicts.

meinsta (approved):

PR #2646, opened 2026-08-20
fix(docs): prevent iOS Safari from navigating on sidebar section tap

abdonpijpelink:
On iOS/iPadOS, tapping a collapsed section chevron in the documentation
sidebar navigated to the section page instead of expanding the nav, and
closed the hamburger menu as a side effect.

The root cause is that every `<summary>` in the sidebar contains an `<a>`
tag. Safari on iOS treats a tap on `<summary>` as a link click rather than
a `<details>` toggle when an `<a>` is present inside it.

The fix is a small `click` handler in `documentation.js` that intercepts
the tap when the section is collapsed, calls `preventDefault()`, and
manually sets `details.open = true`. When a section is already open,
normal link navigation is allowed so users can still reach the section
root page.

## Validation

I've checked that this fixes it on iPhone and iPad.

### DIFF
--- a/qdrant-landing/themes/qdrant-2024/assets/js/documentation.js
+++ b/qdrant-landing/themes/qdrant-2024/assets/js/documentation.js
@@ -16,10 +16,13 @@ document.addEventListener('DOMContentLoaded', () => {
   document.querySelectorAll('.docs-menu details > summary > a').forEach(function (link) {
     link.addEventListener('click', function (e) {
       const details = this.closest('details');
-      if (!details.open) {
-        e.preventDefault();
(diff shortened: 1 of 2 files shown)


trean:
@abdonpijpelink But now, if I want to go directly to the subsection index page, I need to click the link twice. Let's keep expanding functionality to the chevron only. WDYT?

abdonpijpelink:
Good catch! I've pushed a change that wraps each title text in a `<span />`, so we can check whether the click is on the chevron or on the text.

trean (approved):
Thank you!

PR #2644, opened 2026-08-19
Add a customer quote component with customer avatar and logo

KShivendu:
I'm suggesting a small improvement in our testimonials page to highlight the quote from the customer possibly with their photo and standardizing the same component. Imo, that's the most important part in a testimonial. Most people skip architecture diagram but quotation catches their eye. 

This PR adds a quote component and applies it to three case studies as a starting point:

- **Bayer** — all twelve quotes converted
- **Tripadvisor** — the "2-3x revenue" quote moved up from the bottom of the page
- **Sprinklr** — quote lifted out of the paragraph it was buried in

Every word is already published, so nothing here needs new customer sign-off. I'll apply this component to all testimonials in the next PR if you agree.

2-tripadvisor

PR #2648, opened 2026-08-20
Apply the quote component to more case studies, and add a script to find the rest

KShivendu:
Follows on from #2644, which added the quote component. This applies it to three more case studies.

**OpenTable, &AI and Dust** — 16 quotes, all verbatim. Only quotes that say something about Qdrant were lifted; setup and problem-statement quotes stay in the prose they were written for.

Where the original splits a quotation across "said Turner", the card takes only the contiguous half about Qdrant and the setup half stays in the prose, so nothing is joined and no words change.

Dust carried the same quote twice: once inside the alt text of `Dust-Quote.jpg`, and again as prose below it. The image is gone and the words are a card, so that quote is text rather than pixels and no longer duplicated.

**Every card now shows the customer logo**, not just the first one in each article. That covers the studies merged in #2644, so Bayer and Sprinklr pick it up too. Worth knowing this hides the company name on those cards, since the component drops it when a logo is present — the name still shows on narrow screens where the logo is hidden.

## On lead quotes

A lead quote has to make sense to someone who has read only the title, and it belongs after the opening paragraph rather than above it.

I got this wrong first time on OpenTable — "Creating a Qdrant Cloud cluster was one of the easiest parts of the project" was sitting above every line of prose, where it read as a punchline with no setup. It's back where its context is, and the article's existing quote about why OpenTable built Concierge is the lead instead, without being moved anywhere.

## Still to do

Around 190 quotes are still in prose across the other case studies. Not all of them deserve a card — a pull quote has to make a claim worth stopping for, and plenty of what is quoted is context, an example search query, or a company slogan.


KShivendu:

| Case study | Before | After | Cards |
|---|---|---|---|
| OpenTable | [live](https://qdrant.tech/blog/case-study-opentable/) | [preview](https://deploy-preview-2648--condescending-goldwasser-91acf0.netlify.app/blog/case-study-opentable/) | 4 |
| &AI | [live](https://qdrant.tech/blog/case-study-and-ai/) | [preview](https://deploy-preview-2648--condescending-goldwasser-91acf0.netlify.app/blog/case-study-and-ai/) | 8 |
| Dust | [live](https://qdrant.tech/blog/dust-and-qdrant/) | [preview](https://deploy-preview-2648--condescending-goldwasser-91acf0.netlify.app/blog/dust-and-qdrant/) | 4 |
| Sprinklr | [live](https://qdrant.tech/blog/case-study-sprinklr/) | [preview](https://deploy-preview-2648--condescending-goldwasser-91acf0.netlify.app/blog/case-study-sprinklr/) | 4 |

Worth looking at specifically:

- **Dust** opens with the biggest change. The quote about cutting cost by 2x used to be a JPEG with the words in its alt text, plus the same quote repeated as prose underneath. It is one card now, and the text is selectable.
- **OpenTable** has three different speakers, so each card names its own.
- **&AI** shows the tiering: 8 cards, but only 2 carry the logo. The face and the logo are what make a card read as a headline, so they only work if most cards go without.
- **Sprinklr** is worth scrolling to the "Why Qdrant?" list. Three quotes are still inline in those bullets, deliberately - lifting them out would restructure the list, which is a layout call rather than a mechanical one.

Bayer is not in this PR. It picked up logos on all twelve cards at one point and has been reverted to exactly what is on master.

KShivendu:
@daniel-azoulai embedded Youtube Case study video with Dust directly in the case study page. It has a great thumbnail too. We should do more of them! 

embed

daniel-azoulai (approved):

PR #2657, opened 2026-08-21
Fastembed supported models

abdonpijpelink:
[Preview](https://deploy-preview-2657--condescending-goldwasser-91acf0.netlify.app/documentation/fastembed/fastembed-models/)

Adds a page that lists all models that are supported by FastEmbed. 

The six tables can be generated with the included script in `automation/fastembed-tables/`. We can set up workflow automation to keep the tables updated using this script, but let's do that in a separate PR.


qdrant-landing/content/documentation/fastembed/fastembed-models.md
```diff
@@ -0,0 +1,63 @@
...
+model_types = {
+    "Dense text": TextEmbedding,
+    "Sparse text": SparseTextEmbedding,
+    "Late interaction": LateInteractionTextEmbedding,
+    "Late interaction multimodal": LateInteractionMultimodalEmbedding,
+    "Image": ImageEmbedding,
+    "Reranking": TextCrossEncoder,
+}
+
+for label, cls in model_types.items():
```
joein: `cls` is kinda a builtin in python (used instead of `self` within classmethods)

joein:
My concerns here
1) The main one is that if we add these tables, the process of generating them should be completely automated, otherwise who's gonna maintain them (I doubt this should be someone's responsibility to manually generate them with each fastembed release)?
2) I am not sure whether it is a good idea to install and run fastembed everytime in the CI, it might be slow, maybe we could come up with another solution (e.g. have a file with the supported models auto-generated in fastembed and then we could just fetch it from there)

Since of the pt.1 , I think we should merge it only when we decide on the automation and how to implement it.

qdrant-landing/content/documentation/fastembed/fastembed-models.md
abdonpijpelink: Renamed it into `model_class`

PR #2663, opened 2026-08-24
Article: Your Search Defaults Are Wrong (draft)

jkupchanko:
## **[PREVIEW](https://deploy-preview-2663--condescending-goldwasser-91acf0.netlify.app/articles/keyword-search-in-qdrant/)**

Draft article on configuring keyword search in Qdrant, plus three figures.

Marked `draft: true`, so it will not publish. Opening it here so it is reviewable and tracked in the normal place rather than as a shared asset.

**What it covers.** Four decisions a developer actually faces: tuning BM25, exact terms and phrases, non-Latin and accented text, and whether to add a sparse vector at all. Each one ends with when to use it, when to skip it, and how to check on your own data.

**Measurements behind it** live in the private `qdrant-keyword-guide` repo: 138 runs across five BEIR corpora, three sparse methods and two dense models, with 76 paired significance tests.

**Still outstanding**

- Hero and preview images are placeholders. @kanungle generates the real ones, so please treat the current set as stand-ins.
- The title's last word is meant to carry the accent colour. Hugo escapes markup in frontmatter titles, so that needs `| safeHTML` on the article title in the shared template. Happy to drop the idea if it is not wanted.
- A few measured figures use `.text-success` and `.text-danger` in a table, and the guidance labels are bold. That is existing Bootstrap rather than new CSS, but inline classes in article markdown is a precedent worth a second opinion.
- Timing: I understand publication is waiting on v1.20. Worth noting the fuzzy matching tracking issue is still open with the public API work unstarted and both implementation PRs unmerged, so that may be worth confirming.

Review comments welcome on anything, including cutting sections. It has already been through several rounds and is shorter for it.


kanungle:
thanks. will review. put the PREVIEW link in first comment please. i did it this time around

kanungle:
This needs a bit of work / expansion on sparse embeddings, but adding @mrscoopers and @KShivendu to take a look for now

jkupchanko:
**Preview:** https://deploy-preview-2663--condescending-goldwasser-91acf0.netlify.app/articles/keyword-search-in-qdrant/

Pushed an update. Still a draft.

Expanded the sparse embeddings part you asked for: what BM25, SPLADE, and miniCOIL each actually do, and a table comparing them across all five corpora.

Fixed three wrong claims, cut 210 words, resized the cover art to match the other articles.

Need two things from you: the hero images, and whether v1.20 is still the gate.

kanungle:
Let's discuss in our 1:1 tomorrow. Overall this looks like an explainer for how full text search and keyword search works, which is for documentation. What I'm looking for is a convincing analysis for why sparse neural retrievers (maybe combined with fts) beats or compares well to traditional keyword search

jkupchanko:
**Preview:** https://deploy-preview-2663--condescending-goldwasser-91acf0.netlify.app/articles/keyword-search-in-qdrant/

Reworked the whole thing from scratch.

New structure, eight new diagrams.

Please review when you get time.

kanungle:
positioning is off (so are diagram styles). see my slack messages

jkupchanko:
**PREVIEW:** https://deploy-preview-2663--condescending-goldwasser-91acf0.netlify.app/articles/keyword-search-in-qdrant/

Went through your feedback and rebuilt it around developer decisions, and around what Qdrant actually gives you for each one.

**Focus:** what to use and when

**Configuration:** what to set in Qdrant

**Measurement:** what to validate yourself

**Benchmarks:** support the guidance, they don't lead it

**Every section:** problem → choice → configuration → takeaway

KShivendu:
I hold strong opinions so feel free to disagree.

I don't like claude's language. It somehow makes my brain hurt to read. You may ask OpenAI models to suggest improvements - explicitly ask it to make it more natural for humans to read. And do further changes from your side. You may use examples to explain concepts better. 

For example, 
> A customer types a part number, and a dense-only system returns things that mean roughly that: related, reasonable, not the part. Qdrant matches the words themselves on two surfaces beside your dense vector, a payload text index that filters and a sparse vector that ranks, and you configure each one separately. Here is what to switch on for four common keyword workloads, what our measurements settled, and what only your own data can settle.

This could look like this:
> A customer types "SKF 6205-2RS bearing", but a dense retrieval only system can search similar or related parts but miss the exact product the customer is looking for. Qdrant gives you two ways to match the actual words alongside your dense vector: a payload text index for filtering and a sparse vector for ranking. You can configure each one independently. In this article, we'll discuss how to handle four common keyword-search workloads, what our measurements found, and what you’ll should to configure based on your own data to make the best out of Qdrant.

This is just example. But I think the whole blog needs to keep this in mind. Same applies for section titles.

kanungle (changes requested):
Thanks for the hard work. I think it's more holistic now. However, i think all of this can be learned from ChatGPT directly though. It doesn't surface anything new/insightful, unfortunately

I also agree with @KShivendu that claude-ish is a terrible language

jkupchanko:
PREVIEW: https://deploy-preview-2663--condescending-goldwasser-91acf0.netlify.app/articles/keyword-search-in-qdrant/

@kanungle @KShivendu

Rewrote the whole thing in my own words. Same findings and same guidance, nothing dropped, just written by me this time.

Fixed the issues too: numbers, code, sources, diagrams.

kanungle:
As discussed last week, let's sync with Jenny and Dylan and change the approach / outcomes of this article goals

mrscoopers:
@kanungle should I read? You told me to wait until John polishes it, to not review something that might be gone:)

kanungle:
@mrscoopers no this is on hold right now. I would like you, John, and Dylan to discuss a full, compelling article on FTS and sparse embeddings, especially considering the FTS features on the roadmap. Timing-wise, I'd be totally cool with the final article going out in end of October

PR #2674, opened 2026-08-27
Apply the quote component to Deutsche Telekom, Voiceflow and Pathwork

KShivendu:
Follows #2648. Three more case studies get the quote component — nine cards, all verbatim.

| Case study | Cards | Marked out |
|---|---|---|
| Deutsche Telekom | 4 | "I looked at Qdrant and immediately loved the simplicity, Rust-based efficiency, and memory management capabilities. These guys knew what they were doing." |
| Voiceflow | 2 | "We started with Pinecone but eventually switched to Qdrant." |
| Pathwork | 3 | "We landed on Qdrant after extensive trial and error… Qdrant's support felt like having an additional principal engineer on our team." |

Also lifted: Deutsche Telekom's comparison against the alternative stack — *"Qdrant has way fewer components, compared to the other solution that required Kafka, Zookeeper… If you rescale it, you get downtime. Qdrant stays up."*

## Judgement calls worth a look

**Deutsche Telekom makes the same point three times.** "This is a distributed systems problem, not an AI problem" appears in three separate quotes in near-identical words. Only the first is a card; the other two stay as prose, so the article doesn't shout the same line three times.

**One quote is anonymous in the original** — "One engineer summarized their findings". The card attributes it to `An engineer on the team` rather than inventing a name. It reads a little oddly in the name slot, so say if you'd rather it stayed inline.

**Voiceflow only gets two cards.** Its best remaining quotes — "The Qdrant team is very responsive, ships features quickly and is a great partner to build with", and the SOC2 / Private Link one — all sit inside bullet lists. Lifting those restructures the list, which is a layout decision rather than a mechanical one. Sprinklr has three in the same position. That's seven quotes waiting on one decision, and it will keep recurring.

## Quotes are reproduced exactly

Where the original splits a quotation across an attribution tag — `"A," said X. "B."` — both halves go in the same card, since that is one statement interrupted by the tag. The tag goes and the comma that introduced it becomes a period. No words change, and each half is checked against the original either side of the join.

### DIFF
--- a/qdrant-landing/content/blog/case-study-deutsche-telekom.md
+++ b/qdrant-landing/content/blog/case-study-deutsche-telekom.md
@@ -23,7 +23,12 @@ partition: case-studies
 
 To achieve this, Telekom developed [*Frag Magenta OneBOT*](https://www.telekom.de/hilfe/frag-magenta?samChecked=true) *(Eng: Ask Magenta)*, a platform that includes chatbots and voice bots, built as a Platform as a Service (PaaS) to ensure scalability across Deutsche Telekom's ten European subsidiaries.
 
-"We knew from the start that we couldn't just deploy RAG, tool calling, and workflows at scale without a platform-first approach," Arun explains. "When I looked at the challenge, it looked a lot like a distributed systems and engineering challenge, not just an AI problem."
+{{< quote
+  text="We knew from the start that we couldn't just deploy RAG, tool calling, and workflows at scale without a platform-first approach. When I looked at the challenge, it looked a lot like a distributed systems and engineering challenge, not just an AI problem."
+  name="Arun Joseph"
+  role="Engineering and Architecture, AI Competence Center"
+  company="Deutsche Telekom"
+  featured="true" >}}
 
 ### Key Requirements for Scaling Enterprise AI Agents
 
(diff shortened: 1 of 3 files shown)


KShivendu:

| Case study | Before | After | Cards |
|---|---|---|---|
| Deutsche Telekom | [live](https://qdrant.tech/blog/case-study-deutsche-telekom/) | [preview](https://deploy-preview-2674--condescending-goldwasser-91acf0.netlify.app/blog/case-study-deutsche-telekom/) | 4 |
| Voiceflow | [live](https://qdrant.tech/blog/case-study-voiceflow/) | [preview](https://deploy-preview-2674--condescending-goldwasser-91acf0.netlify.app/blog/case-study-voiceflow/) | 2 |
| Pathwork | [live](https://qdrant.tech/blog/case-study-pathwork/) | [preview](https://deploy-preview-2674--condescending-goldwasser-91acf0.netlify.app/blog/case-study-pathwork/) | 3 |

Worth looking at specifically:

- **Deutsche Telekom** is the clearest before/after. Seven quotes were sitting in prose; four are cards now and three stay put, because the article makes the same "this is a distributed systems problem" point three times over.
- **Voiceflow** shows the in-bullet problem. Compare the two pages and you can see the strongest remaining quotes — the one about the team shipping features quickly, and the SOC2 / Private Link one — still buried in the bullet list underneath.
- **Pathwork** is worth reading for the quote itself rather than the layout.

PR #2678, opened 2026-08-27
v1.19.1 tokenizes unindexed text

coszio:
After merging https://github.com/qdrant/qdrant/pull/10341, Qdrant 1.19.1 will not use exact substring match for better parity with indexed filter.


timvisee (approved):

szabosteve (approved):
LGTM!

PR #2679, opened 2026-08-27
feat: qdrant internals article on dense vector storage

AstraBert:
### [PREVIEW](https://deploy-preview-2679--condescending-goldwasser-91acf0.netlify.app/articles/internals-of-dense-vector-storage/)

An article that dives deep into Qdrant dense vector storage, highlightinh the three storage variants, kernel concepts behind them, and the lifecycle of a vector (how we write it, read it and keep track of deletions).


qdrant-landing/content/articles/internals-of-dense-vector-storage.md
```diff
@@ -0,0 +1,194 @@
...
+category: qdrant-internals
+weight: 1
+---
+
+This is the first article in a series on Qdrant's storage internals, and it
+covers one question: once a collection has a fixed vector dimension and
+distance metric, how does a single named vector get stored, grown, and
+read back, in RAM, on disk, and through the kernel? 
+
+<!-- TODO: links-->
```
AstraBert, line 26: This marks articles of this series that are not yet ready/does not exist yet

AstraBert: Here and in other places

qdrant-landing/content/articles/internals-of-dense-vector-storage.md
```diff
@@ -0,0 +1,194 @@
+---
+title: "Internals of Dense Vector Storage"
+short_description: "How a single named vector gets stored, grown, and read back in RAM, on disk, and through the kernel, once a collection's dimension and distance metric are fixed."
+description: "Why Qdrant's dense vector storage needs no index, and how the volatile, immutable mmap, and appendable mmap storages differ under the hood."
+preview_dir: /articles_data/internals-of-dense-vector-storage/preview
+social_preview_image: /articles_data/internals-of-dense-vector-storage/preview/social_preview.jpg
```
AstraBert, line 6: TODO: preview images

generall (changes requested):
This article describes internals, which we don't actually want to be described. For instance VectorStorageEnum and stuff like volatile vector storage are pure internal concepts. volatile vector storage is not even used anywhere except for tests.

Those concept are not exposed to users and can be changed any day without warning as we don't provide any guarantees about those structures.

Overall, I don't think the topic is interesting enough to write the whole article about it.

Could you please share your motivation for this type of content? If it is more about optimization tricks we use to speed things up, like allocating pages or something like that, I would prefer to have a general article about performance optimizations we make, not a single component. And this article should be written in collaboration with core engineers. See https://qdrant.tech/articles/immutable-data-structures/

AstraBert:
@generall thanks for the review! The main motivation behind the article was to explore and explain internals of our storage layer to appeal to the technical community, especially to Rust devs, which is something that I think can benefit us in terms of both visibility (given all the traction rust has been gaining lately) and userbase (rust devs are relatively new to the AI world and many people are now starting to develop AI-powered software in rust). 

I now see, given what you said, why this article might not be the best way of connecting to the technical, rust-leaning community: what would you say could be good topics related to optimizations and other tricks you apply in qdrant I could write about? And who would be the people from the core team to talk to so that we can collaborate? thanks!

PR #2695, opened 2026-09-01
feat: consesus failure modes diagnostics

AstraBert:
### [PREVIEW](https://deploy-preview-2695--condescending-goldwasser-91acf0.netlify.app/documentation/tutorials-operations/consensus-failure-modes/)

A tutorial that guides the user through diagnosing consensus failures in distributed Qdrant deployments:

- What happens if a follower node drops
- What happens when there is a majority and a minority split due to a network partition
- What happens when partitions leave no healthy majority
- How to clean up a dead peer that is producing consensus error noise


generall:
> Recovering the Unrecoverable

Not sure about this part. It feels more like a hack then a production solution.

In qdrant operator we have a consensus `--reinit` option, maybe better talk about it? (It might require an assumption that cluster runs under hybrid/private cloud deployment mode rather than self-hosted)

qdrant-landing/content/documentation/tutorials-operations/consensus-failure-modes.md
```diff
@@ -0,0 +1,419 @@
...
+check_node_status_extra () {
+  for i in 0 1 2 3 4; do
+    port=$((6340 + i))
+    kubectl port-forward qdrant-consensus-demo-$i $port:6333 > /dev/null 2>&1 &
+    pf_pid=$!
+    sleep 1
+    echo "=== node $i (local port $port) ==="
+    curl -s http://localhost:$port/cluster | jq '.result.raft_info, .result.consensus_thread_status, .result.message_send_failures'
+    kill $pf_pid 2>/dev/null
+  done
```
generall, line 169: Snippet doesn't fit the length. 

Also, can it be avoided? Bash loops are unnecessary complex.

qdrant-landing/content/documentation/tutorials-operations/consensus-failure-modes.md
```diff
@@ -0,0 +1,419 @@
...
+The rest of the cluster still has quorum, 4 out of the original 5, so it keeps working, but the dead peer stays in the Raft peer list until you remove it explicitly.
+
+Look up its peer ID, then delete it hitting the `DELETE /cluster/peer/:id` endpoint:
+
+```shell
+kubectl port-forward qdrant-consensus-demo-0 6333:6333 > /dev/null 2>&1 &
+sleep 1
+
+peer_id=$(curl -s http://localhost:6333/cluster \
+  | jq -r '.result.peers | to_entries[] | select(.value.uri | contains("qdrant-consensus-demo-4.")) | .key')
```
generall: this looks machine-made ;) I would never be able to write (and even read) whis selector. 

I it might be easier to just point to the peer_id in the raw output

AstraBert:
I re-scoped the tutorial to explictly state that this is aimed at showing consensus failure modes in action, which is not applicable to hybrid cloud. I'd just use this is as reference material for people to understand what can happen in real scenarios and diagnose the failures faster

PR #2697, opened 2026-09-01
Apply the quote component to Bazaarvoice, Alhena, Anima Health and GoPerfect

KShivendu:
Four case studies, 27 quotes. All four were already written with pull quotes, so the set of quotes does not change — only how they are attributed.

Today the speaker's name sits *inside* the dark box as another line of body text, directly under the quote. The component gives it its own row, and distinguishes speakers where an article has more than one.

| Case study | Cards | Speakers | Marked |
|---|---|---|---|
| Bazaarvoice | 8 | Dr. Lou Kratz, Abhijeet Dhupia | 2 |
| Alhena AI | 7 | Kshitiz Parashar, Kang-Chi Ho | 2 |
| Anima Health | 7 | Colin Cooke | 2 |
| GoPerfect | 5 | Idan Shaked, Eylon Etshtein | 0 |

The logos go on the strongest claims only — Alhena replacing FAISS, Pinecone and Weaviate, and cutting retrieval time by 90 percent; Anima on patient privacy and on Qdrant never being something they had to think about. Marking every card would mean marking none.

GoPerfect gets none: there is no GoPerfect logo anywhere in the repo, only article images. Its five cards carry name and role, which the component handles.

## A note on Alhena's attribution

Five of Alhena's seven quotes are signed with just a name — `— Kshitiz Parashar`, no title. Both speakers' full titles appear on the other two quotes in the same article, so the roles are filled in from there. Nothing is invented.

## Checked

Quote text and attribution were parsed out of each file rather than retyped, so no word could drift. Then verified: **27 quotes, 0 altered, 0 words added to any prose, 0 links lost.** Clean build with the dev server stopped — every card renders, no unparsed shortcodes, no card rendering inside a paragraph, and every `logo=` path resolves to a real file.

That last check caught a real mistake: I had guessed a GoPerfect logo path that does not exist, which would have shipped a broken image.

## Not changed

No quote was moved. Each already sat after the prose written for it. None of the four articles opens with an introductory paragraph, so there is no lead card — a quote above a section heading has nothing to land against.


KShivendu:

| Case study | Before | After | Cards | Marked |
|---|---|---|---|---|
| Bazaarvoice | [live](https://qdrant.tech/blog/case-study-bazaarvoice/) | [preview](https://deploy-preview-2697--condescending-goldwasser-91acf0.netlify.app/blog/case-study-bazaarvoice/) | 8 | 2 |
| Alhena AI | [live](https://qdrant.tech/blog/case-study-alhena/) | [preview](https://deploy-preview-2697--condescending-goldwasser-91acf0.netlify.app/blog/case-study-alhena/) | 7 | 2 |
| Anima Health | [live](https://qdrant.tech/blog/case-study-anima-health/) | [preview](https://deploy-preview-2697--condescending-goldwasser-91acf0.netlify.app/blog/case-study-anima-health/) | 7 | 2 |
| GoPerfect | [live](https://qdrant.tech/blog/case-study-go-perfect/) | [preview](https://deploy-preview-2697--condescending-goldwasser-91acf0.netlify.app/blog/case-study-go-perfect/) | 5 | 0 |

What to look at:

- **The attribution.** On every live page the speaker's name is another line of text inside the dark box, directly under the quote. In the preview it has its own row, so the quote and the person saying it are visually separate.
- **Multiple speakers.** Three of the four articles have more than one, which is easy to miss today. Bazaarvoice has Abhijeet Dhupia on the two migration quotes and Dr. Lou Kratz on the other six; Alhena alternates between Kshitiz Parashar and Kang-Chi Ho; GoPerfect has Idan Shaked and Eylon Etshtein.
- **Scroll a whole page rather than one card.** Only two of eight on Bazaarvoice carry the logo, two of seven on Alhena and Anima. The contrast is the point, and it only works because most cards go without.
- **GoPerfect has no marked card at all**, because there is no GoPerfect logo in the repo. Worth seeing whether that reads as an omission or as fine.

All four preview pages verified to contain the cards, not merely to return 200.

PR #2700, opened 2026-09-01
Tag case studies with the customer's industry, deployment and region

KShivendu:
Stacked on #2699 — review that one first.

The customers catalog already records each customer's industry, product, region and use cases, and the filters on `/customers/` run off those exact values. A case study page showed none of it.

Adds a `case-study-tags` shortcode that **reads the catalog by `blog_path`** rather than taking the values as parameters:

```
{{< case-study-tags >}}                                        industry · deployment · region
{{< case-study-tags fields="industry product location use_cases" >}}
{{< case-study-tags id="dust" >}}                              for a page not in the catalog
```

Retyping the values per article would let the page and the filters drift apart. This way adding a tag means adding it to the catalog, where it also becomes filterable.

## Why three fields, not four

`use_cases` is the richest field — RAG appears on 31 customers — but most customers carry three of them, so including it produced six chips wrapping onto two rows and buried the industry and deployment behind a list of capabilities. It stays available as an opt-in.

## Scope

Applied to the **46 case studies whose published URL matches a catalog row**. Six have no row — Bayer, Data Graphs, GoPerfect, Minima, Sapu, Sunny Health — and are left alone rather than carrying a shortcode that renders nothing. Adding rows for them is worth doing separately; they would then also appear in the `/customers/` catalog, which they currently do not.

Placement follows whatever each article opens with: after the hero image, after the title heading, or at the top when it opens straight into prose.

## One data fix included

`locationse:` on the ConvoSearch row. It was the only page rendering two tags instead of three, and the same typo had been quietly keeping ConvoSearch out of the Geography filter on `/customers/`.

## Data problems found but deliberately not fixed here

Putting catalog values on public pages makes catalog errors visible. Two worth a separate pass, because both need someone to decide the right value rather than correct a typo:

- **Pariti** is recorded as `industry: Technology, location: Global`, while its own article opens *"Pariti is a referral-driven hiring marketplace in Africa."*
- One row has `company_size: 1-50 by`.

## Checked

Clean build: 46 pages render tags, every one showing exactly three, no empty tag lists, no unparsed shortcodes.

### DIFF
--- a/qdrant-landing/content/blog/case-study-alhena.md
+++ b/qdrant-landing/content/blog/case-study-alhena.md
@@ -23,6 +23,8 @@ partition: case-studies
 
 # How Alhena AI unified its AI stack and accelerated ecommerce outcomes with Qdrant
 
+{{< case-study-tags >}}
+
 ![How Alhena AI unified its AI stack and improved ecommerce conversions with Qdrant](/blog/case-study-alhena/alhena-bento-box-dark.jpg)
 
 ## Building AI agents that drive both revenue and support outcomes
--- a/qdrant-landing/content/blog/case-study-and-ai.md
+++ b/qdrant-landing/content/blog/case-study-and-ai.md
@@ -20,6 +20,8 @@ partition: case-studies
 
 ![Bento Box](/blog/case-study-and-ai/and-ai-bento.jpg)
 
+{{< case-study-tags >}}
+
 ## How \&AI scaled global patent retrieval with Qdrant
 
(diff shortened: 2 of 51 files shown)


KShivendu:
Superseded by #2707, which takes a better approach.

This version added a shortcode to 51 content files and stacked on #2699. #2707 does it as a layout partial instead: no article is edited, it picks up case studies with custom slugs automatically, and it sits on master rather than stacking.

It also renders the tags inside the byline row every blog post already has, rather than giving case studies a header row other posts do not have.

PR #2711, opened 2026-09-02
Rebuild Cloud Inference landing page

meinsta:
## Read this first: `/cloud-inference/` already existed

MKT-425 is titled "net new LP", but the slug is live today and is the target of the
`Qdrant Cloud Inference` entry in the main nav (`content/headless/menu.md`). It had its
own layout and seven content files.

I rebuilt it in place rather than creating a second slug, so the URL, the nav entry and
any inbound links keep working. Nothing is orphaned. If you wanted the new page at a
different path with the old one kept, stop here and say so, because this PR replaces the
old sections: `cloud-inference-hero`, `-features`, `-cluster`, `-faq`, `-read-doc` and
`-get-started`, their four partials, and five SCSS files. The old hero and cluster images
in `static/img/` are left alone in case anything else references them.

## What this is

`/cloud-inference/` rebuilt to the signed-off desktop frame (`Inference LP`), following
the Security page structure.

Sections, top to bottom: hero, Why It Matters, the model-request callout, What You Get,
How It Works, Cloud Inference Approaches, FAQs, footer CTA.

- Jira: [MKT-425](https://qdrant.atlassian.net/browse/MKT-425)
- Frame from the design review: https://www.figma.com/design/ubPe8s8pViytom0Mapcno8/Qdrant-Web-Design?node-id=20638-114912
- Frame linked on the Notion card: https://www.figma.com/design/ubPe8s8pViytom0Mapcno8/Qdrant-Web-Design?node-id=10947-44381
- Copy: the Google Doc linked in the MKT-425 description

## Reused vs. new

Reused as-is: `pricing-banner` for the "Need a Model We Don't Host Yet?" callout,
`cta-banner`, `accordion`, `features-table` (same call shape as `security/access`),
`picture`, the `card` component and the Bootstrap grid. No shared partial was modified.

New: six section partials and their SCSS, plus four icons exported from the frame
(`shield-check-blue`, `layers-blue`, `puzzle-blue`, `square-pen-blue`) and two
illustrations exported at 2x as PNG with WebP siblings.

Some of this structure mirrors the GPU Indexing PR. The two are deliberately independent
branches per the brief, so neither depends on the other; worth a follow-up to factor out
the shared section shell once both land.

## Fact check

Cloud Inference runs at both upsert time and query time. Nothing on the page implies
query-only: step 02 says "storing the vector when you write and matching against it when
you search", and the first card says "every upsert and query stays on one path".

## Things worth a look

1. **I fixed four typos in the copy doc.** Each is mechanical and easy to revert:
   - table row "Qdrant-hosted eEmbedding models" to "Qdrant-hosted embedding models"
   - step 02 "plus the model to use.." to "plus the model to use."
   - FAQ 2 "separate from Cloud Inference To add a model" to "...Cloud Inference. To add a
     model"
   - collapsed some double spaces
2. **Two bare URLs became links.** FAQ 3 ended with a raw
   `https://qdrant.tech/documentation/inference/.` and FAQ 5 with a bare
   `qdrant.tech/contact-us/`.
3. **`_index.md` is untouched.** Its title and meta description still describe the old
   positioning and no longer match the new H1. Tell me if you want them updated.
4. **One shared file changed: `layouts/partials/js.html`.** `features-table.js` was
   loaded only on the security section, but the comparison table needs it: every value
   cell is `display: none` until that script adds `--mobile-active`. On mobile the table
   rendered its column tabs above three empty rows. The include now covers
   `cloud-inference` as well. Any other page adopting `features-table` will hit the same
   thing.
5. **Table header icons are my pick** from the theme's existing cloud icons, matching the
   frame's colours: `cloud-managed`, `cloud-hybrid-blue`, `cloud-private-teal`.

Checked locally with `hugo serve` against both the desktop and the mobile frame at
375px.


trean (changes requested):
@meinsta There are some inconsistencies with the design of this page in Figma (strong text is for merge blockers):
**1. There is no space between the hero description and the buttons**
image

2. [nit] "Pick the Right Embedding Model, Or Bring Your Own" is three lines in the Figma design
image

3. "Follow the embedding model migration guide" is in thin red text
image

4. [nit] The title should be two lines
image

5. These cards are missing some design elements between them 
image

6. The first icon in the table should be violet
image

7. FAQ section background is wrong. In the design, it is the same color as the above section
image

**8. On the mobile version, the hero illustration looks not like in the Figma**
image

9. In the mobile version, the table should use the `.features-table--mobile-tabs` variant instead of arrows.
image

meinsta:
Thanks @trean — both blockers and five of the seven nits are fixed. Pushed in 9dc0e3c and the two commits before it.

**Both blockers**

1. **Hero description / buttons spacing.** Root cause was a specificity bug, not a missing margin. `cloud-inference-section-header` was being `@include`d from *inside* `.cloud-inference-hero`, so every selector it generated compiled to `.cloud-inference-hero .cloud-inference-hero__description` and outranked the overrides written below the include. The 24px bottom margin was there in the source the whole time and was being silently dropped — as was the hero's `$h2` title size. Both mixins now emit at `@at-root`, so source order decides, and every section that includes them gets its intended overrides back.

8. **Mobile hero illustration.** The screenshot was `width: 125%; max-width: none` at every breakpoint, so on mobile it blew up and got cropped mid-panel. The bleed is now desktop-only, matching how `qdrant-for-startups-hero` handles the same treatment. Below `lg` the console sits inside the container and you can see all of it.

**Nits**

- 2 & 4 — "Pick the Right Embedding Model, Or Bring Your Own" wraps to three lines, "From Raw Content to Search Results in Three Steps" to two.
- 3 — the migration guide link dropped `link_dark`, so it renders in the red accent.
- 6 — added `cloud-managed-violet.svg`, recoloured from `cloud-managed.svg` to `#8547FF`, matching how the blue and teal column icons are derived.
- 7 — FAQ moved to `$neutral-94`, same as Approaches above it.
- 9 — the Approaches table dropped the `mobile-arrows` modifier. There's no `features-table--mobile-tabs` class in the theme; tabs are what you get with no modifier at all, which is what it uses now.

**Still open**

- 5 — the three numbered cards. I can see the divider inside each card is rendering, so I think you mean something *between* the cards, but I can't tell what from the screenshot. Connector arrows between steps? If you can point me at the frame or name the element I'll add it.

Also added `rel="noopener noreferrer"` to the hero's external links while I was in there — Copilot flagged the same gap on the GPU Indexing hero this one was copied from.

trean (approved):
Thank you so much, @meinsta! 
I can still see a few minor issues to address, but none of them block publishing this page. Please merge when you're ready.
**Please ask the agent to rebase the branch onto master, resolving conflicts before merging**. For this type of conflict we have here, it should have no problems doing this.

## Notes
9. yes, you're right, my comment about the 'tabs' modifier was misleading. Thanks for pointing that out!

## Would be good to fix as follow up
2. The mobile hero illustration looks closer to the Figma design now, but it should use another image according to Figma:

| **Preview (cut image)** | **Figma** |
|---------|-------|
| image | image |

_\* ignore the different sizes of screenshots, please_

5. I was talking about these connectors between cards (see screenshot), you can see them in Figma

image

meinsta:
Rebased onto `master` and pushed — the conflict was `js.html`, where master had moved the `features-table.js` block further down and widened its slice to `security resilience quantization observability`. Resolved by dropping my stale copy and adding `cloud-inference` to master's list. Verified in a local build that `/cloud-inference/` loads `features-table.min.js`, same as `/security/`.

Both follow-ups are done too, in 3086315.

**5 — step connectors.** Pulled them from the frame. Each is a 64px line fading from 10% to full in the accent colour of the step it points at (blue into 02, teal into 03), capped with a ring at each card edge, aligned on the number pill row.

One thing worth flagging: the row gutter at `lg` was 16px, and the connector is 74px wide, so there was nowhere to put it. The frame runs a 74px gutter with 318px cards, so the gutter now matches — which means the cards are ~38px narrower than they were on the preview you reviewed. That's closer to the frame, but it is a visible change, so please sanity-check it.

The rings carry three stacked drop shadows in the frame at 9%, 5% and 1% opacity. On a 9px circle they're invisible and they made the SVG six times larger, so the shipped asset omits them.

**2 — mobile hero.** You were right that it's a different image, not just a different scale — the mobile frame's console is a wider crop with the full right-hand panel visible, where the shipped asset is a tall 1274x998 crop. Exported the mobile frame at 4x and added `console-inference-mobile.png`/`.webp`. `picture.html` now takes an optional `mobileSrc` and serves it below `lg`; callers that don't pass one render exactly as before, so no other page changes.

PR #2715, opened 2026-09-03
CBI-509 CBI-510: document that API keys survive user removal

pedro-tramontin:
Neither Cloud Management Keys (CBI-509) nor Database API keys (CBI-510) are deleted or revoked when the user who created them is removed from the account. They keep working with whatever permissions they were granted, so they can still be used to manage the account or access its clusters after their creator loses access.

Adds a note about this to the Cloud API authentication section, the Database Authentication page, and the User Management "Removing a User" flow, pointing at revoking keys separately when offboarding a user.


abdonpijpelink (approved):
LGTM

PR #2717, opened 2026-09-03
Note that max_segment_size scales with CPU count

Dylancouzon:
Scope reduced, see comments
Deploy preview: [Merge Optimizer](https://deploy-preview-2717--condescending-goldwasser-91acf0.netlify.app/documentation/ops-optimization/optimizer/#merge-optimizer)

A user asked on Discord how to predict the `max_segment_size_kb` default. The docs said only that it is "automatically selected considering the number of available CPUs".

Adds two sentences to the Merge Optimizer section: the default scales with CPU count and differs between machines, and anyone who needs a predictable segment size should set the parameter explicitly. Per @generall, the calculation itself stays out of the public contract.

🤖 Generated with [Claude Code](https://claude.com/claude-code) - reviewed by hand

### DIFF
--- a/qdrant-landing/content/documentation/ops-optimization/optimizer.md
+++ b/qdrant-landing/content/documentation/ops-optimization/optimizer.md
@@ -51,7 +51,7 @@ Qdrant requires at least one small segment to handle frequently updated data eff
 
 The target number of segments is specified by the `default_segment_number` parameter, which typically defaults to the number of CPUs. During optimization, the optimizer may merge the three smallest segments into one, aiming to balance segment size and system performance.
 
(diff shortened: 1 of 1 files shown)


generall:
this is a kind of thing, we don't actually want to document. I don't think it should be part of public contract.

Dylancouzon:
@generall Noted. Should we at least document the behavior without getting into the formula? Basically, the default scales with CPU count, and anyone who needs a predictable segment size should `set max_segment_size_kb` explicitly.

generall:
ok

abdonpijpelink (approved):

PR #2725, opened 2026-09-07
feat: slicing filter tutorial

AstraBert:
### [PREVIEW](https://deploy-preview-2725--condescending-goldwasser-91acf0.netlify.app/documentation/tutorials-search-engineering/slicing-filter/)

Add a tutorial for the recently introduced `slice` filter, which covers:

- How to use it for deterministic sampling
- How to use it for parallel scrolling
- How to use it for more complex sampling by combining with other filters
- The properties of the returned slices (disjoint, slice `n` of `m` is a subset of slice `n` of `p` where `p < m`)


AstraBert:
Examples PR: https://github.com/qdrant/examples/pull/127

qdrant-landing/content/documentation/tutorials-search-engineering/slicing-filter.md
```diff
@@ -0,0 +1,241 @@
...
+## Setup
+
+The [Cloud Quickstart](/documentation/cloud-quickstart/) covers creating a cluster and connecting a client. This tutorial uses [Qdrant Cloud Inference](/documentation/inference/cloud-inference/) to embed the sample data server-side, with the free model `sentence-transformers/all-MiniLM-L6-v2`.
+
+```python
+import os
+
+from qdrant_client import AsyncQdrantClient, models
+
+client = AsyncQdrantClient(
```
Dylancouzon, line 40: Is there a reason that we're using the Async client? 
When running the code `await client.upload_points` returned `TypeError: 'NoneType' object can't be awaited`
Also, every snippet on the page uses top-level `await`, so a reader copying them into a script gets a `SyntaxError`.

qdrant-landing/content/documentation/tutorials-search-engineering/slicing-filter.md
```diff
@@ -0,0 +1,241 @@
...
+
+A point matches slice `index` of `total` when `hash(id) % total == index`, using SipHash-2-4 over the point ID bytes. For a fixed `total`, slices `0` through `total - 1` are disjoint and together cover the whole collection. Unlike [random sampling](/documentation/search/search/#random-sampling), a given slice always returns the same points, and it composes with any other filter condition, including a payload filter.
+
+This tutorial covers three uses: parallel `scroll` across workers, reproducible sampling for evaluation, and stratified sampling by combining `slice` with a payload filter.
+
+## Setup
+
+The [Cloud Quickstart](/documentation/cloud-quickstart/) covers creating a cluster and connecting a client. This tutorial uses [Qdrant Cloud Inference](/documentation/inference/cloud-inference/) to embed the sample data server-side, with the free model `sentence-transformers/all-MiniLM-L6-v2`.
+
```
Dylancouzon, line 36: Add an install line with `pip install qdrant-client`

qdrant-landing/content/documentation/tutorials-search-engineering/slicing-filter.md
```diff
@@ -0,0 +1,241 @@
...
+```python
+import os
+
+from qdrant_client import AsyncQdrantClient, models
+
+client = AsyncQdrantClient(
+    url=os.environ["QDRANT_URL"],
+    api_key=os.environ["QDRANT_API_KEY"],
```
Dylancouzon:
```suggestion
    url="https://your-endpoint.cloud.qdrant.io:6333",
    api_key="<paste-your-key>",
```
Consistency with other tutorials

qdrant-landing/content/documentation/tutorials-search-engineering/slicing-filter.md
```diff
@@ -0,0 +1,241 @@
...
+aliases:
+  - /documentation/tutorials/slicing-filter/
+---
+
+# Deterministic Slicing with the Slice Filter
+
+| Time: 20 min | Level: Intermediate | Output: [GitHub](https://github.com/qdrant/examples/blob/master/slicing-filter/Slicing_Filter.ipynb) | [![Open In Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://githubtocolab.com/qdrant/examples/blob/master/slicing-filter/Slicing_Filter.ipynb) |
+| --- | ----------- | ----------- | ----------- |
+
+Exporting or re-embedding a large collection with a single `scroll` call is slow, and splitting the work across worker processes usually means paging through the collection and dividing IDs by hand, or reaching for random sampling that returns a different set of points on every run. Qdrant's `slice` filter condition, available as of v1.19.0, solves both: it divides a collection into a fixed number of deterministic, disjoint subsets, so several workers can each claim one subset with no coordination, and a single subset makes a reproducible sample for evaluation or a train/test split.
```
Dylancouzon: - Scroll is undefined
- Random sampling isn't an alternative to exporting a collection.

qdrant-landing/content/documentation/tutorials-search-engineering/slicing-filter.md
```diff
@@ -0,0 +1,241 @@
...
+
+A point matches slice `index` of `total` when `hash(id) % total == index`, using SipHash-2-4 over the point ID bytes. For a fixed `total`, slices `0` through `total - 1` are disjoint and together cover the whole collection. Unlike [random sampling](/documentation/search/search/#random-sampling), a given slice always returns the same points, and it composes with any other filter condition, including a payload filter.
```
Dylancouzon: l.17
My thoughts: 
I had to read through a few times to understand this. 
The hash formula is a lot of detail this early, `SipHash-2-4` is undefined, is this information needed for the reader at this stage? 

Suggestion:
I'd cut it to the plain version and link /documentation/search/filtering/#slice for the mechanics, something like: "Each point falls into exactly one slice, picked by a hash of its ID. Ask for slice 3 of 8 and you get the same eighth of the collection on every run, and on any Qdrant version." Keep the last sentence

qdrant-landing/content/documentation/tutorials-search-engineering/slicing-filter.md
```diff
@@ -0,0 +1,241 @@
...
+
+await client.upload_points(collection_name=collection_name, points=points, wait=True)
+```
+
+## Scrolling a Single Slice
+
+A `SliceCondition` takes an `index` and a `total`. This scrolls slice `3` of `8`, one eighth of the collection:
+
+```python
+result, _ = await client.scroll(
```
Dylancouzon: l.84
The page sells exporting or re-embedding a large collection, and every scroll here stops after one page.
we should name `next_page_offset` and probably show how to use it

qdrant-landing/content/documentation/tutorials-search-engineering/slicing-filter.md
```diff
@@ -0,0 +1,241 @@
...
+    print(f"worker {i}: {len(s)} points")
+print(f"sum across workers: {sum(len(s) for s in slices)}")
+# worker 0: 115 points
+# worker 1: 143 points
+# worker 2: 118 points
+# worker 3: 124 points
+# sum across workers: 500
+```
+
+Every point appears in exactly one slice, so the counts add up to the full collection with no overlap and no gaps:
```
Dylancouzon: The output above shows 115 points for one worker and 143 for another, a ~25% difference. Would using more slices than workers help balance the load better? If yes, maybe worth calling that out as a best practice. Just a thought, not sure if I’m right here.

qdrant-landing/content/documentation/tutorials-search-engineering/slicing-filter.md
```diff
@@ -0,0 +1,241 @@
...
+overlap = set.intersection(*ids_per_slice)
+union = set.union(*ids_per_slice)
+
+count_result = await client.count(collection_name=collection_name)
+
+print(f"overlapping IDs: {len(overlap)}")
+print(f"union covers full collection: {len(union) == count_result.count}")
+# overlapping IDs: 0
+# union covers full collection: True
+```
```
Dylancouzon: I'd cut this snippet. The block above already prints sum across workers: 500

qdrant-landing/content/documentation/tutorials-search-engineering/slicing-filter.md
```diff
@@ -0,0 +1,241 @@
...
+## Concurrent Scrolling Across Workers
+
+The main use case is splitting a full scroll into `N` independent, non-overlapping scrolls, one per worker. Each worker only needs its own `index`, and since the slices don't overlap, the requests have nothing to coordinate on and can run concurrently. Use `AsyncQdrantClient` with `asyncio.gather` to fire all requests at once instead of waiting on them one at a time.
+
+```python
+import asyncio
+
+
+async def scroll_slice(index: int, total: int) -> list[models.Record]:
+    records, _ = await client.scroll(
```
Dylancouzon: Even if we don't use it, I'd name `next_page_offset` everywhere, so we don't leave things undefined in a tutorial

qdrant-landing/content/documentation/tutorials-search-engineering/slicing-filter.md
```diff
@@ -0,0 +1,241 @@
...
+
+print(f"overlapping IDs: {len(overlap)}")
+print(f"union covers full collection: {len(union) == count_result.count}")
+# overlapping IDs: 0
+# union covers full collection: True
+```
+
+## Reproducible Sampling
+
+Because the hash is stable across runs and Qdrant versions, a single slice makes a reproducible sample. Requesting slice `0` of `total: 10` always returns the same 10% of the collection, which makes it a solid choice for a recall benchmark or a train/test split that has to be repeatable.
```
Dylancouzon: Could we explain what goes wrong without it? 
It's a lot of code and still we never show the payoff. This section draws the sample and proves the sample is the same sample.

qdrant-landing/content/documentation/tutorials-search-engineering/slicing-filter.md
```diff
@@ -0,0 +1,241 @@
...
+    with_payload=False,
+    with_vectors=False,
+)
+
+fine_ids = {p.id for p in fine}
+coarse_ids = {p.id for p in coarse}
+
+print(f"slice 0/4 is a subset of slice 0/2: {fine_ids.issubset(coarse_ids)}")
+# slice 0/4 is a subset of slice 0/2: True
+```
```
Dylancouzon, line 214: I'd cut this snippet. The sentence above already makes the point, and the code only prints True for something the filter guarantees. If you want to keep anything here, say what it buys the reader: you can go from total: 2 to total: 4 to halve your sample, and every point in the smaller sample was in the bigger one, so nothing you already evaluated drops out.

Dylancouzon (changes requested):
Overall, pretty good tutorial. Feel free to push back on any of my inline comments. I dug into `slice` and experimented with it a bit to make sure I understood the mechanics before reviewing this, but my knowledge may still have gaps so some of my comments may be worth validating.

Two things I'd add:
- `slice` works in `query_points` too I believe, I'd add something about that and it would make us do something with the vectors we ingest.
- I'd add a note that`slice` works on shards and needs no payload index.

On the Notebook, I had Claude run it and it only had 2 pieces of feedback:
- Read the URL and key from Colab `userdata` instead of `os.environ`. The page ships a Colab badge, and in a fresh Colab runtime that cell raises `KeyError: 'QDRANT_URL'`.
- The random sampling link goes to `.../search/search/index.md#random-sampling`, which serves the raw markdown and drops the anchor. Use the same URL the page uses.

Notebook will need to be brought back to parity once all feedback is applied.

Dylancouzon:
I also recommend running the `/neil-review` skill after making changes; it provides mostly good formatting and best‑practice feedback and usually speeds up the review process

qdrant-landing/content/documentation/tutorials-search-engineering/slicing-filter.md
AstraBert: yeah, the await on upload_points slipped through (that is one of the sync methods we have on the async client), but in this particular tutorial if we don't use async we can't showcase concurrent scrolling of points with asyncio.gather (a sync client can only do sequential work unless you get it to work with threading which is arguably more complicated than asyncio)

qdrant-landing/content/documentation/tutorials-search-engineering/slicing-filter.md
AstraBert: sound good, I'll add that 🫡

qdrant-landing/content/documentation/tutorials-search-engineering/slicing-filter.md
AstraBert: Yeah , sorry, sometimes the ruff/basedpyright that's in me gets the best of my coding 😃  Will adjust 👍

qdrant-landing/content/documentation/tutorials-search-engineering/slicing-filter.md
AstraBert: For how the tutorial is designed here you get a 1:1 ratio for slice:worker: arguably adding more slices (thus more workers) would smoothen out the differences and you would get a better distributed load, but I would probably need to test it. Adding more slices than workers here would mean that each workers fetches one or more slices internally, which I am not fully sure would help distribute the load better (I'd have to try, just reasoning theoretically)

AstraBert: ok, i briefly tested it an neither a bigger number of slices nor more slices per worker are a strong predictor of smoother load, but adding more workers seem to help a bit more

qdrant-landing/content/documentation/tutorials-search-engineering/slicing-filter.md
```diff
@@ -0,0 +1,262 @@
...
+
+Getting all the points of a large collection with a single call to the [Scroll API](/documentation/manage-data/points/#scroll-points) is slow, and splitting the work across worker processes usually means paging through the collection and dividing IDs by hand. 
+
+Qdrant's [`slice` filter condition](/documentation/search/filtering/#slice), available as of v1.19.0, addresses this pain point: it divides a collection into a fixed number of deterministic, disjoint subsets, so a fixed number of workers can each claim one subset with no coordination, and a single subset makes a reproducible sample for evaluation or a train/test split.
+
+Qdrant assigns each point to one slice by hashing its ID, and, for a fixed `total`, slices `0` through `total - 1` are disjoint and together cover the whole collection. Unlike [random sampling](/documentation/search/search/#random-sampling), a given slice always returns the same points, and it composes with any other filter condition, including a payload filter. Because `slice` matches on the point ID hash rather than payload data, it needs no payload index: like `has_id`, it is checked per candidate point within each shard that receives the query.
```
Dylancouzon:
```suggestion
Qdrant assigns every point to one slice by hashing its ID. For a fixed `total`, slices `0` through `total - 1` never overlap and together cover the entire collection. Unlike [random sampling](/documentation/search/search/#random-sampling), the same slice always returns the same set of points. You can also combine `slice` with any other filter condition, including payload filters. Because slicing is based on the point ID hash rather than payload data, it does not require a payload index. Like `has_id`, the slice condition is checked against each candidate point within every shard that receives the query.
```
Slightly Simplified the language, not all of our readers first language is English. Minor nit, feel free to reject.

qdrant-landing/content/documentation/tutorials-search-engineering/slicing-filter.md
```diff
@@ -0,0 +1,262 @@
...
+```bash
+pip install qdrant-client
+```
+
+While the installation completes, follow the [Cloud Quickstart](/documentation/cloud-quickstart/) to create a free cluster and retrieve the credentials to connect the client to it. This tutorial uses [Qdrant Cloud Inference](/documentation/inference/cloud-inference/) to embed the sample data server-side, with the free model `sentence-transformers/all-MiniLM-L6-v2`.
+
+```python
+import os
+
+from qdrant_client import AsyncQdrantClient, models
```
Dylancouzon:
```suggestion
from qdrant_client import AsyncQdrantClient, models
```
os not needed anymore

qdrant-landing/content/documentation/tutorials-search-engineering/slicing-filter.md
Dylancouzon: Works for me, top-level awaits still throw `SyntaxError` in a plain .py though

qdrant-landing/content/documentation/tutorials-search-engineering/slicing-filter.md
Dylancouzon: Thanks for testing it :)

qdrant-landing/content/documentation/tutorials-search-engineering/slicing-filter.md
```diff
@@ -0,0 +1,262 @@
...
+```python
+from siphash24 import siphash24
+
+def slice_of(point_id: int, total: int) -> int:
+    digest = siphash24(point_id.to_bytes(8, "little")).digest()
+    return int.from_bytes(digest, "little") % total
+
+print(f"all results in slice 0: {all(slice_of(p.id, 5) == 0 for p in results.points)}")
+# all results in slice 0: True
+```
```
Dylancouzon: Not sure I see the point of that part. Installing a library just to print True. 
We could just print what the search returned above

Dylancouzon (changes requested):
Important: The tutorial needs to be listed on `tutorials-lp-overview.md`

Should be good to go once those changes are in - Including the reopened comment above

qdrant-landing/content/documentation/tutorials-search-engineering/slicing-filter.md
AstraBert: yeah, i know, it really sucks that python  does not work with async as first class citizen like TS/JS. that's why I just decided to build a script runner that execute scripts top-level awaits lol :see_no_evil: https://github.com/AstraBert/tlawait

qdrant-landing/content/documentation/tutorials-search-engineering/slicing-filter.md
AstraBert: yeah, I was also a bit unsure about this but it was the only section that did not show any "assertion" (like the other ones all assert properties of the slice filter and show them), so I decided to include it here to show that the retrieved points are actually getting filtered (which might not be immediately intuitive otherwise since the hash of the point ID is pretty abstract and it's not like a payload field that you can just print out)

AstraBert:
> Important: The tutorial needs to be listed on `tutorials-lp-overview.md`
> 
> Should be good to go once those changes are in - Including the reopened comment above

@Dylancouzon mmmh my understanding is that Hugo already renders the headless search-engineering.md file (which I already modified in this PR):

```md
### Search Engineering Tutorials
*Master vector search modalities, reranking, and retrieval quality.*

{{% include "content/documentation/headless/content/tutorials/search-engineering.md" %}}
```

qdrant-landing/content/documentation/tutorials-search-engineering/slicing-filter.md
Dylancouzon: Unless I'm misunderstanding, could you do the same proof without the dependency? scroll slice 0 of 5, keep the IDs, then show the search results are a subset of that set

qdrant-landing/content/documentation/tutorials-search-engineering/slicing-filter.md
Dylancouzon: Ha, nice. For the page itself I would keep it to one sentence before `## Setup` pointing at the Colab notebook, with `asyncio.run(main())` as the script fallback. That way a reader who copies into a .py file knows what to do before hitting the first snippet.

Dylancouzon:
> mmmh my understanding is that Hugo already renders the headless search-engineering.md file (which I already modified in this PR):

You are correct, my bad

qdrant-landing/content/documentation/tutorials-search-engineering/slicing-filter.md
AstraBert: oh yeah, absolutely, it's actually a better way to show it :) I'll update

qdrant-landing/content/documentation/tutorials-search-engineering/slicing-filter.md
```diff
@@ -0,0 +1,260 @@
...
+# worker 8: 46 points
+# worker 9: 49 points
+# sum across workers: 500
+```
+
+Every point appears in exactly one slice, so the counts add up to the full collection with no overlap and no gaps.
+
+## Restricting a Vector Search to a Slice
+
+`slice` is not limited to `scroll`. It also works as a `query_points` filter, so a vector search can be restricted to a fixed, reproducible subset of the collection, for example to hold out a portion of the data for evaluation while searching the rest:
```
generall: I am not sure that's recommended. HNSW is most likely won't be happy about such a slicing and we are not building any extra links. 

Unless you know some really practical usage for this filter in query, I would propose to drop the section

qdrant-landing/content/documentation/tutorials-search-engineering/slicing-filter.md
```diff
@@ -0,0 +1,260 @@
...
+```python
+import random
+
+random.seed(0)
+
+products = {
+    "electronics": ["wireless earbuds", "4K monitor", "mechanical keyboard", "USB-C hub", "smartwatch"],
+    "books": ["science fiction novel", "cookbook", "history book", "graphic novel", "poetry collection"],
+    "clothing": ["running shoes", "wool sweater", "denim jacket", "rain jacket", "cotton t-shirt"],
+    "home": ["cast iron pan", "ceramic mug", "throw blanket", "desk lamp", "storage basket"],
```
generall: doesn't fit horizontal size of the renderred snippet

qdrant-landing/content/documentation/tutorials-search-engineering/slicing-filter.md
AstraBert: Yeah, the only cases I can think of (isolate a specific time slice if you're using time-based numeric IDs, or designing IDs around the hashing function) are a bit of a stretch. Will drop it

Dylancouzon:
- The intro still mentions a section that was removed
- `next_page_offset` is still discarded twice

Should be good to go after that. Feel free to re-request review when all is done

Dylancouzon (approved):
lgtm - let's wait on Neil's confirmation before merge

kanungle (changes requested):
Nice tutorial, just 2 minor improvements needed:

1) some lines of code exceed the width of the code snippets frames
2) I believe tutorials like this benefit from diagrams that illustrate the concept, at least at the beginning

AstraBert:
@kanungle done :)

kanungle (approved):

PR #2734, opened 2026-09-09
feat: securing and exposing a hybrid cloud qdrant instance (auth, tls and ingress guide)

AstraBert:
### [PREVIEW](https://deploy-preview-2734--condescending-goldwasser-91acf0.netlify.app/documentation/tutorials-operations/secure-and-expose-hybrid-cloud/)

Tutorial on how to set up, for your Hybrid Cloud cluster:

- An API key
- A Qdrant-level TLS
- An ingress controller with a way to expose the instance to external traffic while still keeping end-to-end encryption (TLS stays on Qdrant, ingress controller encrypts internal traffic over HTTPS)


qdrant-landing/content/documentation/tutorials-operations/secure-and-expose-hybrid-cloud.md
```diff
@@ -0,0 +1,212 @@
...
+EOF
+```
+ 
+In production, replace the self-signed certificate with one from a trusted CA and drop this resource entirely, along with the annotation that references it below.
+ 
+### Configure the backend connection
+ 
+Traefik needs two pieces of configuration to reach an HTTPS backend: which protocol scheme to use, and which `ServersTransport` to trust it with. Both go **on the Qdrant Service, not the Ingress**:
+ 
+```bash
```
bashofmann: In order to make this persistent and not have the operator overwrite this, this should be set through the Kubernetes configuration in the UI. There is a service annotations field for this.

AstraBert: Sounds good, I will swap that in, although this specific change persisted also when i applied it with kubectl :)

AstraBert: Done!

bashofmann (approved):
lgtm

qdrant-landing/content/documentation/tutorials-operations/secure-and-expose-hybrid-cloud.md
```diff
@@ -0,0 +1,212 @@
...
+kubectl apply -f - <<EOF
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: qdrant-ingress
+  namespace: qdrant
+spec:
+  ingressClassName: traefik
+  rules:
+  - host: qdrant.<your-ingress-external-ip>.nip.io
```
bashofmann: Could we use sslip.io instead of nip.io?

https://en.wikipedia.org/wiki/Nip

bashofmann (changes requested):

qdrant-landing/content/documentation/tutorials-operations/secure-and-expose-hybrid-cloud.md
AstraBert: So sorry, i wasn't aware of the meaning of the word. I will change it immediately

AstraBert: Done!

kanungle (approved):

Dylancouzon:
A few Claude nits, feel free to disregard 

Must fix

1. Replace the hardcoded `qdrant` namespace with `$KUBENS` everywhere. The page already exports and uses `$KUBENS` for `kubectl get`, but still hardcodes `qdrant` in both secret commands, the `mkcert` SAN, `ServersTransport`, `Ingress`, and the transport annotation. That annotation also depends on the namespace because `<namespace>-<name>@kubernetescrd` changes with it.

2. Re-open the port-forward after saving TLS. The save triggers a rolling restart, and `kubectl port-forward` stays attached to the old pod. By the TLS verification step, the existing forward is dead, so readers get a connection error instead of the response shown.

Should fix

3. Explain where the edge certificate comes from. There’s no `tls:` block on the Ingress, so Traefik serves its default self-signed cert, while `-k` hides that. The Next Steps section then says to replace “the self-signed certificate,” but at that point there are actually two.

4. Either redirect HTTP to HTTPS or pin the Ingress to `websecure`. Right now the route still accepts plain HTTP, and the verification command sends the API key in a header, so the key can cross the wire unencrypted.

5. Don’t hardcode `HTTP/0.9` as the expected failure. The exact curl error varies by build and TLS stack. Better to say a TLS/connection error is expected, with the key point being that HTTPS succeeds.

AstraBert:
@Dylancouzon done! just skipped 4 as, per my understanding, that is not accurate(?) at least in the last code snippet, which I think the comment is referring to, we use HTTPS

bashofmann (approved):

PR #2739, opened 2026-09-10
Island shortcode

trean:
This PR makes a proposal and provides an example of how we can integrate interactive diagrams into our articles and documentations.

Main principles:

- Diagrams should not explode markdown. Tag link to diagram is minimal

- It should stay readable even if diagram is unavailable. Hence the fallback to image

- Diagram code should be self-contained. It should not depend on third-party libs or require modifications in global page code. In other words deletion of diagram folder should not affect anything on the website except for diagram itself.

- Diagrams should follow the dark/light color schema of the website

---

Preview:

https://deploy-preview-2739--condescending-goldwasser-91acf0.netlify.app/documentation/manage-data/multitenancy/

https://deploy-preview-2739--condescending-goldwasser-91acf0.netlify.app/documentation/manage-data/quantization/


generall:
## How to add a new interactive diagram (island)

Reference implementations: `content/documentation/headless/multitenancy/*` and `content/documentation/headless/quantization/*`.

### 1. Create the island folder

Group islands by the page that uses them, inside the headless bundle so the raw files are never published:

```
content/documentation/headless/<page>/<island-name>/
  index.js    required — ES module exporting mount(node)
  index.css   optional — styles scoped under the island's root class
```

`index.js` must export one function:

```js
export function mount(node) {
  node.classList.add('qi-xx');          // unique root class, all CSS is scoped under it
  node.innerHTML = '...';                // controls + <svg viewBox="..."> + <p class="qi-xx__status">
  // ...wire up state and events...
  node.dispatchEvent(new CustomEvent('island:ready', { bubbles: true }));  // hides the fallback image
}
```

Conventions that the existing islands follow, keep them for consistency:

- Pure SVG + CSS, no dependencies. Use a `viewBox` and `width: 100%` so it scales with the column.
- Chrome colors (text, borders, control backgrounds) come from CSS custom properties defined on the root class, with a dark override under both `@media (prefers-color-scheme: dark) { html:not([data-theme='light']) .qi-xx {…} }` and `html[data-theme='dark'] .qi-xx {…}`. Take values from `_theme-variables.scss`. Data colors (tenant hues, sign hues) stay constant across themes.
- Pad the `viewBox` by a few units (e.g. `viewBox="-3 0 766 266"`) so strokes on edge cells are not clipped.
- Highlight on hover with a neutral outline (`--xx-hot`, foreground color), not the red brand accent.
- Reserve a fixed height for the status text (`min-height: calc(2 * 1.45em)`) and keep every message within two lines at ~600px width, otherwise the page jumps on hover.
- Do not render a title or caption inside the island; the shortcode renders `title` as a shared `<figcaption>`.
- Put the geometry constants at the top of the file and a header comment explaining what the island shows.

### 2. Embed it in the markdown

Always use the paired form so the old static image stays as the no-JS fallback:

```
{{< island
    path="content/documentation/headless/<page>/<island-name>"
    width="80%" ratio="2 / 1"
    title="Caption shown under both the fallback image and the live island."
>}}
![Alt text](/docs/old-static-image.png)
{{< /island >}}
```

- `ratio` reserves the space before hydration. Set it to the island's natural aspect ratio (controls + svg + status), not the image's, so nothing shifts when it mounts. Measure it once in the browser.
- `width` behaves like `figure`'s width.

### 3. Verify

- The loader and shell CSS ship only on pages that contain the shortcode, nothing else to wire up.
- Check both themes, hover states, and that the island height does not change on hover.
- **Dev server caveat:** Hugo does not track `readFile` across bundles, so editing only the island files does not re-render the page. Save any change to the `.md` page (or a template) to see the update; the content hash in the resource name then picks up the new file. A restart of `hugo server` also works.

trean:
**One more caveat:** they are not **true** islands - they are not isolated from the other code on the page (including each other), so make sure that class names in one visualization on the page don't match class names in another, to avoid overwriting styles (unless it's intentional).

Dylancouzon:
Really like the direction here. A few changes would make these diagrams feel more like Qdrant and give us a cleaner system to reuse across the docs:

- I’d replace the success, warning, and error colors used for tenants. Those colors imply status, even though the tenants are neutral categories. We could use blue (`#2F6FF0`), violet (`#8547FF`), teal (`#038585`), and Qdrant red (`#DC244C`) instead.
- Let’s use the same colors, backgrounds, borders, fonts, and button styles across all the diagrams. Each diagram should only introduce new colors when they help explain something specific.
- Geist Mono works well here. I’d use it consistently across controls and diagram labels
- The cards should use thin neutral borders, modest corner radii, and no shadows. The toggle and chips should be consistent between diagrams
- Some captions are doing too much. I’d keep each caption to one short sentence and leave the detailed explanation in the article.

generall:
>  We could use blue (#2F6FF0), violet (#8547FF), teal (#038585), and Qdrant red (#DC244C) instead.

some diagrams require different "saturation" of the same color. Having just 3 colors is not enough. Plus used colors are the same as in original diagram and they were driven by the style book.

> The cards should use thin neutral borders, modest corner radii, and no shadows. The toggle and chips should be consistent between diagrams

are they not already?

> Some captions are doing too much. I’d keep each caption to one short sentence and leave the detailed explanation in the article.

You mean stuff like this?

```
Values from −1.5 to −1.0 are below −σ → bucket −1 → bits 00. 3 of the 16 components fall here.
```

Dylancouzon:
> some diagrams require different "saturation" of the same color. 

Agreed, but I believe the default colors should still be within Qdrant's design theme

> are they not already?

My phrasing wasnt clear: in the Multitenancy page, the first pills (Tenant A, B, C…) are more rounded than those in the subsequent diagram (user_2, user_3…)

> You mean stuff like this?

It’s mainly the caption beneath the diagram. The prose describes a concept, the diagram visualizes it, and the caption explains the diagram, hence re-explains the prose above e.g.: 
> Grouping tenants together by tenant ID, if is_tenant=true is used, enables more efficient disk reads. Rather than many random seeks across the file, Qdrant can read the data for a specific tenant with a sequential read.

(These could have been pre‑existing)

generall:
> Agreed, but I believe the default colors should still be within Qdrant's design theme

I tried,  doesn't work with those diagrams. Maybe extended pallet?

 > My phrasing wasnt clear: in the Multitenancy page, the first pills (Tenant A, B, C…) are more rounded than those in the subsequent diagram (user_2, user_3…)

fixed

Dylancouzon:
What do we think of this ?

image
image

generall:
@Dylancouzon I had the same result and didn't like it. Not contrast enough.

Specifically violet (#8547FF)

generall:
island-palette-dark
island-palette-light

generall (approved):

Dylancouzon (approved):
This looks great! I’m happy to have the DevRel team review and update our existing content.

PR #2743, opened 2026-09-11
Publish each page's markdown at /path.md as well as /path/index.md

KShivendu:
## What

Every page already publishes a markdown version at `/path/index.md`. This adds a second copy at `/path.md`, so both work.

```
/articles/immutable-data-structures.md        404  ->  200 text/markdown
/articles/immutable-data-structures/index.md  200      200 (unchanged)
```

Nothing moves and no published URL changes. These are additions.

## Why

`/path/index.md` is correct and stays. It is what the [llms.txt convention](https://llmstxt.org/) prescribes for directory-style URLs:

> …either with .md appended (page.html.md) or with the extension replaced by .md (page.md). **(URLs without file names should append index.html.md or index.md instead.)**

It is also the form `/llms.txt` advertises for 670 pages, and the target of the `<link rel="alternate" type="text/markdown">` we already emit in every page head. That last one is worth keeping in mind: on discoverability we are ahead of several documentation sites that ship no alternate link at all.

The gap is that `/path/index.md` is not **guessable**. Every other documentation site I checked exposes a page's markdown by swapping the extension:

| site | `page.md` | `/index.md` |
|---|:---:|:---:|
| Stripe | ✅ | ❌ |
| Anthropic | ✅ | ❌ |
| Vercel | ✅ | — |
| **Qdrant (today)** | **❌** | **✅** |

A client holding only a URL can build `/articles/foo.md` on the spot. Reaching `/articles/foo/index.md` means fetching the HTML first to read the alternate link, which is the round trip the markdown endpoint exists to avoid.

## How

A post-build step, `qdrant-landing/scripts/emit-md-aliases.sh`, copies every `public/**/index.md` to `public/**.md` after Hugo runs. Netlify already serves `.md` as `text/markdown`, so there is no header or redirect config to add.

Cost: **7.6 MB against a 1.4 GB build, 0.5%.**

Deliberately a build step rather than a Netlify redirect. A splat has to be terminal, so `/articles/*.md` is not a pattern Netlify accepts, and a rule that cannot be verified locally is worse than a copy that can.

## Testing

`npm run md:test` builds the site, runs the script over the output, and asserts:

- all 789 aliases exist and match their source byte for byte
- the canonical `index.md` and `index.html` are untouched
- nothing is written outside the publish directory
- a second run overwrites nothing

4/4 pass. I also stubbed the script out and confirmed the two tests that assert aliases exist do fail, so they are not passing vacuously.

The test removes its build directory on exit. A build is ~1.4 GB, so a suite that leaves temp directories behind fills a disk faster than anyone notices.

## Not in scope

`Accept: text/markdown` content negotiation on the HTML URL. Stripe, Anthropic and Vercel all support it, but on Netlify it needs an edge function running on every request to the site. Worth considering separately if anything actually asks for it.

### DIFF
--- a/install-and-build.sh
+++ b/install-and-build.sh
@@ -31,4 +31,5 @@ curl -LJO "https://github.com/sass/dart-sass/releases/download/${DART_SASS_VERSI
     tar -xf "${SASS_ARCHIVE}" && \
     rm "${SASS_ARCHIVE}" && \
     export PATH="${CURRENT_DIR}/dart-sass:${PATH}" && \
-    cd qdrant-landing && npm install && hugo --gc --minify --config config.toml,config-theme.toml --buildFuture -b ${DEPLOY_PRIME_URL}
+    cd qdrant-landing && npm install && hugo --gc --minify --config config.toml,config-theme.toml --buildFuture -b ${DEPLOY_PRIME_URL} && \
+    ./scripts/emit-md-aliases.sh public
--- a/qdrant-landing/package.json
+++ b/qdrant-landing/package.json
@@ -38,5 +38,8 @@
   "version": "0.1.0",
   "workspaces": [
     "packages/hugoautogen"
-  ]
+  ],
+  "scripts": {
+    "md:test": "node --test \"test/markdown/**/*.test.mjs\""
+  }
 }
--- a/qdrant-landing/scripts/emit-md-aliases.sh
+++ b/qdrant-landing/scripts/emit-md-aliases.sh
@@ -1,0 +1,40 @@
+#!/usr/bin/env bash
+#
+# Publish every /path/index.md a second time as /path.md.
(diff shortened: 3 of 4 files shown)


abdonpijpelink:
It's too bad that Netlify doesn't have a cleaner way to implement "Markdown for agents". I don't feel strongly about whether we need `/articles/immutable-data-structures.md` in addition to `/articles/immutable-data-structures/index.md`, but if the consensus is that we do, then this seems like a reasonable workaround.

generall (approved):
tests seems redundant, otherwise lgtm

KShivendu:
Tests trimmed: 87 lines to 39, no Hugo build, 0.26s. Kept the two cases the script actually decides — skip the site root, never overwrite.

PR #2754, opened 2026-09-15
docs: GraphRAG over PDF Page Images with ColPali and Neo4j

anush008:
## NOTE

The PR here is only for review.
The tutorial will be published on the partner's site.


Dylancouzon (changes requested):
Apologies for the delay here. A few changes needed:
1. Make this one runnable script, built block by block. It should include dependency installation and run as-is from top to bottom. A Colab notebook would probably make the most sense.
2. Use full URLs for every link.
3. Follow Neo4j’s structure for technical posts. Their guide calls for a few things we’re still missing:
   - An intro explaining why a graph is a good fit for this problem.
   - A schema diagram, plus a Neo4j Browser screenshot of the actual graph.
   - The exact Cypher used to create constraints and indexes.
   - A conclusion.
4. Replace “Next Steps” and “Further Reading” with a single ending that works naturally on Neo4j’s site.
5. Cut the re-ingest section down to one or two sentences.
6. Reword “Why use both databases?” It makes Qdrant sound like a database.
7. Show the text-extraction baseline failing, including its actual output.

anush008:
> 1. Make this one runnable script, built block by block. It should include dependency installation and run as-is from top to bottom. A Colab notebook would probably make the most sense.

We have repo that can be run and explored locally.
https://github.com/qdrant-labs/colpali-graphrag-demo

Does that work? A single notebook would probably be too dense

Dylancouzon:
Yes, the repo works if Neo4j is okay with it. The page should then be a walkthrough of it.

1. Change "Set Up the Project" to git clone + cp .env.example .env + uv sync, so those are the only commands readers run. Every other code block should be an exact excerpt from the repo, labeled with its file path. Copying .env.example also fixes the model name: the page sets OPENAI_MODEL=openai/gpt-5.6-luna with no OPENAI_BASE_URL, and that name only resolves on a router like OpenRouter.
2. Keep the collection config, query_points, the prompt, and the Cypher. Summarize the UUID helpers, PointStruct, and .env in one sentence each. The "Ask a Question" snippet isn't in the repo, so swap it for the real code from pipeline.py.
3. For the diagram, you can generate one using this skill: https://qdrant.slack.com/archives/C097TKK8FQQ/p1787265187166269. Once it's in, the ASCII architecture and schema blocks can go.
4. For the Browser screenshot, a matching Qdrant UI screenshot next to it would be great: the same page's point, showing its ID and payload. That shows the shared ID better than the prose does, so it can replace the "copy the id and look it up" paragraph.

Overall, it's more complicated than it needs to be. To simplify:

- Build the page around one question. Right now neither question needs both the image and the graph: the signature question needs the image but not the graph, and DynCorp needs the graph but probably not the image. Try to find a question that needs a fact only in a scanned page plus a hop, and make it the main example. That's also the question for the text-extraction baseline to fail on, which is still missing.
- Once you have that question, shrink the signature question to a short "ColPali finds the scanned page" moment, or drop it.
- Cut the "By the end, you will have a small CLI" block. It repeats "Run the Pipeline".
- Cut "Render Pages and Create Page IDs" to one sentence: the page ID is a UUID5 of the document and page number, and both systems use it.
- Mention the BP International limitation once, in the conclusion, and cut the conclusion's first paragraph to two sentences.

